PR #97, opened 2015-02-03
Update to new FromStr

maurer:
This updates rust-postgres to work with the new FromStr in the git master version of rustc.

You might want to wait to merge this until rustc manages to release a new nightly, otherwise Travis will fail all tests.


sfackler:
Thanks!

PR #99, opened 2015-02-15
Update openssl version

s-panferov:
Current openssl version conflicts with [hyper](https://github.com/hyperium/hyper) and [cookie-rs](https://github.com/alexcrichton/cookie-rs). 

I need them both in my [postgres-example](https://github.com/rustless/rustless/tree/master/examples/postgres).


smashwilson:
:+1: I just hit this as well.

sfackler:
Thanks! Published v0.6.3

PR #130, opened 2015-06-09
make rust-crypto dependency optional

jmesmon:
- Some projects &/or people don't need md5 auth.
 - rust-crypto appears fairly inactive (no merges since April)

### DIFF
--- a/tests/test.rs
+++ b/tests/test.rs
@@ -710,11 +710,13 @@ fn test_plaintext_pass_wrong_pass() {
 }
 
 #[test]
+#[cfg_attr(not(feature = "rust-crypto"), ignore)]
 fn test_md5_pass() {
     or_panic!(Connection::connect("postgres://md5_user:password@localhost/postgres", &SslMode::None));
 }
 
 #[test]
+#[cfg_attr(not(feature = "rust-crypto"), ignore)]
 fn test_md5_pass_no_pass() {
     let ret = Connection::connect("postgres://md5_user@localhost/postgres", &SslMode::None);
     match ret {
@@ -725,6 +727,7 @@ fn test_md5_pass_no_pass() {
 }
 
 #[test]
+#[cfg_attr(not(feature = "rust-crypto"), ignore)]
 fn test_md5_pass_wrong_pass() {
     let ret = Connection::connect("postgres://md5_user:asdf@localhost/postgres", &SslMode::None);
     match ret {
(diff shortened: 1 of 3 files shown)


sfackler:
Why does the activity level of rust-crypto matter? It has an MD5 implementation that works and will continue to work in the future.

jmesmon:
@sfackler Unfortunately, rust-crypto is broken on some arm platforms https://github.com/DaGenix/rust-crypto/pull/303 , and until a fix is merged I've essentially got to require that users of my code create a `.cargo/config` with the appropriate `paths = ` variable, as there is no way to override a package in Cargo.toml.

The inability to override a package combined with my lack of confidence in a timely update to rust-crypto means I'm stuck doing the next best thing: working with libraries that depend on rust-crypto to work-around my inability to sanely use rust-crypto in a cargo based project.

Yes, you'd be right in saying that this is an awful workaround for a lack of cargo support for package overrides, but I don't really have many other options.

sfackler:
That's unfortunate. We might as well just stick a copy of rust-crypto's MD5 module into rust-postgres and drop the dependency entirely at that point. It's Apache/MIT licensed so it should be fine if we stick a copy of the license in THIRD_PARTY.

tests/test.rs
```diff
@@ -710,11 +710,13 @@ fn test_plaintext_pass_wrong_pass() {
 }
 
 #[test]
+#[cfg_attr(not(feature = "rust-crypto"), ignore)]
```
zr40: `#[cfg(feature = "rust-crypto")]` will exclude the test entirely unless this feature is enabled.

jmesmon: I've chosen to use `ignore` here so that the tests are still compiled even when the feature is disabled, hopefully avoiding accidental compile breakages.

I based it on some ignored tests (for unix-socket).

Is there a reason you believe we should prefer disabling them?

sfackler: Yep, I usually conditionally ignore tests when possible.

sfackler:
I've pulled rust-crypto's md5 impl into rust-postgres. It's pure Rust, so there shouldn't be any platform specific weirdness: https://github.com/sfackler/rust-postgres/commit/53aafc36f6b97e5c6cf95d5cfa4cc3bd2122357e

jmesmon:
Closed as it is no longer a dependency. Thanks @sfackler.

sfackler:
I just published v0.9.2 which no longer has a rust-crypto dependency.

jmesmon:
Thanks!

PR #137, opened 2015-08-01
Implement ToSql for &'a T where T: ToSql

reem:
This allows more flexible use of ToSql in generic contexts by allowing
references to be used for encoding.

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -815,6 +815,25 @@ pub trait ToSql: fmt::Debug {
                       -> Result<IsNull>;
 }
 
+impl<'a, T> ToSql for &'a T where T: ToSql {
+    fn to_sql_checked(&self, ty: &Type, out: &mut Write, ctx: &SessionInfo)
+                      -> Result<IsNull> {
+        if !<&'a T as ToSql>::accepts(ty) {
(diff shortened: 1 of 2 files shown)


sfackler:
Can you add a quick test to just make sure the to_sql call doesn't accidentally infinitely recurse?

reem:
Test added.

src/types/mod.rs
```diff
@@ -815,6 +815,16 @@ pub trait ToSql: fmt::Debug {
                       -> Result<IsNull>;
 }
 
+impl<'a, T> ToSql for &'a T where T: ToSql {
+    to_sql_checked!();
```
sfackler: I believe all you need to do to fix the ICE is to implement this manually - see the implementations for &str and &[u8] for example.

reem:
Manually implemented to_sql_checked, it does appear to have fixed the ICE.

sfackler:
No dice: https://travis-ci.org/sfackler/rust-postgres/jobs/73953899#L159

I believe you'll need to replace `Self` with `&'a T` to fix it.

reem:
Ok, actually tested on stable this time, it should work.

PR #140, opened 2015-09-28
To_String() cleanup for to_owned on method calls

ddrmanxbxfr:
Hi,

This PR cleans up the calls of to_string in method calls. 

We should favor the usage of to_owned as to_string uses the whole formatting machinery just to clone a string. 

Using .to_owned() is lighter on resources.

Thanks for looking into it !


sfackler:
Thanks!

PR #156, opened 2016-01-29
Add keywords to improve linkability on crates.io

kud1ing:
* https://crates.io/keywords/postgres
* https://crates.io/keywords/postgresql


sfackler:
Weird, I thought the readme was indexed for search as well - is that not the case @alexcrichton?

alexcrichton:
The readme is indexed, yeah, but it has a much lower search priority than the keywords, for example.

kud1ing:
Thank you.

I think keyword links are more precise. If you search for "orm" you also get https://crates.io/crates/rustless/ which i would not consider an ORM.

kud1ing:
My motivation is to add crates.io links to https://github.com/kud1ing/awesome-rust and i want them to be precise.

PR #158, opened 2016-02-14
Macaddr

abaumhauer:
Here is my implementation of MACADDR handling using my crate eui48.  There are only minor changes to your code to support it.  Thanks for making it so easy to extend your crate for additional types.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,4 +10,4 @@ before_script:
 - "./.travis/setup.sh"
 script:
 - cargo test
-- cargo test --features "uuid rustc-serialize time unix_socket serde_json chrono openssl bit-vec"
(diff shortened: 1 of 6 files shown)


sfackler:
Can you tweak this line so that travis tests this feature? https://github.com/sfackler/rust-postgres/blob/master/.travis.yml#L13

Looks good other than that, thanks!

abaumhauer:
Ok, change made.  Do I need to do another pull request, or is the
current one ok?
Andy

sfackler:
Nope, should be fine in this PR. I forgot one more thing - could you add a test for this? You should be able to model it after the tests for other optional types like UUID: https://github.com/sfackler/rust-postgres/blob/master/tests/types/uuid.rs

abaumhauer:
Added travis test to exercise parse_str() function as requested.

sfackler:
Awesome, thanks!

PR #164, opened 2016-03-12
Added crude BigRational ToSql FromSql for numeric

kvsari:
Please have a look at this code and let me know if it's acceptable.


sfackler:
Could you add some tests to exercise these implementations?

kvsari:
I want to have a random battery of tests but for that I need to import the rand crate I imagine.

sfackler:
You'll need to adjust .travis.yml to add the new num feature.

kvsari:
Cancelling pull request. I don't know what to do about 1/3.

kvsari:
I've added some battery tests and set the 1/3 test to fail.

sfackler:
I'm not sure that `BigRational` is the right type to match to `NUMERIC` due to the issues with e.g. 1/3. It seems like it really needs something like [BigDecimal](https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html).

ELD:
It really does since BigDecimal allows you to specify the precision.

Adding a BigDecimal type to `num` is something I've been meaning to experiment with. I understand the concepts, I just need to sit down and try to write it out.

kvsari:
Ok, I'll close this request until a BigDecimal is available.

PR #172, opened 2016-04-17
Support chrono::Date<UTC> for type DATE

billiob:
Also do it for chrono::Date<Local> and chrono::Date<FixedOffset>

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,10 @@ types. The driver currently supports the following conversions:
         <tr>
             <td>
                 <a href="https://github.com/lifthrasiir/rust-chrono">chrono::NaiveDate</a>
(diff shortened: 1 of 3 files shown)


sfackler:
It seems somewhat inaccurate to map a Postgres `DATE`, which is associated with no timezone, to a chrono `Date<UTC>`.

billiob:
The types `Date<UTC>` or `Date<Local>` are a bit strange.
I was using `UTC::today()` and figured `Date<UTC>` did not have `ToSql`. I could also convert them to `NaiveDate`.

sfackler:
I would recommend converting to `NaiveDate` and using the existing implementations for that type.

PR #178, opened 2016-05-13
Ignore InvalidCatalogName errors for pg_catalog queries

### LINKED ISSUE #171: Make it compatible with cockroachdb
cockroachdb is compatible with the postgres wire protocol but not with all the postgres internals. Many other postgres drivers work fine with cockroachdb, but this one doesnt!

Error:

> thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Io(Error { repr: Custom(Custom { kind: InvalidInput, error: StringError("the server returned an unexpected response") }) })', ../src/libcore/result.rs:688

Repro:

```
wget https://binaries.cockroachdb.com/cockroach-beta-20160407.linux-amd64.tgz && tar xfz cockroach-beta-20160407.linux-amd64.tgz && cd cockroach-beta-20160407.linux-amd64
./cockroach start
./cockroach sql 
> CREATE DATABASE root;
```

 ```
 let conn = Connection::connect("postgresql://root@localhost:26257/root?sslmode=disable", SslMode::None) .unwrap();
	println!("connection established");
    conn.execute("CREATE TABLE accounts (id INT PRIMARY KEY, balance INT)", &[]).unwrap();
	println!("executed statement");
```

nvanbenschoten:
Fixes #171. See https://github.com/sfackler/rust-postgres/issues/171#issuecomment-219171884.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -421,6 +421,8 @@ impl InnerConnection {
             Err(Error::Io(e)) => return Err(ConnectError::Io(e)),
             // Old versions of Postgres and things like Redshift don't support enums
             Err(Error::Db(ref e)) if e.code == SqlState::UndefinedTable => {}
+            // Some Postgres compliant databases are missing a pg_catalog
+            Err(Error::Db(ref e)) if e.code == SqlState::InvalidCatalogName => return Ok(()),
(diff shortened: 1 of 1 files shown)


sfackler:
Thanks!

PR #183, opened 2016-05-19
Handle malformed percent-encoding

### LINKED ISSUE #182: Panic on passwords without percent encoding
Passwords that have not properly been percent encoded
```rust
fn main() {
    let _ = Connection::connect("postgres://user:%1*@localhost", SslMode::None);
}
```

result in a panic
```
thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidHexCharacter { c: '*', index: 1 }', ../src/libcore/result.rs:746
note: Run with `RUST_BACKTRACE=1` for a backtrace.
error: Process didn't exit successfully: `target/debug/rspsql` (exit code: 101)
```

millerjs:
I made the mistake of assuming percent encoding would be done internally and hit a panic. This fix checks the result of `Vec::<u8>::from_hex` for invalid hex values that could result from a user passing a password like `%1*`.

Closes #182

### DIFF
--- a/src/url.rs
+++ b/src/url.rs
@@ -131,8 +131,17 @@ fn decode_inner(c: &str, full_url: bool) -> DecodeResult<String> {
                             }
                         };
 
+                        let bytes_from_hex = match Vec::<u8>::from_hex(&bytes) {
+                            Ok(b) => b,
+                            _ => {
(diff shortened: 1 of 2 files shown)


sfackler:
Thanks!

PR #184, opened 2016-05-25
Add savepoint method to Transaction

### LINKED ISSUE #179: Custom savepoint names for subtransactions
Of the few Postgres drivers that support subtransactions, it seems like most permit access to the subtransaction's "name". For instance, the official [C++ driver](http://pqxx.org/development/libpqxx/) takes a name in as a [parameter](http://pqxx.org/development/libpqxx/browser/trunk/src/subtransaction.cxx) when creating a subtractions. This name is then used as the subtransaction's savepoint name where applicable. It seems like this driver currently [hardcodes](https://github.com/sfackler/rust-postgres/blob/master/src/transaction.rs#L246) the name `"sp"` in the three places that a name is needed.

Would you be open to providing access to this savepoint name on subtransactions through either:
- an added parameter to the current constructor (breaking change)
- a secondary constructor with the extra parameter (eg. `transactionWithName` or `namedTransaction`)
- a setter method (which would require extra logic to prevent it's use on non-nested transactions)

If so, I'd be happy to send a PR with the change!

nvanbenschoten:
Fixes #179 

This change creates a `Transaction.savepoint` method, which is equivalent
to `Transaction.transaction`, but takes a custom name for the nested
transaction's savepoint name.

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -151,6 +151,7 @@ impl Config {
 pub struct Transaction<'conn> {
     conn: &'conn Connection,
     depth: u32,
+    savepoint_name: Option<&'conn str>,
     commit: Cell<bool>,
     finished: bool,
 }
@@ -177,6 +178,7 @@ impl<'conn> TransactionInternals<'conn> for Transaction<'conn> {
         Transaction {
             conn: conn,
             depth: depth,
+            savepoint_name: None,
             commit: Cell::new(false),
             finished: false,
         }
@@ -195,14 +197,13 @@ impl<'conn> Transaction<'conn> {
     fn finish_inner(&mut self) -> Result<()> {
         let mut conn = self.conn.conn.borrow_mut();
         debug_assert!(self.depth == conn.trans_depth);
-        let query = match (self.commit.get(), self.depth != 1) {
-            (false, true) => "ROLLBACK TO sp",
-            (false, false) => "ROLLBACK",
-            (true, true) => "RELEASE sp",
-            (true, false) => "COMMIT",
-        };
         conn.trans_depth -= 1;
-        conn.quick_query(query).map(|_| ())
+        match (self.commit.get(), self.savepoint_name) {
+            (false, Some(savepoint_name)) => conn.quick_query(&format!("ROLLBACK TO {}", savepoint_name)),
+            (false, None) => conn.quick_query("ROLLBACK"),
+            (true, Some(savepoint_name)) => conn.quick_query(&format!("RELEASE {}", savepoint_name)),
+            (true, None) => conn.quick_query("COMMIT"),
+        }.map(|_| ())
     }
 
     /// Like `Connection::prepare`.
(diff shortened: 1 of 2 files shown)


src/transaction.rs
```diff
@@ -151,6 +151,7 @@ impl Config {
 pub struct Transaction<'conn> {
     conn: &'conn Connection,
     depth: u32,
+    savepoint_name: Option<&'conn str>,
```
sfackler: I'd just make this an `Option<String>` - the runtime cost of the allocation is ~0 compared to talking over the network.

src/transaction.rs
```diff
@@ -233,22 +234,33 @@ impl<'conn> Transaction<'conn> {
...
+        self.savepoint("sp")
+    }
+    
+    /// Like `Connection::transaction`, but creates a nested transaction
+    /// with the provided name.
+    ///
+    /// # Panics
+    ///
+    /// Panics if there is an active nested transaction.
+    pub fn savepoint<'a>(&'a self, name: &'a str) -> Result<Transaction<'a>> {
```
sfackler: After the above changes, this can just change to `name: &str`.

tests/test.rs
```diff
@@ -295,6 +295,15 @@ fn test_trans_with_nested_trans() {
 }
 
 #[test]
+#[should_panic(expected = "active transaction")]
+fn test_trans_with_savepoints() {
+    let conn = or_panic!(Connection::connect("postgres://postgres@localhost", SslMode::None));
+    let trans = or_panic!(conn.transaction());
+    let _sp = or_panic!(trans.savepoint("custom"));
```
sfackler: Can you add `commit()` calls to the savepoints here to make sure that they commit successfully?

sfackler: Derp, nevermind.

src/transaction.rs
nvanbenschoten: Sounds good to me. How's that?

I'll squash the commits down to one if/when this PR gets the thumbs up by the way.

sfackler:
:+1: Thanks!

nvanbenschoten:
Hi @sfackler! I noticed that this functionality was lost in the newest release (`0.17`). Was that intentional? E.g. did you decide that this was bloating the API? Or was it just lost in the Tokio rewrite and missed in https://github.com/sfackler/rust-postgres/commit/0d3e18b251137418271f917e320134a439cafe52? If the latter, do you mind if I send a patch to add it back in?

For context, CockroachDB is still very interested in this functionality for the same reason as we originally discussed in https://github.com/sfackler/rust-postgres/issues/179. This client-side retry pattern is documented [in the CRDB docs](https://www.cockroachlabs.com/docs/dev/build-a-rust-app-with-cockroachdb.html#transaction-with-retry-logic) and we'd love to be able to get people onto the newest version of this library. This is especially true because we've seen an uptick in people using Rust with CRDB recently (including me on a few side projects 😃) and also an uptick in people interested in Tokio + async/await.

sfackler:
Oops - the removal was an oversight during the big rewrite. Happy to take a PR adding it back, or I can put one up at some point alternatively.

nvanbenschoten:
Great! I'll try to get a PR out for this today.

PR #185, opened 2016-05-27
address clippy and rust lints

oli-obk:


src/priv_io.rs
```diff
@@ -1,5 +1,4 @@
 use byteorder::ReadBytesExt;
-use net2::TcpStreamExt;
```
sfackler: This needs to stay to support pre-1.9.0 users.

src/lib.rs
```diff
@@ -1,4 +1,4 @@
-//! A pure-Rust frontend for the popular PostgreSQL database.
+//! A pure-Rust frontend for the popular `PostgreSQL` database.
```
sfackler: Why?

oli-obk: The `doc_markdown` lint triggers on camel case words outside ticks

sfackler: That does not seem like a particularly good reason to put PostgreSQL in ticks :P

oli-obk:
addressed comments

sfackler:
Thanks!

PR #188, opened 2016-06-24
Add date and timestamp wrappers that can represent infinity

alex-gulyas:
This patch adds the `types::Date` and `types::Timestamp` enums, which
can represent either a date/timestamp, or positive or negative
infinity.

### DIFF
--- a/src/types/special.rs
+++ b/src/types/special.rs
@@ -1,0 +1,118 @@
+use std::io::prelude::*;
+use std::{i32, i64};
+
+use byteorder::{ReadBytesExt, WriteBytesExt, BigEndian};
+
+use Result;
+use error::Error;
+use types::{Type, FromSql, ToSql, IsNull, SessionInfo};
+
+/// A wrapper that can be used to represent infinity with `Type::Date` types.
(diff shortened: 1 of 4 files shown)


src/types/special.rs
```diff
@@ -0,0 +1,129 @@
...
+
+        let mut buf = [0; 4];
+        try!(raw.read_exact(&mut buf));
+        let mut cursor = Cursor::new(&mut buf);
+
+        match try!(cursor.read_i32::<BigEndian>()) {
+            i32::MAX => Ok(Date::PosInfinity),
+            i32::MIN => Ok(Date::NegInfinity),
+            _ => {
+                try!(cursor.seek(SeekFrom::Start(0)));
```
sfackler: This can be replaced with just `cursor.set_position(0)`, but I think we can avoid the `Cursor` entirely by using `&mut [u8]`s `Read` implementation:

https://is.gd/XIpI1U

sfackler:
Nice! A couple of little tweaks, but looks good in general.

alex-gulyas:
Fixed!

sfackler:
Thanks!

PR #190, opened 2016-07-07
Fix empty array conversion

### LINKED ISSUE #189: Error converting empty array to Vec<T>
For example,

```rust
fn empty_array(conn: &Connection) {
    for row in &conn.query("SELECT '{}'::varchar[]", &[]).unwrap() {
        let _: Vec<String> = row.get(0);
    }
}
```

results in this error:

```
thread '<main>' panicked at 'error retrieving column 0: Conversion(StringError("array contains too many dimensions"))', /Users/curtis/.cargo/registry/src/github.com-88ac128001ac3a9a/postgres-0.11.9/src/rows.rs:204
```

which comes from [this check](https://github.com/sfackler/rust-postgres/blob/5938db205b7d7721acdcbd18541c779c6f4e6540/src/types/mod.rs#L361).

Based on the following, it seems likely that Postgres encodes empty arrays as having zero dimensions.

```
curtis=# SELECT array_ndims('{}'::varchar[]);
 array_ndims
─────────────
           ¤
(1 row)
```

programble:
Fixes #189.

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -358,9 +358,11 @@ impl<T: FromSql> FromSql for Vec<T> {
             _ => panic!("expected array type"),
         };
 
-        if try!(raw.read_i32::<BigEndian>()) != 1 {
-            return Err(Error::Conversion("array contains too many dimensions".into()));
-        }
+        match try!(raw.read_i32::<BigEndian>()) {
+            0 => return Ok(Vec::new()),
+            1 => (),
+            _ => return Err(Error::Conversion("array contains too many dimensions".into())),
+        };
 
         let _has_nulls = try!(raw.read_i32::<BigEndian>());
         let _member_oid = try!(raw.read_u32::<BigEndian>());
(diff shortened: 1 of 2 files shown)


src/types/mod.rs
```diff
@@ -358,9 +358,11 @@ impl<T: FromSql> FromSql for Vec<T> {
             _ => panic!("expected array type"),
         };
 
-        if try!(raw.read_i32::<BigEndian>()) != 1 {
-            return Err(Error::Conversion("array contains too many dimensions".into()));
-        }
+        match try!(raw.read_i32::<BigEndian>()) {
+            0 => return Ok(Vec::new()),
```
sfackler: Does a 0 dimensional array skip the rest of the metadata fields (has_nulls, member_oid, etc)? Does anything break if we just continue on when the number of dimensions is 0?

programble: If I let it keep reading, it hits EOF trying to read `_index_offset`. Seemed easier to just return an empty Vec early.

sfackler: Ok cool, I guess the database doesn't transmit the rest of the stuff.

sfackler:
Thanks!

PR #194, opened 2016-08-06
Add SSL example to README

strangelittlemonkey:
* Added SSL example to README
* Added minor error handling and removal of unwrap

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,14 +33,7 @@ struct Person {
 }
 
 fn main() {
-    let uri = "postgres://postgres@localhost";
-    let conn = match Connection::connect(uri, SslMode::None) {
-        Ok(conn) => conn,
-        Err(e) => {
-            println!("Connection error: {}", e);
-            return;
-        }
-    };
+    let conn = Connection::connect("postgres://postgres@localhost", SslMode::None).unwrap();
     conn.execute("CREATE TABLE person (
                     id              SERIAL PRIMARY KEY,
                     name            VARCHAR NOT NULL,
--- a/README.md
+++ b/README.md
@@ -12,6 +12,13 @@ You can integrate Rust-Postgres into your project through the [releases on crate
 postgres = "0.11"
 ```
 
+If using SSL, integration needs to include the SSL feature in the Cargo.toml
+```toml
+# Cargo.toml
+[dependencies]
+postgres = { version="0.11", features=["openssl"] }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,7 +33,60 @@ struct Person {
...
+use openssl::ssl::{SslContext, SslMethod};
+
+struct Person {
+    id: i32,
+    name: String,
+    data: Option<Vec<u8>>,
+}
+
+fn main() {
+    let context = SslContext::new(SslMethod::Sslv23).unwrap();
```
sfackler: I'm a bit hesitant to use this specifically as an example since it doesn't do any certificate validation at all. OpenSSL unfortunately gives you all of the rope you need to hang yourself :(. The situation here should be better in the next release with the new logic in here: https://github.com/sfackler/rust-postgres/blob/breaks/src/io/openssl.rs#L30-L39

README.md
```diff
@@ -26,7 +33,60 @@ struct Person {
...
+            name: row.get(1),
+            data: row.get(2),
+        };
+        println!("Found person {}", person.name);
+    }
+}
+```
+
+Same example as above but with SSL
+```rust
```
sfackler: It seems like we don't need to duplicate the entire example - just the context setup and connection line should be enough to show people what to do.

README.md
```diff
@@ -26,7 +33,60 @@ struct Person {
 }
 
 fn main() {
-    let conn = Connection::connect("postgres://postgres@localhost", SslMode::None).unwrap();
```
sfackler: Any particular reason to change this?

README.md
strangelittlemonkey: Unless I misunderstand I feel that's a separate issue as it's a bug in the
openssl crate and not the postgres crate

README.md
strangelittlemonkey: I'll reduce the example

README.md
strangelittlemonkey: I was under the impression that unwrap is generally discouraged

sfackler: It's discouraged in "real" code, but used in examples to focus them on the APIs they're showing you.

sfackler:
This is now documented in the 0.12 release: http://sfackler.github.io/rust-postgres/doc/v0.12.0/postgres/#ssltls

PR #206, opened 2016-10-11
Added convenience impl IntoConnectParams for String

p512:
Currently one must convert `String`s to `&str` using `.as_str()` in order to invoke `Connection::connect`.
This may be avoided if `String`s refer to the implementation of `&str` for `IntoConnectParams`

Here's an example before and after this change.

Before:
```
let conn = Connection::connect(env::var("DATABASE_URL").unwrap().as_str(), 
                               TlsMode::None)
```
After:
```
let conn = Connection::connect(env::var("DATABASE_URL").unwrap(), TlsMode::None)
```

### DIFF
--- a/src/params.rs
+++ b/src/params.rs
@@ -66,6 +66,12 @@ impl<'a> IntoConnectParams for &'a str {
     }
 }
 
+impl IntoConnectParams for String {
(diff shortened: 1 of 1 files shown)


sfackler:
Seems reasonable, thanks!

PR #207, opened 2016-10-12
Small syntax error in documentation...

Caspinol:
...messing up the formatting


sfackler:
Those aren't syntax errors. A leading `# ` hides that line from being rendered in the resulting rustdoc output.

Caspinol:
Ok, sorry thought that was what was causing the following:
screen shot 2016-10-12 at 09 54 56

In the "impl Connection" section of the documentation.

PR #224, opened 2016-12-30
Enusre `Box<T: TlsHandshake>` implements `TlsHandshake`

alexcrichton:


sfackler:
I'm a bit unsure as to why this is necessary - can't you just `&*foo`?

alexcrichton:
Oh right yeah this wasn't strictly needed, but it would have avoided the `*` and have been a tiny bit more ergonomic if it were there.

PR #228, opened 2017-01-31
Upgrade serde_json to 0.9.x

### LINKED ISSUE #227: Support serde_json 0.9.x
Currently the project supports a maximum of serde_json < 0.9.
0.9 has some sweet features like javascript-style field accessors,
and i use rust-postgres in a lib of mine so i can't upgrade til we support 0.9 in rust-postgres.

I cloned this repo and the project builds fine against serde_json 0.9,
i'll put up a PR as soon as i can get the test suite to run locally (need to figure out where to set postgres path etc).

aleksanb:
This should enable serde_json 0.9 support on rust-postgres.
Let's see if travis agrees.

The change from {:?} to {} is prompted by the debug impl for the new value type actually being a debug impl: https://github.com/serde-rs/json/blob/master/json/src/value.rs#L101.
Postgres doesn't seem to accept debug printed JSON structures as valid JSON 💃 

Fixes #227.


sfackler:
Thanks!

Issue #244, opened 2017-03-09
Request for feature: support geometric types with rust-geo


kestred:
Would it make sense to support postgreSQL's builtin geometric types using the https://github.com/georust/rust-geo crate?

It is the most used (and i.m.o. seems the most well designed) crate that exists for geo/gis in rust. 

## Type Correspondence

<table>
    <thead>
        <tr>
            <th>Rust Type</th>
            <th>Postgres Type</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>geo::Coordinate, geo::Point</td>
            <td>POINT</td>
        </tr>
        <tr>
            <td>(possibly geo::LineString or geo::MultiPoint)</td>
            <td>LINE, LSEG</td>
        </tr>
        <tr>
            <td>geo::Bbox</td>
            <td>BOX</td>
        </tr>
        <tr>
            <td>geo::LineString</td>
            <td>PATH</td>
        </tr>
        <tr>
            <td>geo::Polygon</td>
            <td>POLYGON</td>
        </tr>
    </tbody>
</table>

sfackler:
Sure, sounds reasonable to me! I'd be happy to take PR adding the conversions, or I can take a crack at it when I have a chance.

kestred:
Great.  I'll implement it and create a PR.

pka:
This could lead to some confusion. People working with geometries within PostgreSQL are using PostGIS. https://github.com/andelf/rust-postgis isn't using rust-geo geometries, because it leads to overhead for converting between different geometry implementations. There was a long [discussion](https://github.com/georust/rust-geo/issues/67) about that, which ended in the consensus that rust-geo needs a trait based implementation.

rofrol:
Close issue?

PR #245, opened 2017-03-10
Support geo types with georust/geo

kestred:
This pull request implements part of https://github.com/sfackler/rust-postgres/issues/244.  Supporting the default postgres geographic types with https://github.com/georust/rust-geo.

The types LINE and LSEG haven't been implemented yet because there is no obvious mapping in the current implementation of the geo crate.  I hope to discuss this with the maintainer of that crate.


kestred:
While implementing this, I had another idea; either an alternative or an additional implementation for geo types.

The geo types could be supported via tuples (eg. `impl ToSql for (f64, f64)`).   This would allow queries to be written like:

```rust
let lat: f64 = 40.018782;
let lng: f64 = -105.278201;
db.execute("INSERT INTO places (loc) VALUES ($1)", &[(&lat, &lng)])
```

--------

The type mappings using tuples COULD be:

<table>
    <thead>
        <tr>
            <th>Rust Type</th>
            <th>Postgres Type</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>(f64, f64)</td>
            <td>POINT</td>
        </tr>
        <tr>
            <td>(f64, f64, f64)</td>
            <td>LINE</td>
        </tr>
        <tr>
            <td>((f64, f64), (f64, f64))</td>
            <td>LSEG, BOX</td>
        </tr>
        <tr>
            <td>[(f64, f64), ...] OR Vec&lt;(f64, f64)&gt;</td>
            <td>PATH</td>
        </tr>
        <tr>
            <td>[(f64, f64), ...] OR Vec&lt;(f64, f64)&gt;</td>
            <td>POLYGON</td>
        </tr>
    </tbody>
</table>

sfackler:
Thanks! Could you add some tests? See here for reference: https://github.com/sfackler/rust-postgres/tree/master/postgres/tests/types

postgres-shared/src/types/geo.rs
```diff
@@ -0,0 +1,104 @@
...
+    to_sql_checked!();
+}
+
+impl FromSql for LineString<f64> {
+    fn from_sql(_: &Type, raw: &[u8]) -> Result<Self, Box<Error + Sync + Send>> {
+        if raw.len() < 5 {
+            return Err("invalid message length".into());
+        }
+
+        // let _ = types::bool_from_sql(&raw[0..1])?; // is path open or closed
```
sfackler: Is it going to be problematic that we're discarding this information?

kestred: This will not cause a database error to occur.

The more interesting part of the issue is with ToSql where there are two different ways to create a path:  an open path `path '[(1,2),(3,4)]'` and a closed path `path '((1,2),(3,4))'`.
The PR handles this by always inserting an open path.  To insert a closed path, you would have to write `pclose($1)` and let Postgres flip that bit.   This behavior is documented in the README.

I opted for that behavior because:
 * The Polygon type is already very similar to a Closed path (see https://www.postgresql.org/docs/9.2/static/datatype-geometric.html#DATATYPE-GEO-TABLE)
 * A LineString is treated like an Open path in most uses

sfackler:
Thanks!

PR #250, opened 2017-04-22
Update serde_json dependency to version 1.0

kimrutherford:
I have a project that uses serde_json and rust-postgres.  I needed this change after upgrading
to serde v1.0.


sfackler:
Can you tweak the bounds to `>= 0.9, < 2.0` so we maintain support for 0.9.x? Thanks!

sfackler:
Thanks!

PR #253, opened 2017-05-01
Use the bytes crate for backend message parsing

sfackler:


sfackler:
@carllerche Am I using `BytesMut` in a reasonableish way here?

carllerche:
Looks like a big diff :) Do you have any specific questions?

sfackler:
I'm most interested in the buffer management here: https://github.com/sfackler/rust-postgres/pull/253/files#diff-0b7f4149c3b9a969d89794de035c50bcR22
And the parsing logic here: https://github.com/sfackler/rust-postgres/pull/253/files#diff-cd3f23a5e64a842c168c010da1a8ffb8R49

I ended up making this buffer type because it's it's not super straightforward to know up front if the message size is right, which `Buf`'s methods seem to be built around: https://github.com/sfackler/rust-postgres/pull/253/files#diff-cd3f23a5e64a842c168c010da1a8ffb8R214

PR #264, opened 2017-06-06
Make Rows a fully owned type

### LINKED ISSUE #263: Enable Rows to be owned and sent to 'static threads
Currently `Rows` is tied to the lifetime of `Statement` which has a borrowed `Connection`. This prevents `Rows` from being owned and sent to a `'static` thread.

The only data used by `Rows` from the `Statement` is already wrapped in an `Arc`. Solution here is to remove `Statement` and give it a clone of the `Arc<StatementInfo>`. This will remove the lifetime from `Rows` and allow it to be owned and sent to 'static thread.

`Row` should remain as a borrowed type to prevent unnecessary reference count bumps to `Arc<StatementInfo>`.

jwilm:
This allows Rows to outlive a statement and be sent to 'static threads.

Resolves #263.

### DIFF
--- a/postgres/src/rows.rs
+++ b/postgres/src/rows.rs
@@ -9,8 +9,10 @@ use std::fmt;
 use std::io;
 use std::ops::Deref;
 use std::slice;
+use std::sync::Arc;
+use std::marker::PhantomData;
 
-use {Result, RowsNew, LazyRowsNew, StatementInternals};
+use {Result, RowsNew, LazyRowsNew, StatementInternals, StatementInfo};
 use transaction::Transaction;
 use types::{FromSql, WrongType};
 use stmt::{Statement, Column};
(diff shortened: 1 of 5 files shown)


postgres/src/rows.rs
```diff
@@ -259,13 +256,21 @@ impl<'a> Row<'a> {
 pub trait RowIndex {
     /// Returns the index of the appropriate column, or `None` if no such
     /// column exists.
-    fn idx(&self, stmt: &Statement) -> Option<usize>;
+    fn idx<I: AsRef<StatementInfo>>(&self, info: &I) -> Option<usize> {
```
sfackler: Making this change to `RowIndex` is still backwards incompatible to implementors. My gut says that just changing the trait will cause any issues in practice though since it's basically an implementation detail.

sfackler: Since we're changing it, maybe make it just take a `&[Column]` to avoid making `StatementInfo` public?

jwilm: Oh, bummer. I didn't realize RowIndex was actually intended to be part of the public API.

At the point this is a breaking change, should we just drop the lifetime parameter from `Rows`?

jwilm: Accepting `&[Column]` does seem like a better approach.

sfackler: It's all about the degree of broken-ness - the odds that anyone has code that implements `RowIndex` or calls `idx` manually are I believe very low, but the odds that anyone refers to `Rows<'a>` with a lifetime seems much higher. We can either cut a new major version or just make the RowIndex change and revert if anyone yells.

jwilm:
The latest commit contains the discussed `RowIndex` changes (`idx` accepts `&[Column]`). `Rows` lifetime was left intact.

sfackler:
Thanks!

jwilm:
Travis build failed, but I think GitHub wasn't available to receive the status update.

The Debug impl is actually used in a couple of places. It's the same impl that was originally on `Statement` and was used by `Rows`. Moving it to `StatementInfo` allowed both sites to continue sharing the same code.

sfackler:
Oh derp, right, sorry. I think it can just be derived, though. The manual impls are just there when I want to skip a field.

jwilm:
Derived form will have a slight difference in struct name. Was trying to preserve formatting to a T ;).

If that doesn't matter, derive is definitely the way to go.

sfackler:
Ah sure :)

jwilm:
Thanks, @sfackler! Always a pleasure working with you.

jwilm:
Would you mind publishing a release with this patch?

sfackler:
Done!

jwilm:
Thanks!

PR #269, opened 2017-06-24
Update the chrono dependency to 0.4

Kerollmops:


sfackler:
Covered by #272 which fixes all of the places that needs to be fixed!

PR #283, opened 2017-08-03
Add Eq, ParialEq and Hash to ConnectParams (etc)

rory:
I wanted to be able to use ConnectParams as a key of a hash (to do simple connection pooling), so I've added these Derives.

### DIFF
--- a/postgres-shared/src/params/mod.rs
+++ b/postgres-shared/src/params/mod.rs
@@ -8,7 +8,7 @@ use params::url::Url;
 mod url;
 
 /// The host.
-#[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


sfackler:
Thanks!

rory:
Could you please bump the version of `postgres-shared` and [the dependency in `postgres`'s `Cargo.toml`](https://github.com/sfackler/rust-postgres/blob/92808c22f5adc8a0d7570cb91453f2513a9ae562/postgres/Cargo.toml#L71). Current version of `postgres-shared` is 0.4.0 which predates this patch being accepted

sfackler:
Done!

PR #287, opened 2017-09-04
bump generic-array dependents' versions

FauxFaux:
Please update `postgres-protocol` to use newer versions of `generic_arary`. This is just a `Cargo.toml` change; no code changes are necessary.

Without this change, `cargo build` on my project either hangs indefinitely during resolution, or fails with odd errors about `typenum`, which I have been otherwise unable to resolve:

```
error: failed to select a version for `typenum` (required by `generic-array`):
all possible versions conflict with previously selected versions of `typenum`
  version 1.9.0 in use by typenum v1.9.0
  possible versions to select: 1.5.2, 1.5.1, 1.5.0
```


sfackler:
Thanks!

PR #296, opened 2017-10-09
Correct build error on non-unix targets for tokio-postgres

Meralis40:
The latest release of tokio-postgres (version "0.3.0") don't build on non unix platform due to a missing error renaming (forgot when error reform was done).


sfackler:
Oops, thanks!

Issue #303, opened 2017-11-18
Error querying PgBouncer


jwilm:
PgBouncer seems to implement some, but not all, of the postgres protocol. Specifically, I was running into this error when trying to query it:

```rust
Db(DbError {
    severity: "ERROR",
    parsed_severity: None,
    code: ProtocolViolation,
    message: "unsupported pkt type: 80",
    detail: None,
    hint: None,
    position: None,
    where_: None,
    schema: None,
    table: None,
    column: None,
    datatype: None,
    constraint: None,
    file: None,
    line: None,
    routine: None
})
```

The error message is from PgBouncer internals (admin.c:1365). Basically, looks like the `'P'` frame isn't handled. This is generated from `postgres_protocol::messages::frontend::parse`.

Any suggestions on how to resolve?

sfackler:
It looks like PGBouncer doesn't support prepared statements. You can use [`batch_execute`](https://docs.rs/postgres/0.15.1/postgres/struct.Connection.html#method.batch_execute), but we don't currently expose a method that returns result rows.

The non-prepared statement API doesn't behave like the prepared statement API, so you won't be able to use the `FromSql` API - you'll just get a bunch of stings back.

jwilm:
> you'll just get a bunch of stings back

How can I do that? `batch_execute` returns `Result<()>`.

sfackler:
This is the (private) method that batch_execute wraps: https://github.com/sfackler/rust-postgres/blob/master/postgres/src/lib.rs#L953

jwilm:
That sounds like the right thing, but as you mentioned, it's private. How do you feel about exposing something like this through the public API?

jwilm:
I tried this out -- simply changed the `batch_execute` API to pass through the Result from `quick_query`. Mostly works for my use case, but it doesn't seem to provide access to _Row description_ (backend frame type `T`).

I'm thinking about adding a method similar to quick_query that would also provide `Vec<Column>` along with the string results. Actually, thinking about this a bit further, it seems that we would be able to provide `Rows` out of such a method given that we have the row description. The only blocker here seems to be the `Arc<StatementInfo>` on `Rows`. It looks to me like `Rows` and `Row` only ever use the `columns` field of `StatementInfo`. Maybe we could just stick that in an `Arc` instead of the whole `StatementInfo`?

To summarize this chain of thought, I'm proposing something like:

1. Add new public query interface like `static_query(query: &str) -> Result<Rows>`
[...]

PR #304, opened 2017-11-21
Implement simple query API

jwilm:
The simple query API is a more robust version of `batch_execute`. Like
that method, `simple_query` allows passing a `&str` of semicolon
delimited queries. Divergence from `batch_execute` is in the return
type; instead of nothing, a `Vec<TextRows>` is returned. Each entry in
this `Vec` is the result set of one query in the query string. Thus if
there are two semicolon delimited queries, there will be two entries in
this `Vec`.

The `TextRows` and `TextRow` types returned from `simple_query` closely
mirror existing `Rows` and `Row` types with one major difference: only
string values can be retrieved from them.

There are a few TODOs in the code:

* Are text values in this case guaranteed to be utf-8 encoded?
* unwrap call in simple_query which assumes RowDescription is always
  sent
* documentation (denoted with either STUB or TODO)

One other question that came up was whether `get` and `get_opt` should handle `null` specially. `TextRow::get_inner` currently returns an `Option<Option<&str>>`, but perhaps an `Option<&str>` can be used with an empty `""` value when the field is null.

This is the API discussed in #303. For reference, this is what was discussed/implemented:

```rust
pub trait GenericConnection {
    // ...

    /// New method on GenericConnection; sends a simple "Query" frame
    ///
    /// Returns `Vec<TextRows>` since there may be multiple, semicolon delimited
    /// queries in the `query` string. From a protocol level, each query yields T, D, C
    /// (row description, data row(s), command completion), and finally a "ready
    /// for query" frame Z is returned.
    fn simple_query(&self, query: &str) -> Result<Vec<TextRows>>;
}

/// Result of one query sent via the `simple_query` API.
struct TextRows {
    /// Result of parsing Row description for query
    columns: Vec<Column>,
    /// Data rows returned in text format
    rows: Vec<RowData>,
}

/// One Row whose data is only available in string format
struct TextRow {
    columns: &[Column],
    raw: &RowData,
}

// Mirror existing Rows/Row APIs where it makes sense
impl IntoIterator for &TextRows {}
impl TextRows {
    fn columns(&self) -> &[Column];
    fn len(&self) -> usize;
    fn is_empty(&self) -> bool;
    fn get(&self, idx) -> TextRow;
    fn iter(&self) -> TextRowIter;
}

impl TextRow {
    fn columns(&self) -> &[Column];
    fn len(&self) -> usize;
    fn is_empty(&self) -> bool;
    fn get<I: RowIndex>(&self, idx: I) -> &str;
    fn get_opt<I: RowIndex>(&self, idx: I) -> Option<&str>;
}
```


jwilm:
@sfackler, no need to review yet, but your input on first two TODOs would be helpful at this point.

sfackler:
You can assume strings are UTF8. You can also assume that the backend will send you the data it's documented to, but I'd return a bad_message error rather than panicking.

jwilm:
Hey @sfackler, I finally got around to polishing this up, and it's ready to review! Thanks for letting this sit idle for so long in the PR queue.

postgres/src/text_rows.rs
```diff
@@ -0,0 +1,180 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Returns a slice describing the columns of the `TextRow`.
+    pub fn columns(&self) -> &[Column] {
+        self.columns
+    }
+
+    /// stub
```
sfackler: Need to finish the docs here.

sfackler:
This seems pretty reasonable to me. Seems like we might want to deprecate/remove batch_execute since this basically replaces it, right?

postgres/src/lib.rs
```diff
@@ -919,6 +926,49 @@ impl InnerConnection {
...
+                }
+                backend::Message::RowDescription(body) => {
+                    columns = Some(self.parse_cols(Some(body))?);
+                }
+                backend::Message::CommandComplete(_) => {
+                    if let Some(cols) = columns.take() {
+                        result.push(TextRows::new(cols, mem::replace(&mut rows, Vec::new())));
+                    }
+                }
+                _ => {}
```
sfackler: Are there other messages we'd expect to get here? Ideally we'd be a bit more defensive and return an unexpected message error like we do elsewhere.

jwilm:
Both of the review items have been resolved. For `batch_execute`, I think it should be deprecated with a message that points to `simple_query` instead. Would you like me to push a commit for that?

jwilm:
Friendly ping in case this slipped past your radar

sfackler:
Sorry for the delay! Yeah, let's deprecate batch_execute.

While you're here, I'd be interested for your thoughts on #346 :)

jwilm:
The deprecation was added in the most recent commit.

jwilm:
Oops, I guess I should fix the tests as well!

jwilm:
Fixed; tests are green now.

PR #318, opened 2018-01-27
Specify required version of optional crates

chinedufn:
Please let me know if you'd prefer this information to be formatted different or go elsewhere in the README!


sfackler:
Thanks!

PR #320, opened 2018-01-29
Add handling of DB-Errors in Notification processing

dunkelstern:
Handle the error that is generated when the DB connection goes down instead of panicing with
an unreachable code error. This allows the notification client to retry establishing the
connection instead of crashing the complete program.

I am very new to rust so please advise if this looks wrong to you.

### DIFF
--- a/postgres/src/notification.rs
+++ b/postgres/src/notification.rs
@@ -3,9 +3,11 @@
 use fallible_iterator::{FallibleIterator, IntoFallibleIterator};
 use std::fmt;
 use std::time::Duration;
-use postgres_protocol::message::backend;
+use postgres_protocol::message::backend::{self, ErrorFields};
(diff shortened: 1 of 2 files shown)


sfackler:
Thanks! I'd like to add some tests to make sure this doesn't regress in the future, but I'm not sure of a great way to inject an error into a connection. Is there any way you can think of? We can merge as-is if not.

sfackler:
Ah, `pg_terminate_backend()` might be able to do it: https://www.postgresql.org/docs/current/static/functions-admin.html

dunkelstern:
I'll try to write a test, will push to this PR

dunkelstern:
Test is committed, seems to run ok on the CI

sfackler:
Thanks!

PR #321, opened 2018-01-30
Fixes for bb8

khuey:
To write bb8 (https://github.com/khuey/bb8) I made a few modifications to tokio_postgres. These do change the exposed interface.

- `Handshake` no longer consumes itself. This wasn't necessary and allows a single `Handshake` to be shared among multiple connections.
- An `is_desynchronized` is added, mirroring the one in sync `postgres`.
- The `TlsMode` enum now holds references, like its equivalent in sync `postgres`. This requires bounding a couple futures to the lifetime of the `TlsMode`. I expect this to be the most controversial part. Since you can still shove your `Handshake` in a `lazy_static!` to get the static lifetime if you really need it I don't think this is a huge issue.

### DIFF
--- a/tokio-postgres/src/lib.rs
+++ b/tokio-postgres/src/lib.rs
@@ -1136,6 +1136,15 @@ impl Connection {
     pub fn parameter(&self, param: &str) -> Option<&str> {
         self.0.parameters.get(param).map(|s| &**s)
     }
+
+    /// Returns whether or not the stream has been desynchronized due to an
+    /// error in the communication channel with the server.
+    ///
+    /// If this has occurred, all further queries will immediately return an
+    /// error.
+    pub fn is_desynchronized(&self) -> bool {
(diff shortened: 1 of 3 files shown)


sfackler:
You can't put a Handshake in a lazy_static if it needs any runtime-derived configuration. I think it may make more sense for bb8 to define a `trait MakeHandshake { fn make_handshake(&self) -> Box<Handshake>; }` instead?

khuey:
If it needs runtime configuration other than environment variables or similar, sure. Do you know of examples of usage that I can look at?

sfackler:
I don't know of any public examples offhand, but it's relatively common for an application to read a config file and initialize things based off of that at startup, including its connection pool.

khuey:
Alright lets just drop those csets then.

khuey:
Poke?

khuey:
ty!

PR #342, opened 2018-04-22
Allow FromSql to borrow from the buffer

### LINKED ISSUE #281: Allow borrowed deserialization
If we changed FromSql to look like

```rust
pub trait FromSql<'a> {
    fn from_sql(ty: &Type, buf: &'a [u8]) -> Result<Self>;
}
```

This would allow implementations for `&str` and `&[u8]` which would be pretty handy. It's unfortunately a breaking change so it'll have to wait

cc @jwilm

sfackler:
This allows for in-place deserialization of text and bytea values in
particular.

Row::get_bytes is removed since it previously existed for this use case.

Closes #281

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
   build:
     working_directory: ~/build
     docker:
-      - image: rust:1.21.0
+      - image: rust:1.23.0
         environment:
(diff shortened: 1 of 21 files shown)


sfackler:
@jwilm

PR #370, opened 2018-08-25
Handle non unix case in poll_start of PollConnect

Lorfdail:
Someone forgot to handle the non unix case for this match which prevented a build on windows.

A similar bit of code already existed here in line 51 to 56:
https://github.com/sfackler/rust-postgres/blob/5ad78500095c0a9a59533b9164f46edd23cb9e1b/tokio-postgres/src/stream.rs


sfackler:
Ah right, thanks!

PR #372, opened 2018-09-25
Allow leaving user at default value in URLs

jcaesar:
(tests were not run…)

### DIFF
--- a/postgres/src/lib.rs
+++ b/postgres/src/lib.rs
@@ -262,15 +262,6 @@ impl InnerConnection {
         let params = params.into_connect_params().map_err(error::connect)?;
         let stream = priv_io::initialize_stream(&params, tls)?;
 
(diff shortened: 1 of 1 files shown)


sfackler:
The user parameter is required, at least according to the Postgres protocol documentation: https://www.postgresql.org/docs/10/static/protocol-message-formats.html.

jcaesar:
Now that you say it, I wonder how I thought this would ever work.

I was trying to reproduce the behaviour of psql when launched with no further parameters. But now I'm not so sure this is a good idea anymore. What would you think about a PR that sets the user to the current system user name if it is not specified?

sfackler:
I'd like to avoid a bunch of platform-specific glue to track down the username, but if there's a clean library that provides this it seems like it'd be okay to use.

PR #373, opened 2018-09-25
Add Json wrapper type

srijs:
The purpose of this is to enable more performant serialization and deserialization that does not need to go via the `Value` type.


sfackler:
Thanks!

PR #374, opened 2018-09-25
Fix 371 savepoint nameing in nested transactions

rustonaut:
Rolling back to a savepoint doesn't release it so,
reusing the same name in nested transaction means
inner transactions only roll back to the 
most transaction/last savepoint.

If the outer most transaction commits but did multiple
consecutive rollbacks in the nested transactions this
did cause an unexpected state in the database before
the fix.


rustonaut:
An alternate solution would be to change the rollback code to do a rollback followed by a release.
But this would get in the way for adding functionality for people allowing the full usage of savepoints.
I.e. adding a `rollback_and_retry` functionality (which rolls back the transaction and restarts it, expect that it's "just" a rollback as we have savepoints and not nested transactions).

PR #376, opened 2018-10-05
Update tokio

jwilm:
This set of patches makes tokio-postgres and tokio-postgres-openssl ~~compatible with~~ utilize re-exports with the latest versions of tokio.

The only testing done on this is to ensure the tests pass locally.

### DIFF
--- a/tokio-postgres-openssl/Cargo.toml
+++ b/tokio-postgres-openssl/Cargo.toml
@@ -7,9 +7,6 @@ authors = ["Steven Fackler <sfackler@gmail.com>"]
 bytes = "0.4"
 futures = "0.1"
 openssl = "0.10"
-tokio-io = "0.1"
+tokio = "0.1.9"
 tokio-openssl = "0.2"
 tokio-postgres = { version = "0.3", path = "../tokio-postgres" }
-
-[dev-dependencies]
-tokio = "0.1.7"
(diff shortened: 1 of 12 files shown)


sfackler:
It should already be compatible with the latest version of tokio - what issues were you seeing?

jwilm:
> It should already be compatible with the latest version of tokio

Ah, this seems this was true. I was reviewing code before starting an integration and noticed this wasn't using the modules via `tokio`. Feel free to close this if you don't want to bump min tokio to 0.1.9.

sfackler:
My understanding is that libraries should avoid depending on tokio directly, but rather the individual bits they need. That way you don't force consumers into depending on tokio proper if they're using e.g. a custom runtime.

jwilm:
I'm a bit confused on that. I was under the impression that modules were originally moved out of tokio and naturally this forced library authors to depend on the module crates instead, but once tokio 0.1 was released, that changed since now the module crates were all reexported and not feature flagged whatsoever.

The README states

>  Library authors should depend on the sub crates, which have greater guarantees of stability.

so perhaps you're right. I don't quite understand how that statement is true, though.

jwilm:
Ah, I get it now. This is all just convenience for apps using the tokio runtime. Sorry for the noise!

Issue #380, opened 2018-10-22
What the reason why tokio_postgres::Client is not deriveing Clone?


andreytkachenko:
Is any specific reason why it not deriveing `Clone`?:

```rust
pub struct Client(proto::Client);
```

Note: `proto::Client` is clonable;

sfackler:
Connections are stateful, so it seems a bit dangerous to have multiple handles to the same one.

andreytkachenko:
To be on the same page, under `Connection` you meant `Client` struct right?
```rust
pub struct Client(proto::Client);
```
and `proto::Client`:
```rust
struct Inner {
    state: Mutex<State>,
    sender: mpsc::UnboundedSender<Request>,
}

#[derive(Clone)]
pub struct Client(Arc<Inner>);
```
not the `Connection` struct:
```rust
pub struct Connection {
    stream: Framed<Box<TlsStream>, PostgresCodec>,
    cancel_data: CancelData,
    parameters: HashMap<String, String>,
    receiver: mpsc::UnboundedReceiver<Request>,
    pending_request: Option<RequestMessages>,
    pending_response: Option<Message>,
    responses: VecDeque<mpsc::Sender<Message>>,
    state: State,
}
```
???

sfackler:
I meant the "idea" of a connection to a postgres database, not the specific `Connection` type, yes.

coder543:
Without `#[derive(Clone)]` on `Client`, I have been unable to prove that it is possible to build and execute a meaningful transaction.

The only two tests that I could find for Transaction currently prepare and execute a single query each. I'm not sure what the use cases are for transactions that only execute one query.

In my case, I need to be able to execute a chain of queries based on previous results.

1. Transaction begins
2. Query A selects from Table 1.
3. Query B selects from Table 2.
4. Queries C...F insert/update/delete into Tables 1, 2, and 3 based on the results from A and B.
5. Transaction is committed (assuming no previous errors)

I need all of these queries to be run inside a single transaction so that if any of them fail, or if the connection fails, or whatever else could happen, it maintains transactional guarantees.

As it is, it is not possible to have a chain of queries, since subsequent queries would be constructed inside an `and_then` (or similar) closure, and require that the Client be `move`d into the closure. Once that happens, it is no longer possible to wrap the outer query future in a transaction. It is not possible to share references to the Client, because mutable access is required for... everything.

By forking this library and adding the `#[derive(Clone)]`, I was able to make everything work just fine.

[...]

PR #384, opened 2018-11-06
Add a TransactionBuilder struct

DarrenTsung:
The TransactionBuilder is useful when trying to create a Transaction
around a future which takes ownership of the client. See doc comment
for more details.

See comments on #380 on why the current `transaction()` API does not 
work, as well as the proposed solution implemented here.

---

This allows us to work with the transaction API with minimal changes, although there is some follow-up discussions around how the async API works with connection pooling. 

Namely, creating a clone of the inner `Client` leaves behind untracked state when it is pooled. I think this is a longer discussion though, as the current API allows the user to create `Query`, `Prepare`, and other futures that don't require the `Client` to be present when later executed.

### DIFF
--- a/tokio-postgres/src/lib.rs
+++ b/tokio-postgres/src/lib.rs
@@ -116,7 +116,17 @@ impl Client {
         // FIXME error type?
         T::Error: From<Error>,
     {
-        Transaction(proto::TransactionFuture::new(self.0.clone(), future))
+        self.transaction_builder().build(future)
+    }
+
+    /// Creates a TransactionBuilder, which can later be used to create
+    /// a Transaction around a future.
+    ///
+    /// Use this when Client is moved into the future being built.
+    /// For example, when executing multiple statements that depend
+    /// on the previous statement's result.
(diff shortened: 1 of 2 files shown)


DarrenTsung:
Not sure what's going on with the tests:
```
error[E0599]: no function or associated item named `from_bits` found for type `f32` in the current scope
  --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/libm-0.1.2/src/math/rem_pio2f.rs:56:13
   |
56 |     tx[0] = f32::from_bits(ix - (e0 << 23) as u32) as f64;
   |             ^^^^^^^^^^^^^^ function or associated item not found in `f32`
   |
   = help: items from traits can only be used if the trait is in scope
   = note: the following trait is implemented but not in scope, perhaps add a `use` for it:
           candidate #1: `use core::num::Float;
```

sfackler:
The compiler we're using for CI is too old - you can bump it up to 1.30 here: https://github.com/sfackler/rust-postgres/blob/master/.circleci/config.yml#L25

PR #385, opened 2018-11-07
add send to TlsStream traits to make sendable for tokio-postgres

Cohedrin:
These were previously implicitly not sendable, though there doesn't seem
to be a good reason for it (that I can see).

If there is a reason that I'm overlooking that would make this a bad idea, please let me know!

### DIFF
--- a/tokio-postgres-openssl/src/lib.rs
+++ b/tokio-postgres-openssl/src/lib.rs
@@ -52,7 +52,7 @@ impl TlsConnect for TlsConnector {
         &self,
         domain: &str,
         socket: Socket,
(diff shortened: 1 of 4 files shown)


sfackler:
This is implied by the bounds on the TlsStream trait itself: https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/src/tls.rs#L61

Cohedrin:
Ah you're right. Thought this was needed due to a lacking trait bound on some code I had written. Looked at it again with what your brought up in mind and was able to resolve it without this. 

Thanks!

Issue #390, opened 2018-11-22
Document tokio_postgres master and add examples


kosta:
Are you interested in a PR? (Not promising anything, just asking)

Issue #117, opened 2015-04-24
Allow sending multiple requests before response


jeff-davis:
I mentioned this at the last rust meetup: I think you can send multiple requests before the response to the first one arrives, i.e. pipelined requests. This can hide network latency when sending many requests when the application doesn't immediately need the response (e.g. INSERT/UPDATE/DELETE without RETURNING).

Again, this is not a fully-formed idea, but I wanted to record it now so it's not forgotten.

sfackler:
This seems like a neat idea. One possible API:
```rust
pub enum BatchRequest<'a> {
    Prepare(&'a str),
    Query(&'a Statement<'a>, &'a [&'a ToSql]),
    Execute(&'a Statement<'a>, &'a [&'a ToSql]),
}

pub enum BatchResponse<'a> {
    Prepare(Statement<'a>),
    Query(Rows<'a>),
    Execute(u32),
    Error(Error),
}

impl Connection {
    pub fn batch_request<'a>(&self, requests: &[BatchRequest<'a>]) -> Vec<BatchResponse<'a>> { ... }
}
```

Everything within a single batch would execute within one transaction, which does mean that if Postgres runs into an error at request *n* statements *n+1* etc won't be processed. So, the result `Vec` would either be the same length as the requests or it would terminate early and the last entry would be a `BatchResponse::Error`.

jeff-davis:
Is there a reason they need to be done as a batch? Why not just single statements, but without waiting for a response?

sfackler:
I was working on an implementation of this but realized that there are some issues that I think will keep it from working for now, unfortunately. Specifically, I'm worried about TCP deadlocks.

Imagine we're making a batched request consisting of a query that will return many rows followed by something else, say some other query. The client sends off the request to execute the first query, and then tries to send the second query immediately afterwards. Let's say that there are some enough parameters in the second query that the whole thing is larger than the sum of the client's outbound TCP buffer and the server's inbound buffer combined. So, it's going to block on write until the server starts reading again. But, the server's going to handle the first query and start returning the rows to the client. If there's enough data to fill the client's inbound TCP buffer and the server's outbound buffer, the server's going to block waiting for the client to read, so we deadlock.

[...]

PR #392, opened 2018-11-22
One example, doing a `select 1;`

kosta:
Is this the intended usage?

Part of #390 

In #117 you said you wanted feedback on the new tokio API. Here it is:

* I was confused by `Connection` no longer having a `connect` method but the "free" `connect` method is quick to find, so that's ok I guess. Can be fixed by documentation if you want to keep this design.
* `connect` returning a Handshake which is a `Future<Item=(Connection, Client)>` is fine. However, what confused the heck out of me was that Connection is a `Future<Item=()>` and basically does nothing user-visible. `Client` does all the things (prepare, query, execute, ...). So first I tried ignoring `Connection` and just calling `Client.prepare()`, which yields a Connection closed error. Weird. Then I tried waiting for `Connection` to resolve `and_then` calling `Client.prepare()`, which block forever. Hm. Then I got the idea to maybe call `connection.join(client...)` and that worked! This is really not intuitive. I'm not deep enough into the architecture to have any say whether this makes sense internally, but at the very least, this needs good documentation. (Maybe the example in this PR is a start)

One word about performance: It rocks! 🥇 

I'm using this to validate some database experiments, and the single-threaded insert performance of tokio_postgres with "perpared statements without (explicit) transactions" is about 4 times that of "chunk 1000 inserts into one transaction" for my use case.  So I'm very happy :D

Let me know what you think, I'm happy to make any adjustments you propose.


sfackler:
The more common approach is to simply spawn off the connection to run in the background: https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/tests/test.rs#L176

kosta:
Makes sense! I will update the example accordingly.

kosta:
I tried to update it to use `runtime.spawn(connection)`, but I see that you removed the `connect()` method. Is there still a way to parse a `postgres://` url to connect? I could only find using the Builder() struct.

Also, does this example make much sense? Should I instead to adopt the "Readme" example for tokio-postgres?

kosta:
I tried to update it to use `runtime.spawn(connection)`, but I see that you removed the `connect()` method. Is there still a way to parse a `postgres://` url to connect? I could only find using the Builder() struct.

Also, does this example make much sense? Should I instead to adopt the "Readme" example for tokio-postgres?

sfackler:
> Is there still a way to parse a `postgres://` url to connect?

Not currently - that's the next bit of functionality I'm planning on working on.

kosta:
Hm, from skimming the commit messages I feel that writing examples _currently_ might be not the best idea while you're doing major rewrites. 

Is it ok to close this? 

Feel free to ping me (on this issue or #390) once the API is in a state where adding examples and documentation makes more sense...

sfackler:
Yeah, I think there's probably enough churn right now that examples will be immediately out of date :)

I'll be sure to ping you when it starts to stabilize! I think we're pretty close. The remaining bits are mostly the conveniences around connecting.

kosta:
Closing.

This example was bad anyway, I'll try to re-create the "person" example from current (non-tokio) release once this stabilizes.

PR #395, opened 2018-12-04
Allow setting user/password via options rather than in host.

jamii:
This is explicitly allowed in
https://www.postgresql.org/docs/10/libpq-connect.html and is supported by
most libraries.


jamii:
As best I can tell this is the same logic as psql/libpg, except that it if you specify the user in two places psql silently picks one and here I decided to return an error instead.

jamii:
Replaced by https://github.com/sfackler/rust-postgres/commit/8dcad81224a0beced3dd3c24609f9fe4bb13332a

PR #401, opened 2018-12-19
Implement  HashSet FromSql ToSql

kingxsp:


sfackler:
I'm not sure I agree with this - sets are not the same thing as arrays.

kingxsp:
Is there any way to store a HashSet?

sfackler:
If you want to store the values as an array in postgres, you can convert to an array in Rust as well. Not sure if there's a native set type off the top of my head.

kingxsp:
yes. but  I just want to make it easy.

PR #405, opened 2018-12-28
add derive(Clone) to tokio_postgres::Client

dicej:
This allows it to be shared among multiple futures.


sfackler:
Connections are stateful objects that shouldn't be shared and used concurrently.

PR #409, opened 2019-01-10
Use relaxed memory ordering for next_statement / next_portal

kosta:
As far as I understood the code, no memory ordering is imposed on the next_statement / next_portal counters, they just need to be unique, so Ordering::Relaxed should be fine.

Rationale for this change:
I'm currently protoyping / benchmarking an application that can process a huge amount of "insert" statements into postgresql, using your tokio-postgres master branch. (Thank you again for implementing pipelining just in time, by the way!). 

I noticed that if I use many connections, my program has much higher throughput if I split it into multiple processes. This indicates some sort of global bottleneck in the hot path. Apart from memory allocations, the only global state (across connections) which I count pinpoint are these two counters.

[Ordering::Relaxed should be fine, at least reddit thinks so](https://www.reddit.com/r/rust/comments/aeii1i/is_memory_orderingrelaxed_sufficient_to_generate/). (I didn't mention rust-postgres in the discussion because I wasn't sure whether I should or whether it's relevant).

With just a pipeline of single "INSERT INTO table VALUES ... ON CONFLICT ... WHERE", I get roughtly the following numbers on my test data set (~10M rows, 8 GB). Note that this is "upserting" live data, so I can't just use COPY_IN:

1 process, 6 connections:
master branch@5b04594: 41884 inserts / second
relaxed branch: 42770 inserts / second
+ ~2% (no real effect, yet)

1 process, 12 connections:
master branch@5b04594: 43589 inserts / second
relaxed branch: 50922 inserts / second
+ ~17%

I also tried to [move the counters into the Client struct](https://github.com/kosta/rust-postgres/commits/no-static-ids), which was even faster for many connections but seemed slower for few connections. I guess that is due to more indirections, so I'm not sure that that branch is the right trade-off.

If you're interested, I will explore some more performance improvements as time permits. I want to make sure network and "disk" bandwidths are the real bottleneck when using this library :)

Let me know that you think.

### DIFF
--- a/tokio-postgres/src/lib.rs
+++ b/tokio-postgres/src/lib.rs
@@ -111,12 +111,12 @@ pub mod types;
 
 fn next_statement() -> String {
     static ID: AtomicUsize = AtomicUsize::new(0);
-    format!("s{}", ID.fetch_add(1, Ordering::SeqCst))
+    format!("s{}", ID.fetch_add(1, Ordering::Relaxed))
 }
 
 fn next_portal() -> String {
     static ID: AtomicUsize = AtomicUsize::new(0);
-    format!("p{}", ID.fetch_add(1, Ordering::SeqCst))
+    format!("p{}", ID.fetch_add(1, Ordering::Relaxed))
 }
 
 /// A convenience function which parses a connection string and connects to the database.


sfackler:
I don't think I really believe that those numbers reflect the cost of differing memory orderings. For performance to improve by 17%, that means that minimally the program was spending 17% of its time performing an atomic addition (as opposed to making system calls, talking to the internet, etc)! x86 already has a highly coherent memory model, so the differences between relaxed and sequentially consistent operations are minimal at best.

kosta:
> For performance to improve by 17%, that means that minimally the program was spending 17% of its time performing an atomic addition

I don't think that math works out. If the performance increase was 100%, did the program spent 100% of its time performing atomic addition? ;)

43589 inserts / s = 22.94 microseconds per insert statement
50922 inserts / s = 19.64 microsecond per insert statement

That means Relaxed would be ~2.3 microseconds faster on average, which is roughly 10% of it's time per request. 

I really am no expert at the performance effects on memory orderings, so I cannot say at all what numbers are realistic and what effects SeqCst has on program optimization etc.

Still, my observation holds: When I create multiple connections, throughput is sub-par compared to creating multiple programs with less connections. 

On the other hand, my "real-world" benchmark is not helpful at all. It probably does too many things and I cannot share it. What about:

1. Micro-benchmarking SeqCst vs Relaxed
2. Benchmarking how fast one can send "SELECT 1;" to postgres with and without this change? (so we have a kinda real-world postgres connection scenario but do not depend on disk speeds etc)

Thanks!
Kosta

sfackler:
> That means Relaxed would be ~2.3 microseconds faster on average, which is roughly 10% of it's time per request.

This is like 3 orders of magnitude longer than an atomic operation takes.

sfackler:
https://rust.godbolt.org/z/Qq4mit

kosta:
That is a very good point, I must admit :)

kosta:
I guess my benchmarks were just noise. I hope I learned from that.

Still, I am pretty sure that there is some bottleneck when using many (e.g. 16) connections in a single process as opposed to using the same number connections across multiple processes. I will see if I can reproduce this independently.

kosta:
Also, thank you for your insights and sorry for wasting your time!

PR #410, opened 2019-01-12
Failing test case for parallel transactions

kosta:
Currently, two transactions started in parallel on the same client interfere with each other. As a first starting point to fix this, I created this failing test case. Once the transactions are properly serialized, it will succeed.

Would this style be ok? If not, please let me know what to improve. (Too many `block_on`? Top deeply nested? etc). 

Do you want to leave this open as a base branch to fix the transaction issue or shall I mark the test as     `#[ignore]` so that you can merge?

Are you even interested in this? If not let me know and I will drop this :) No need to waste any more of your time :)

As always, thanks in advance!

Cheers,
Kosta

### DIFF
--- a/tokio-postgres/tests/test/main.rs
+++ b/tokio-postgres/tests/test/main.rs
@@ -685,6 +685,76 @@ fn transaction_builder_around_moved_client() {
 }
 
 #[test]
+fn parallel_transactions() {
+    use std::sync::{Arc, Mutex};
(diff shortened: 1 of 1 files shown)


sfackler:
I don't think it really makes sense to have tests that don't pass.

After thinking about this more, I don't think there's really anything to do here, other than emphasizing in documentation how this transaction API behaves.

PR #412, opened 2019-01-13
Add legacy 'with-uuid-0.7' feature.

### LINKED ISSUE #411: Release timeline and UUID (again)
The master branch seems to  have seen considerable changes with a tokio-first approach,
and there hasn't been a release in a year.

Is a release of the refactor imminent / on the horizion?

If not, it would be great to have a patch release with a `0.15` branch with for example a `with-uuid-0.7` feature.

theduke:
Note: this obviously can't go into master.

I based it on the `postgres-v0.15.2` tag, it could go into a `postgres-v0.15` branch.

I can do the small release preparation stuff as well for a patch release.

Closes #411

### DIFF
--- a/postgres-shared/Cargo.toml
+++ b/postgres-shared/Cargo.toml
@@ -15,6 +15,7 @@ with-rustc-serialize = ["rustc-serialize"]
 with-serde_json = ["serde_json"]
 with-time = ["time"]
 with-uuid = ["uuid"]
+"with-uuid-0.7" = ["uuid-07"]
 
 [dependencies]
 hex = "0.2"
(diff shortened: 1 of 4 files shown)


sfackler:
This would bring the minimum supported Rust version of 0.15 up to 1.31, which is pretty close to a breaking change.

theduke:
Mhm, this is an interesting case. 

It's somewhat of a grey area since if you are using the feature, you will have `uuid = { version = "0.7" }` dependency and therefore have self-imposed the rustc minimum version already.

sfackler:
Yes. Multi-version dependency support was added in 1.31.

theduke:
Yeah it's probably close to a breaking change.

Let's just wait it out then.

PR #420, opened 2019-02-19
Omit passwords from debug output

TheSandwichMakr:
I can backport if you want, I don't know if you plan on updating the old versions.

### DIFF
--- a/tokio-postgres/src/config.rs
+++ b/tokio-postgres/src/config.rs
@@ -69,45 +69,6 @@ pub(crate) struct Inner {
     pub(crate) target_session_attrs: TargetSessionAttrs,
 }
 
-// Omit password from debug output
-impl fmt::Debug for Inner {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
-            "Inner {{ \
-             user: {:?}, \
-             password: {}, \
-             dbname: {:?}, \
-             options: {:?}, \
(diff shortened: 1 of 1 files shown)


tokio-postgres/src/config.rs
```diff
@@ -69,6 +69,45 @@ pub(crate) struct Inner {
     pub(crate) target_session_attrs: TargetSessionAttrs,
 }
 
+// Omit password from debug output
+impl fmt::Debug for Inner {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
```
sfackler: This should use https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct.

tokio-postgres/src/config.rs
```diff
@@ -69,6 +69,45 @@ pub(crate) struct Inner {
     pub(crate) target_session_attrs: TargetSessionAttrs,
 }
 
+// Omit password from debug output
+impl fmt::Debug for Inner {
```
sfackler: I'd probably impl this directly on Config rather than Inner

tokio-postgres/src/config.rs
TheSandwichMakr: Fixed. I added a struct to show an underscore with debug_struct, to prevent confusion from showing (). Let me know if you just want to use unit instead.

sfackler:
Thanks!

PR #424, opened 2019-03-06
Allow creating new io errors.

khuey:
bb8 needs to create TimedOut errors, ideally in the same type that tokio-postgres uses so I don't have to create a wrapper.


sfackler:
A connection pool timing out trying to retrieve a connection is not the same thing as a client timing out trying to connect to a database. It seems like bb8 should have its own error type here.

Issue #329, opened 2018-03-13
Add ConnectionLike trait for Connection and Transaction in tokio-postgres


asonix:
This would allow for reusable functions that query for data, but can be used on a connection or a transaction.

A basic idea would be the following

```rust
trait ConnectionLike {
    fn batch_execute(self, query: &str) -> Box<Future<Item = Self, Error = (Error, Self)> + Send>;
    fn execute(self, query: &str) -> Box<Future<Item = Self, Error = (Error, Self)> + Send>;
    fn prepare(self, query: &str) -> Box<Future<Item = (Statement, Self), Error = (Error, Self)> + Send>;
    fn execute(self, statement: &Statement, params: &[&ToSql]) -> Box<Future<Item = (u64, Self), Error = (Error, Self)> + Send>;
    fn query(self, statement: &Statement, params: &[&ToSql]) -> Box<Future<Item = Row, State = Self, Error = Error> + Send>;
}

impl ConnectionLike for Connection {
    ...
}

impl ConnectionLike for Transaction {
    ...
}
```
Which could allow for things like the following. This example function, `fetch_events` could be called in production with a Connection, but could be called in tests with a Transaction, allowing for easy rollbacks.
```rust
fn fetch_events<C: ConnectionLike>(conn: C) -> impl Future<Item = (Vec<Event>, C), Error = (Error, C)> {
    conn.prepare("SELECT evt.id, evt.title FROM events AS evt")
        .and_then(move |(s, conn)| {
            conn.query(&s, &[&id])
                .map(move |row| Event {
                    id: row.get(0),
                    title: row.get(1),
                })
                .collect()
        })
}

// in the test (this could likely be extracted into a boilerplate function)
connection
    .transaction()
    .and_then(|transaction| {
        fetch_events(transaction)
        .map_err(|(error, transaction)| {
            transaction.rollback().map(|connection| (error, connection))
        })
        .map(|(events, transaction)| {
            transaction.rollback().map(|connection| (events, connection))
        })

// in the production code
fetch_events(connection)
```

sfackler:
Seems reasonable - we already have this in non-tokio postgres: https://docs.rs/postgres/0.15.2/postgres/trait.GenericConnection.html

enzious:
I have "service" factories that use the GenericConnection trait from 0.15.*. Pretty useful imo.

jsdw:
[...]

PR #433, opened 2019-04-08
Add generic immutable interface.

enzious:
I added a GenericClient trait over Client and Transaction, and replaced &mut with & to support the same style of trait as in previous versions.

I'm using GenericConnection in the old versions and I just wanted to try to add it back. I just named it GenericClient to match the new Client struct.

Thoughts?

Related issue:
#329

### DIFF
--- a/postgres/src/client.rs
+++ b/postgres/src/client.rs
@@ -375,14 +375,10 @@ where
 /// A trait allowing abstraction over connections and transactions
 pub trait GenericClient {
     /// Like `Client::execute`.
-    fn execute<T>(&self, query: &T, params: &[&dyn ToSql]) -> Result<u64, Error>
-    where
-        T: ?Sized + ToStatement;
+    fn execute(&self, query: &str, params: &[&dyn ToSql]) -> Result<u64, Error>;
 
     /// Like `Client::query`.
-    fn query<T>(&self, query: &T, params: &[&dyn ToSql]) -> Result<Vec<Row>, Error>
-    where
-        T: ?Sized + ToStatement;
+    fn query(&self, query: &str, params: &[&dyn ToSql]) -> Result<Vec<Row>, Error>;
 
     /// Like `Client::prepare`.
     fn prepare(&self, query: &str) -> Result<Statement, Error>;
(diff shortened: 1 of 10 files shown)


sfackler:
The methods should take `&mut self`.

enzious:
Could you explain that? It wasn't like this in the previous versions. Is this because of tokio?

In the old version I was using Rc<Connection>, but with this I'll have to use Rc<RefCell<Client>> everywhere.

sfackler:
Internal mutability is very rare in Rust APIs. It only existed in this library because of an old, pre-Rust 1.0 API design that was overhauled significantly in the latest release.

If you want to use internal mutability in your code, you can use a RefCell as you've noted, but a connection pool like r2d2 is possibly a more appropriate choice.

enzious:
r2d2 uses an Arc which is little heavy for what I need.

sfackler:
What percentage of your runtime has your profiling indicated is spent in Arc code?

enzious:
It's a discord/twitch clone (weird I know), but if users are spamming I don't want a ton of Arc locks.

sfackler:
That wasn't really an answer to my question.

enzious:
I haven't profiled this app, but in a previous experience with Arcs it was pretty bad.

sfackler:
Arcs are used in proximity to all kinds of high performance code (and an Arc is not a lock in the first place). I'm not interested in adding a bunch of internal mutability back to this crate.

enzious:
What about a GenericClient trait?

sfackler:
Sure, I'd be happy to add that trait back.

enzious:
And you're right, I was thinking of Arc<Mutex<T>>

colingm:
@sfackler did you ever look at bringing back the GenericConnection trait into the newer versions?

sfackler:
Ah, I have not yet but it seems reasonable to do!

colingm:
@sfackler was there a specific change or re-architecture that prompted its removal? Just wanting to make sure there wasn't a better way to handle that need that I was missing.

sfackler:
The crate was rewritten from scratch, and the trait just never made it back in.

Issue #445, opened 2019-06-20
Can't use `&dyn ToSql` with Async/Await syntax with tokio-postgres


kaibyao:
See https://github.com/kaibyao/tokio-postgres-tosql-multithread-issue for a reference implementation of this problem.

```rust
// example endpoint using actix_web 1.0
async fn index() -> Result<HttpResponse, Error> {
    // ...
    let prepared_values: Vec<&dyn ToSql> = vec![&true];
    let rows = client.query(&statement, &prepared_values).collect().compat().await;
    let rows: Vec<i32> = rows.unwrap().into_iter().map(|row| row.get(0)).collect();

    Ok(HttpResponse::build(StatusCode::OK).json(rows))
}

// in another thread
fn returns_future() -> impl Future01<Item = HttpResponse, Error = Error> {
    index().boxed().compat()
    //      ^^^^^^^  `dyn tokio_postgres::types::ToSql` cannot be shared between threads safely
}
```

What's the best way to pass `Vec<&dyn ToSql>` (for prepared statements) in multi-threaded environments?

The following don't work:
- Using `&(dyn ToSql + Send)` instead of `&dyn ToSql`
- Creating a "wrapper" trait `ToSqlSend` that does have `Send` and `ToSql` and changing `&dyn ToSql` to `&dyn ToSqlSend`.

sfackler:
Are you sure you need specifically a `Vec<&dyn ToSql>`? `client.query(&statement, &[&true])`.

kaibyao:
Yes: it's a simplified example above, but my use case is that I will be sending variables of different types in the prepared statement.

Will I have to fork the library and add `Send` to `ToSql`? I'm still new to rust/systems programming so I'm not exactly sure what the ramifications for this are.

sfackler:
But are you dynamically building the set of values? If not, you can still avoid creating the Vec entirely: `client.query(&statement, &[&foo, &bar, &baz])`.

kaibyao:
Unfortunately I am. The prepared statement is built based on external input at runtime.

sfackler:
Ah awkward. #265 might be a solution there but the ergonomics in the simple case are a bit worse last time I tried.

sfackler:
This should work as a (gross) workaround in the meantime:

```rust
async fn index() -> Result<HttpResponse, Error> {
    // ...
    let future = {
        let prepared_values: Vec<&dyn ToSql> = vec![&true];
[...]

Issue #265, opened 2017-06-08
Take IntoIterator rather than `&[&ToSql]`.


sfackler:
It might be pretty tricky to figure out the exact bounds that will work, but we don't need a slice of `&ToSql` specifically.

elpiel:
Can you give a bit more information, I would like to try and implement this?

PR #446, opened 2019-06-21
Converted ToSql usage to impl IntoIterator

kaibyao:
Addresses #265 and #445. This is about as far as I could get with me being new to Rust.

There's still issues:
- I'm not happy that empty params (`&[]`) has to be explicity cast to `&[some type that implements ToSql (like i32)]`. Example: `&[] as &[i32]`. I'd prefer being able to use the existing method of just sending an empty `&[]`, as that would cause no breakage with existing usage.
- Also, test `test_bpchar_params` fails because I don't know how to work with `None::<&'static str>`.
- I haven't figured out the syntax for sending parameters that contain different types.

@sfackler, could you take a look at this and see how we might be able to fix these things?

### DIFF
--- a/postgres/src/client.rs
+++ b/postgres/src/client.rs
@@ -77,9 +77,15 @@ impl Client {
     /// # Ok(())
     /// # }
     /// ```
-    pub fn execute<T>(&mut self, query: &T, params: &[&dyn ToSql]) -> Result<u64, Error>
+    pub fn execute<T, I>(
+        &mut self,
+        query: &T,
+        params: impl IntoIterator<IntoIter = I, Item = I::Item>,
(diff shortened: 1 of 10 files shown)


sfackler:
The type inference regressions in things like `&[]` is pretty prohibitive on making this change IMO. One option would be to keep the current methods and add new ones that take iterators.

kaibyao:
I agree; closing this for now as it's not ideal.

norcalli:
The solution for the `&[]` empty case is typically to make a utility function which has a concrete type such as `fn empty() -> &'static [u64; 0]` which users can call as `.execute(query, empty())`. Considering 0.16 is still in RC mode, would it be something that could go in that?

PR #478, opened 2019-09-05
document prepared statements should be used for user specified data

toidiu:
Wanted to add this note in the main Readme so its easier to find. From experience, this took a lot of searching.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,12 @@ for (bar, baz) in updates {
 }
 ```
 
+**Note**
+Prepared statements should be use for any query which contains user-specified
+data, as they provided the functionality to safely imbed that data in the request.
+
+More on [prepared statements] (https://en.wikipedia.org/wiki/Prepared_statement).
+
 ### Transactions
 The `transaction` method will start a new transaction. It returns a
 `Transaction` object which has the functionality of a


sfackler:
```rust
let updates = conn.execute("UPDATE foo SET bar = $1 WHERE baz = $2", &[&1i32, &"biz"])?;
```
is just as secure as
```rust
let stmt = conn.prepare("UPDATE foo SET bar = $1 WHERE baz = $2")?;
stmt.execute(&[&1i32, &"biz"])?;
```
but the added paragraph seems to imply that that's not the case.

toidiu:
Ahh, I wasn't sure. Looking at the docs here https://docs.rs/postgres/0.16.0-rc.2/postgres/struct.Client.html#method.query, they only mention that simple_query is not safe. Does that mean all other operations, even `query` is safe?

sfackler:
The unsafe thing is if you manually construct a query string containing user-provided data.

toidiu:
So if I am understanding you correctly, `simple_query_*` are the only unsafe operations on the client.

That wasnt obvious to me from reading the docs/readme. I think it could be useful to mention that somewhere?

Do let me know if you agree and if the Readme or docs or both might be a good location to document that. Also thanks for the quick responses!

sfackler:
`simple_query` and `batch_execute` are the only methods that have warnings about not using them with user-provided parameters: https://docs.rs/postgres/0.16.0-rc.2/postgres/struct.Client.html#warning

toidiu:
got it, that seems sufficient then. thanks

PR #488, opened 2019-09-25
Add channel_binding=disable/prefer/require to config

### LINKED ISSUE #487: Support parameter channel_binding={require|prefer|disable}
A new parameter, ``channel_binding={require|prefer|disable}``, was added to libpq:

https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=d6e612f837e235db0411e8b67558c9a6b3e9f41f

The whole point of channel binding in SCRAM is for the client to authenticate the server, so the implementation shouldn't implicitly trust the server before it's authenticated.

We need to protect against four scenarios:

1. The SCRAM exchange does not begin at all, e.g. the server immediately sends ``AuthenticationOk``.
1. The SCRAM exchange is incomplete, e.g. the server begins a SCRAM exchange, but after the first message sends an ``AuthenticationOk``.
1. The SCRAM exchange does not select a mechanism that supports channel binding.
1. The server sends a totally different authentication mechanism that the client is not expecting that could compromise sensitive information during the authentication process, e.g. the server sends an ``AuthenticationCleartextPassword`` or a ``AuthenticationMD5Password``.

sfackler:
Closes #487

@jeff-davis

### DIFF
--- a/tokio-postgres/src/connect_raw.rs
+++ b/tokio-postgres/src/connect_raw.rs
@@ -255,7 +255,11 @@ where
     let (channel_binding, mechanism) = if has_scram_plus {
         match channel_binding {
             Some(channel_binding) => (channel_binding, sasl::SCRAM_SHA_256_PLUS),
-            None => (sasl::ChannelBinding::unsupported(), sasl::SCRAM_SHA_256),
+            None => {
+                can_skip_channel_binding(config)?;
+
+                (sasl::ChannelBinding::unsupported(), sasl::SCRAM_SHA_256)
+            },
         }
     } else if has_scram {
         can_skip_channel_binding(config)?;
--- a/tokio-postgres/src/connect_raw.rs
+++ b/tokio-postgres/src/connect_raw.rs
@@ -256,14 +256,10 @@ where
         match channel_binding {
             Some(channel_binding) => (channel_binding, sasl::SCRAM_SHA_256_PLUS),
             None => {
-                can_skip_channel_binding(config)?;
-
                 (sasl::ChannelBinding::unsupported(), sasl::SCRAM_SHA_256)
             },
         }
     } else if has_scram {
-        can_skip_channel_binding(config)?;
-
         match channel_binding {
             Some(_) => (sasl::ChannelBinding::unrequested(), sasl::SCRAM_SHA_256),
             None => (sasl::ChannelBinding::unsupported(), sasl::SCRAM_SHA_256),
@@ -271,6 +267,10 @@ where
     } else {
         return Err(Error::authentication("unsupported SASL mechanism".into()));
     };
+
+    if mechanism != sasl::SCRAM_SHA_256_PLUS {
+        can_skip_channel_binding(config)?;
+    }
 
     let mut scram = ScramSha256::new(password, channel_binding);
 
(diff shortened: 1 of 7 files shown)


tokio-postgres/src/connect_raw.rs
```diff
@@ -192,6 +194,16 @@ where
     }
 }
 
+fn no_channel_binding(config: &Config) -> Result<(), Error> {
+    match config.channel_binding {
+        config::ChannelBinding::Disable | config::ChannelBinding::Prefer => Ok(()),
+        config::ChannelBinding::Require => Err(Error::authentication(
+            "server did not use channel binding".into(),
+        )),
+        config::ChannelBinding::__NonExhaustive => unreachable!(),
```
jeff-davis: I don't understand the point of this. Is it a style thing? It seems like it circumvents the exhaustive-match check.

tokio-postgres/src/connect_raw.rs
```diff
@@ -141,8 +141,13 @@ where
     T: AsyncRead + AsyncWrite + Unpin,
 {
     match stream.try_next().await.map_err(Error::io)? {
-        Some(Message::AuthenticationOk) => return Ok(()),
+        Some(Message::AuthenticationOk) => {
+            no_channel_binding(config)?;
+            return Ok(());
+        }
         Some(Message::AuthenticationCleartextPassword) => {
+            no_channel_binding(config)?;
```
jeff-davis: It feels a little fragile to try to check this for each auth method that might leak information to a malicious server. Perhaps check whether the auth request is expected, given the config, at the top so it covers everything?

tokio-postgres/src/connect_raw.rs
```diff
@@ -141,8 +141,13 @@ where
     T: AsyncRead + AsyncWrite + Unpin,
 {
     match stream.try_next().await.map_err(Error::io)? {
-        Some(Message::AuthenticationOk) => return Ok(()),
+        Some(Message::AuthenticationOk) => {
+            no_channel_binding(config)?;
```
jeff-davis: I feel like the name here is confusing and it's a slightly different case than password, md5, etc. Really what you want to know here is "is the client satisfied with the authenticity of the server?". Maybe something like ``check_server_auth()`` or something?

tokio-postgres/src/connect_raw.rs
```diff
@@ -141,8 +141,13 @@ where
     T: AsyncRead + AsyncWrite + Unpin,
 {
     match stream.try_next().await.map_err(Error::io)? {
```
jeff-davis: Should we have a comment here to clarify that this is only dealing with the first message in the authentication sequence? I was slightly confused at first, thinking that this would reject any ``AuthenticationOk`` message if ``channel_binding=require``, but it only rejects it if it's the first in the sequence.

tokio-postgres/src/connect_raw.rs
sfackler: That's the point - we want to be able to add new channel binding modes in the future (if necessary) without breaking backwards compatibility.

tokio-postgres/src/connect_raw.rs
sfackler: It could be handled in a separate match, but I'd be worried about forgetting to update both locations if we add support for a new authentication method.

Neustradamus:
@jeff-davis, @sfackler: Thanks a lot for all about SCRAM.

Linked to:
- https://github.com/scram-xmpp/info/issues/1

PR #513, opened 2019-11-16
Change connection initialization timezone to UTC

aloucks:
Although CockroachDB supports the postgres wire protocol, it doesn't appear to support using the `GMT` timezone (at least not on Windows). However, cockroach does accept `UTC`. This PR provides a way to configure the timezone connection property.

I've left the default as `GMT` but I think `UTC` is probably a better option. The native client defaults to unspecified:

> timezone (string)
> Sets the time zone for displaying and interpreting time stamps. If not explicitly set, the server initializes this variable to the time zone specified by its system environment. See Section 8.5.3 for more information.

https://www.postgresql.org/docs/9.1/runtime-config-client.html

### DIFF
--- a/tokio-postgres/src/connect_raw.rs
+++ b/tokio-postgres/src/connect_raw.rs
@@ -109,7 +109,7 @@ where
     S: AsyncRead + AsyncWrite + Unpin,
     T: AsyncRead + AsyncWrite + Unpin,
 {
-    let mut params = vec![("client_encoding", "UTF8"), ("timezone", "GMT")];
+    let mut params = vec![("client_encoding", "UTF8"), ("timezone", "UTC")];
     if let Some(user) = &config.user {
         params.push(("user", &**user));
     }


sfackler:
The connection configuration options match what libpq exposes. I think you can set this in the `options` key: https://docs.rs/tokio-postgres/0.5.0-alpha.1/tokio_postgres/config/struct.Config.html#method.options.

aloucks:
Using `options` didn't work.  The hard-coded timezone is set when the connection is initialized in `startup`, which is causing the failure.

https://github.com/sfackler/rust-postgres/blob/ba39e2f53b5f1b6967c646a97eb446e0ca09473d/tokio-postgres/src/connect_raw.rs#L107-L112

I verified that libpq does not exhibit this behavior by testing with diesel (using libpq-sys). 

Would you be open to removing the timezone param from the initial startup? I can adjust the PR to to do this instead.

sfackler:
That will break serialization and deserialization of time values.

sfackler:
It might be best to just change the timezone to UTC if cockroach supports that since it's equivalent to GMT.

aloucks:
@sfackler I updated the default to be `UTC`.

sfackler:
Thanks!

PR #515, opened 2019-11-18
Add Hash implementation to Type

bikeshedder:
In order to provide proper caching of statements I need to be able to hash both the `query` and the `parameter_types`. That's why I added `Hash` to the `#[derive(...)]` of `Type` and all other structs it depends on.

### DIFF
--- a/codegen/src/type_gen.rs
+++ b/codegen/src/type_gen.rs
@@ -266,7 +266,7 @@ use std::sync::Arc;
 
 use crate::{{Type, Oid, Kind}};
 
-#[derive(PartialEq, Eq, Debug)]
+#[derive(PartialEq, Eq, Debug, Hash)]
 pub struct Other {{
     pub name: String,
     pub oid: Oid,
@@ -282,7 +282,7 @@ fn make_enum(w: &mut BufWriter<File>, types: &BTreeMap<u32, Type>) {
     write!(
         w,
         "
-#[derive(PartialEq, Eq, Clone, Debug)]
+#[derive(PartialEq, Eq, Clone, Debug, Hash)]
 pub enum Inner {{"
     )
     .unwrap();
--- a/postgres-types/src/lib.rs
+++ b/postgres-types/src/lib.rs
@@ -209,7 +209,7 @@ mod special;
 mod type_gen;
 
 /// A Postgres type.
-#[derive(PartialEq, Eq, Clone, Debug)]
+#[derive(PartialEq, Eq, Clone, Debug, Hash)]
 pub struct Type(Inner);
 
 impl fmt::Display for Type {
@@ -264,7 +264,7 @@ impl Type {
 }
 
 /// Represents the kind of a Postgres type.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub enum Kind {
     /// A simple type like `VARCHAR` or `INTEGER`.
     Simple,
@@ -285,7 +285,7 @@ pub enum Kind {
 }
 
 /// Information about a field of a composite type.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct Field {
     name: String,
     type_: Type,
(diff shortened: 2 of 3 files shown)


postgres-types/src/lib.rs
```diff
@@ -116,7 +116,7 @@ use std::borrow::Cow;
 use std::collections::HashMap;
 use std::error::Error;
 use std::fmt;
-use std::hash::BuildHasher;
+use std::hash::{BuildHasher, Hash};
```
sfackler: I don't think you should need this import.

postgres-types/src/type_gen.rs
```diff
@@ -3,15 +3,15 @@ use std::sync::Arc;
 
 use crate::{Kind, Oid, Type};
 
-#[derive(PartialEq, Eq, Debug)]
+#[derive(PartialEq, Eq, Debug, Hash)]
```
sfackler: This file is generated by the code over in the codegen crate, so you'll want to go over and update that as well.

bikeshedder:
I removed the `Hash` import and added the `Hash` derive to the code generator.

codegen/src/type_gen.rs
```diff
@@ -282,7 +282,7 @@ fn make_enum(w: &mut BufWriter<File>, types: &BTreeMap<u32, Type>) {
     write!(
         w,
         "
-#[derive(PartialEq, Eq, Clone, Debug)]
+#[derive(PartialEq, Eq, Clone, Debug, Hash)]
```
sfackler: Also up above on 269

bikeshedder:
Fixed in the latest version of this PR.

I tried running the code generator but it generates invalid Rust code on my machine. Is there something special involved to run this code successfully so I can check the output?

sfackler:
It should work anywhere. In what way was the code invalid?

bikeshedder:
> It should work anywhere. In what way was the code invalid?

It modifies the code in strange ways:

```diff
diff --git a/postgres-types/src/type_gen.rs b/postgres-types/src/type_gen.rs
index 8156ccf..9924157 100644
--- a/postgres-types/src/type_gen.rs
+++ b/postgres-types/src/type_gen.rs
@@ -1,7 +1,7 @@
 // Autogenerated file - DO NOT EDIT
 use std::sync::Arc;
 
-use crate::{Kind, Oid, Type};
+use crate::{Type, Oid, Kind};
 
 #[derive(PartialEq, Eq, Debug, Hash)]
 pub struct Other {
@@ -127,7 +127,7 @@ pub enum Inner {
     Internal,
     Opaque,
     Anyelement,
-    RecordArray,
+    ,
     Anynonarray,
     TxidSnapshotArray,
     Uuid,
@@ -293,7 +293,7 @@ impl Inner {
             2281 => Some(Inner::Internal),
             2282 => Some(Inner::Opaque),
             2283 => Some(Inner::Anyelement),
-            2287 => Some(Inner::RecordArray),
+            2287 => Some(Inner::),
             2776 => Some(Inner::Anynonarray),
             2949 => Some(Inner::TxidSnapshotArray),
             2950 => Some(Inner::Uuid),
@@ -459,7 +459,7 @@ impl Inner {
             Inner::Internal => 2281,
             Inner::Opaque => 2282,
             Inner::Anyelement => 2283,
-            Inner::RecordArray => 2287,
+            Inner:: => 2287,
             Inner::Anynonarray => 2776,
             Inner::TxidSnapshotArray => 2949,
             Inner::Uuid => 2950,
@@ -511,166 +511,486 @@ impl Inner {
 
...
```

sfackler:
Oh yeah I see that as well - weird. I'll fix it tonight.

bikeshedder:
It would be nice if this PR would make it into the next alpha release. I would like to release `deadpool-postgres` with typed statement cache support which would make it feature complete.

sfackler:
Fixed! https://github.com/sfackler/rust-postgres/commit/a22f63b5bbf21bd584f300c581a92b2f9874c99a

Can you rebase and rerun the codegen? Will merge after that.

codegen/src/type_gen.rs
sfackler: Ping

sfackler: Oh nevermind, it's stored in an Arc

PR #524, opened 2019-12-16
Add back logging of queries

### LINKED ISSUE #492: Missing logging after upgrading
Hey there, I updated from `postgres = "0.15"` and `r2d2_postgres = "0.14"` to 
`postgres = "0.16.0-rc.1"` / `r2d2_postgres = "0.15.0-rc.1"`, and I'm noticing that the logging was removed.

Any chance we could get some of that back? (Preferably via https://github.com/tokio-rs/tracing !)

Specifically log lines like these: https://github.com/sfackler/rust-postgres/blob/postgres-shared-0.4.x/postgres/src/lib.rs#L608

https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/src/query.rs#L33

bikeshedder:
This fixes #492

I also tried adding this piece of code to `Client::execute_raw` but this did not work because it breaks the `check_send` test.

### DIFF
--- a/tokio-postgres/src/query.rs
+++ b/tokio-postgres/src/query.rs
@@ -21,7 +21,17 @@ where
     I: IntoIterator<Item = &'a dyn ToSql>,
     I::IntoIter: ExactSizeIterator,
 {
-    let buf = encode(client, &statement, params)?;
+    let buf = if log::log_enabled!(log::Level::Debug) {
+        let params = params.into_iter().collect::<Vec<_>>();
+        log::debug!(
+            "executing statement {} with parameters: {:?}",
+            statement.name(),
+            params,
+        );
+        encode(client, &statement, params)?
+    } else {
+        encode(client, &statement, params)?
+    };
     let responses = start(client, buf).await?;
     Ok(RowStream {
         statement,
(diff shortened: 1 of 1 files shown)


sfackler:
Oh, nice call!

tokio-postgres/src/query.rs
```diff
@@ -21,7 +21,17 @@ where
...
 {
-    let buf = encode(client, &statement, params)?;
+    let buf = if log::log_enabled!(log::Level::Debug) {
+        let params = params.into_iter().collect::<Vec<_>>();
+        log::debug!(
+            "executing statement {} with parameters: {:?}",
+            statement.name(),
+            params,
+        );
+        encode(client, &statement, params)?
```
sfackler: Same here.

tokio-postgres/src/query.rs
```diff
@@ -59,7 +69,17 @@ where
...
```
sfackler: Can you run the params through [`slice_iter`](https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/src/lib.rs#L229) here? That should minimize extra instantiations of encode.

sfackler: Nevermind, will require a bit more work to make that happen. I'll deal with it on my end.

PR #525, opened 2019-12-21
Adding in generic connection trait back to the lib

colingm:
I am trying to add back in the generic connection trait that existed for connections and traits in version 0.15.2 but was not put back in when this lib was rebuilt. This work is based on an existing closed PR: https://github.com/sfackler/rust-postgres/pull/433. I wasn't exactly sure where to put the generic connection so I added it in a new file but it could easily be moved to the lib.rs or even client.rs. I also used the old name of GenericConnection instead of GenericClient used in the old PR. All available feedback is requested.

### DIFF
--- a/postgres/src/client.rs
+++ b/postgres/src/client.rs
@@ -1,4 +1,7 @@
-use crate::{Config, CopyInWriter, CopyOutReader, GenericConnection, RowIter, Statement, ToStatement, Transaction};
+use crate::{
+    Config, CopyInWriter, CopyOutReader, GenericConnection, RowIter, Statement, ToStatement,
+    Transaction,
+};
 use std::ops::{Deref, DerefMut};
 use tokio::runtime::Runtime;
 use tokio_postgres::tls::{MakeTlsConnect, TlsConnect};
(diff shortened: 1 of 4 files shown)


postgres/src/client.rs
```diff
@@ -450,3 +453,18 @@ impl Client {
...
 }
+
+impl GenericConnection for Client {
+    fn execute(&mut self, query: &str, params: &[&(dyn ToSql + Sync)]) -> Result<u64, Error> {
+        self.execute(query, params)
+    }
+    fn query(&mut self, query: &str, params: &[&(dyn ToSql + Sync)]) -> Result<Vec<Row>, Error> {
+        self.query(query, params)
+    }
+    fn prepare(&mut self, query: &str) -> Result<Statement, Error> {
```
sfackler: It seems strange that you can prepare a statement but not use it through this interface.

colingm: Whoops sorry I didn't even realize that query (and execute) had a generic, I have updated them to include the same trait and generic as the normal implementations

sfackler: That does mean that the trait isn't object safe, if that's relevant for what you're trying to do with it.

PR #530, opened 2019-12-25
Remove debugging printlns from copy out protocol

benesch:
@sfackler looks like these just slipped in from a debugging session?


sfackler:
Oops!

PR #531, opened 2019-12-25
Support cancellation in synchronous client

benesch:
The tl;dr is that it'd be nice to be able to send cancellation requests when using the synchronous client. This is a first pass at implementing that. Not sure if it's mergeable in its current form, as it relies on exposing some private bits of `tokio-postgres` for the sole use of the `postgres` crate, and then marking those bits as `doc(hidden)`.

### DIFF
--- a/tokio-postgres/src/client.rs
+++ b/tokio-postgres/src/client.rs
@@ -1,5 +1,3 @@
-#[cfg(feature = "runtime")]
-use crate::cancel_query;
 use crate::codec::BackendMessages;
 use crate::config::{Host, SslMode};
 use crate::connection::{Request, RequestMessages};
@@ -14,7 +12,7 @@ use crate::to_statement::ToStatement;
 use crate::types::{Oid, ToSql, Type};
 #[cfg(feature = "runtime")]
 use crate::Socket;
-use crate::{cancel_query_raw, copy_in, copy_out, query, CopyInSink, Transaction};
+use crate::{copy_in, copy_out, query, CancelToken, CopyInSink, Transaction};
 use crate::{prepare, SimpleQueryMessage};
 use crate::{simple_query, Row};
 use crate::{Error, Statement};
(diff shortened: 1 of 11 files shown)


tokio-postgres/src/client.rs
```diff
@@ -451,6 +450,18 @@ impl Client {
         Ok(Transaction::new(self))
     }
 
+    /// Not part of the public API of this crate.
+    #[doc(hidden)]
+    pub fn cancel_token(&self) -> CancelToken {
```
sfackler: I've tried to avoid having private APIs in tokio-postgres just for postgres's sake. I think it's reasonable for this to just be a public API though. I wonder if we should just deprecate cancel_query and cancel_query_raw?

benesch: That certainly works for me, if it works for you! The only potential downside is that the current (async-only) API can guarantee at compile time that the connection you're sending the cancellation request on is still alive. The `CancelToken`, of course, can outlive the connection it was created from. Given how racy cancellation is in general, I don't think this is a big deal, but wanted to raise it nonetheless.

sfackler: It cannot guarantee that. `cancel_query` is defined on `Client`, not `Connection`.

benesch: Oh, good point, I forgot that the client lifetime wasn't tied to the connection lifetime. 👍

benesch:
Ok, updated to deprecate the `tokio_postgres::Client::cancel_query[_raw]` API and publicize the new `CancelToken`-based API in both tokio_postgres and postgres. Thoughts?

tokio-postgres/src/lib.rs
```diff
@@ -125,6 +126,8 @@ mod bind;
 #[cfg(feature = "runtime")]
 mod cancel_query;
 mod cancel_query_raw;
+#[doc(hidden)]
```
sfackler:
```suggestion
```

tokio-postgres/src/lib.rs
```diff
@@ -125,6 +126,8 @@ mod bind;
...
+pub mod cancel_token;
```
sfackler: This can be private.

tokio-postgres/tests/test/main.rs
```diff
@@ -304,6 +304,7 @@ async fn cancel_query_raw() {
     let client = connect("user=postgres").await;
 
     let socket = TcpStream::connect("127.0.0.1:5433").await.unwrap();
+    #[allow(deprecated)]
     let cancel = client.cancel_query_raw(socket, NoTls);
```
sfackler: Let's just switch these tests over to use the new APIs.

tokio-postgres/src/lib.rs
benesch: Oops, thanks.

tokio-postgres/tests/test/main.rs
benesch: You bet. Done.

sfackler:
Thanks!

PR #535, opened 2019-12-29
Expose accessors on Config object

### LINKED ISSUE #534: Reading values out of postgres::Config
Back with postgres 0.15, I had the following hunk of logic to emulate libpq's behavior around defaults for connection parameters:

```rust
let params = url.into_connect_params();

// Fill in default values for parameters using the same logic that libpq
// uses, which involves reading various environment variables.
let params = {
    let mut new_params = postgres::Config::new();

    if let Some(user) = params.user() {
        let password = user
            .password()
            .owned()
            .or_else(|| env::var("PGPASSWORD").ok());
        new_params.user(user.name(), password.as_deref());
    } else {
        let name = env::var("PGUSER").unwrap_or_else(|_| whoami::username());
        let password = env::var("PGPASSWORD").ok();
        new_params.user(&name, password.as_deref());
    }

    if let Some(database) = params
        .database()
        .owned()
        .or_else(|| env::var("PGDATABASE").ok())
    {
        new_params.database(&database);
    }
    new_params.port(params.port());

    let mut host = params.host().clone();
    if let postgres::params::Host::Tcp(hostname) = &host {
        if hostname == "" {
            let host_str = env::var("PGHOST").unwrap_or_else(|_| "localhost".into());
            host = postgres::params::Host::Tcp(host_str);
        }
    }

    new_params.build(host)
};
```

This was possible because of the separation between `postgres::params::ConnectParams` and `postgres::params::Builder`; in postgres 0.17 only the builder exists, so there's no way to introspect the state of a `postgres::Config`.

Would it be possible to bring back some form of introspection for `postgres::Config`?

benesch:
Fix #534.

### DIFF
--- a/tokio-postgres/src/config.rs
+++ b/tokio-postgres/src/config.rs
@@ -56,9 +56,14 @@ pub enum ChannelBinding {
     Require,
 }
 
+/// A host specification.
 #[derive(Debug, Clone, PartialEq)]
-pub(crate) enum Host {
+pub enum Host {
+    /// A TCP hostname.
     Tcp(String),
+    /// A path to a directory containing the server's Unix socket.
+    ///
+    /// This variant is only available on Unix platforms.
     #[cfg(unix)]
     Unix(PathBuf),
 }
@@ -190,6 +195,12 @@ impl Config {
         self
     }
 
+    /// Gets the user to authenticate with, if one has been configured with
+    /// the `user` method.
+    pub fn get_user(&self) -> Option<&str> {
+        self.user.as_deref()
+    }
+
     /// Sets the password to authenticate with.
     pub fn password<T>(&mut self, password: T) -> &mut Config
     where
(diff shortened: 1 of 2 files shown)


benesch:
@sfackler re https://github.com/sfackler/rust-postgres/issues/534#issuecomment-569529222 I wound up exposing the `Host` enum, but lmk if you'd like me to do something different.

tokio-postgres/src/config.rs
```diff
@@ -56,9 +56,12 @@ pub enum ChannelBinding {
     Require,
 }
 
+/// A host specification.
 #[derive(Debug, Clone, PartialEq)]
-pub(crate) enum Host {
+pub enum Host {
+    /// A TCP hostname.
     Tcp(String),
+    /// A path to a directory containing the server's Unix socket.
```
sfackler: Can you add a note that this variant is only available on Unix platforms?

sfackler:
Yeah on second thought I think just exposing Host is fine.

tokio-postgres/src/config.rs
benesch: Yep, done!

sfackler:
Thanks!

PR #536, opened 2019-12-29
Put unused type parameter back into use

nolanderc:
A simple fix to the transaction API. Not sure why there wasn't any warning from the compiler.


sfackler:
Thanks!

PR #538, opened 2019-12-30
Add Debug impls.

derekdreery:
I added/modified some of the `Debug` implementations to help me make sense of why some of my queries didn't work as expected (spoiler alert: I had semantic errors in the queries). I thought these additions might be useful to the community.

This commit makes the following changes

 - Add an opaque `Debug` impl for `Client`.
 - Add a rich `Debug` impl for `Row`.
 - Make the `Debug` impl for `Type` clearer.
 - Change the `Debug` for `Column` to be slightly neater.

### DIFF
--- a/tokio-postgres/src/row.rs
+++ b/tokio-postgres/src/row.rs
@@ -124,6 +124,7 @@ impl fmt::Debug for Row {
         struct RowData<'a>(&'a Row);
 
         impl fmt::Debug for RowData<'_> {
+            #[allow(clippy::match_ref_pats)]
             fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
                 let mut map = f.debug_map();
                 for (idx, col) in self.0.columns().iter().enumerate() {
--- a/tokio-postgres/src/row.rs
+++ b/tokio-postgres/src/row.rs
@@ -100,59 +100,10 @@ pub struct Row {
     ranges: Vec<Option<Range<usize>>>,
 }
 
-/// A macro to map pg types to rust types, for debug display.
-macro_rules! debug_row_type {
-    ($this:expr; $map:expr; $idx:expr; $name:expr; $type:expr; $($pg_ty:tt => $ty:ty),*) => {
-        match $type {
-            $(
-                &Type::$pg_ty => match <$ty as FromSql>::from_sql_nullable(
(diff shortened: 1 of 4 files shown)


tokio-postgres/src/row.rs
```diff
@@ -100,6 +100,63 @@ pub struct Row {
...
+                    Ok(val) => $map.entry(&$name, &val),
+                    Err(_) => $map.entry(&$name, &"<error>"),
+                },
+            )*
+            _ => $map.entry(&$name, &"<opaque type>"),
+        }
+    }
+}
+
+impl fmt::Debug for Row {
```
sfackler: This is a lot of code for a debug impl.

derekdreery: Yes it's a little bit complicated. The reason is that it inspects the postgres type of the column value, and then chooses an appropriate debuggable rust type to convert into, if possible. It gives output like the following:

```text
Row {
    columns: [
        Column {
            name: "id",
            type: Int4,
        },
        Column {
            name: "created_at",
            type: Timestamptz,
        },
        Column {
            name: "message",
            type: Varchar,
        },
        Column {
            name: "email",
            type: Varchar,
        },
        Column {
            name: "name",
            type: Varchar,
        },
    ],
    data: {
        "id": 1,
        "created_at": SystemTime {
            tv_sec: 1577464626,
            tv_nsec: 0,
        },
        "message": "test",
        "email": "test",
        "name": "test",
    },
}
```

derekdreery: I only did a few of the types where I knew what rust type to use, the rest will return `<opaque type>`.

sfackler: I'd rather not try to downcast the data for some subset of types. Just displaying the column information seems good enough IMO.

derekdreery: I'll make the change you request. Still, seems a shame to not make use of the extra data.

derekdreery:
Removed value-debug and rebased.

derekdreery:
Rebased again. Could you let me know if you're unlikely to accept this PR?

sfackler:
Merged, sorry for the delay!

derekdreery:
No worries, sorry if I was pushy :)

PR #542, opened 2020-01-03
Add tokio_postgres::GenericClient

### LINKED ISSUE #357: Support SCRAM channel binding
We'll need to extend the TLS interface to provide access to the Finished message and the peer certificate.

### LINKED ISSUE #537: Adding a generic connection trait to tokio-postgres
PR #525 adds a synchronous trait into rust-postgres for synchronous connections, I believe adding a corresponding trait into tokio-postgres would be benefitial. However, since async methods are not yet supported in traits we have to work around it using dynamic dispatch, such as is made possible by https://github.com/dtolnay/async-trait.

I have implemented a limited version in my own project, and would be happy to add a more full-featured version into the main library, if you'd be fine with such a workaround.

sfackler:
We have to make the trait methods differ from the normal methods a bit
by adding Sync + Send bounds to the ToStatement parameter which is a bit
unfortunate, but necessary until GATs async_trait unnecessary.

Closes #537


benesch:
s/Closes #357/Closes #537/, I think?

sfackler:
Yep, thanks!

PR #561, opened 2020-01-27
Add missing methods to GenericClient

dvic:
This adds `execute_raw`,  `query_one`, `query_opt`, `query_raw`, and `prepare_typed` to the generic trait introduced in #525.

Furthermore, in 1ea8b7b2d467b2da49c94de522dd6ec054b5d159 I attempted to fix the lifetimes (as `&mut` is only required for `transaction()`).

@sfackler I'm not sure if this is the way to go or if you want this in a separate PR, if so, please let me know.

### DIFF
--- a/tokio-postgres/src/client.rs
+++ b/tokio-postgres/src/client.rs
@@ -506,6 +506,7 @@ impl Client {
 
 #[async_trait]
 impl GenericClient for Client {
+    /// Like `Client::execute`.
     async fn execute<T>(&mut self, query: &T, params: &[&(dyn ToSql + Sync)]) -> Result<u64, Error>
     where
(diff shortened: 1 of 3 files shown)


tokio-postgres/src/client.rs
```diff
@@ -506,28 +506,81 @@ impl Client {
 
 #[async_trait]
 impl GenericClient for Client {
-    async fn execute<T>(&mut self, query: &T, params: &[&(dyn ToSql + Sync)]) -> Result<u64, Error>
+    /// Like `Client::execute`.
```
sfackler: Documentation isn't necessary in the implementation of a trait.

sfackler:
Thanks!

PR #562, opened 2020-01-27
Remove Sync and Send requirements in GenericClient

dvic:
I just found out that the `async_trait` macro supports leaving out the `Sync` and `Send` requirements 🙂, see https://github.com/dtolnay/async-trait#non-threadsafe-futures.

### DIFF
--- a/tokio-postgres/src/client.rs
+++ b/tokio-postgres/src/client.rs
@@ -504,19 +504,19 @@ impl Client {
     }
 }
 
-#[async_trait]
+#[async_trait(?Send)]
 impl GenericClient for Client {
     async fn execute<T>(&self, query: &T, params: &[&(dyn ToSql + Sync)]) -> Result<u64, Error>
     where
-        T: ?Sized + ToStatement + Sync + Send,
+        T: ?Sized + ToStatement,
     {
         self.execute(query, params).await
     }
 
     async fn execute_raw<'b, I, T>(&self, statement: &T, params: I) -> Result<u64, Error>
     where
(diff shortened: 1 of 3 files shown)


sfackler:
Won't this prevent the client from being used in any context that does require Send futures?

dvic:
Ah you're right, this is a part of what the macro produces:
```rust
    impl GenericClient for Client {
        fn execute<'life0, 'life1, 'life2, 'life3, 'async_trait, T>(
            &'life0 self,
            query: &'life1 T,
            params: &'life2 [&'life3 (dyn ToSql + Sync)],
        ) -> ::core::pin::Pin<
            Box<dyn ::core::future::Future<Output = Result<u64, Error>> + 'async_trait>,
        >
        where
            T: ?Sized + ToStatement,
            T: 'async_trait,
            'life0: 'async_trait,
            'life1: 'async_trait,
            'life2: 'async_trait,
            'life3: 'async_trait,
            Self: 'async_trait,
        {
            #[allow(clippy::missing_docs_in_private_items, clippy::used_underscore_binding)]
            async fn __execute<T>(
                _self: &Client,
                query: &T,
                params: &[&(dyn ToSql + Sync)],
            ) -> Result<u64, Error>
            where
                (): Sized,
                T: ?Sized + ToStatement,
            {
                _self.execute(query, params).await
            }
            Box::pin(__execute::<T>(self, query, params))
        }
}
```

So `Self` is not required to be `Sync` but the `Send` property is also gone on the resulting future. We'll have to wait a little longer to get rid of these 😉.

PR #564, opened 2020-01-31
Don't suppress notices during startup flow

benesch:
NoticeResponses received during the startup flow were previously being
dropped on the floor. Instead stash them away so they can be delivered
to the user after the startup flow is complete.


sfackler:
Thanks!

Issue #366, opened 2018-07-18
Allow reading composite values into a Row-like struct


kostrowski:
It would be useful to be able to treat the value of a composite-value column as a `Row`.  Presently, to read composite values you need to use `postgres-derive` or handle the raw bytes returned by `Row::get_bytes(...)`.  In certain situations neither option is attractive.   Using `postgres-derive` requires one to define a struct matching the fields of the composite value and results in parsing based on name matching, which could be a performance issue.  Manually handling the raw bytes returned by `Row::get_bytes(...)` could avoid the performance issue, but that would mean duplicating a fair amount of code that is already in `postgres-protocol` in the form of `message::DataRowBody `and `message::DataRowRanges`.

I'm not entirely certain of the best approach within the library (a simple hack is to make `DataRowBody` and `DataRowRanges` externally constructible and feed the field's raw bytes into them).

An example use case would be in working with class-table inheritance, where you might want to query the concrete type variants as composites:

```rust
let sql = "\
    SELECT widget.type, widget.name, machine, toy \
    FROM widget LEFT JOIN machine USING (id) LEFT JOIN toy USING (id) \
";

for row in &conn.query(sql, &[]).unwrap() {
    let ty = str::from_utf8(row.get_bytes(0).unwrap()).unwrap();
    let name = row.get(1);
    let kind = match ty {
        "machine" => {
            let row: Composite = row.get(2);
            WidgetKind::Machine {
                gears: row.get(0),
                operator: row.get(1),
            }
        },
        "toy" => {
            let row: Composite = row.get(3);
            WidgetKind::Toy {
                minimum_age: row.get(0),
            }
        },
    };
    let widget = Widget { name, kind };
    println!("{:?}", widget);
}
```

sfackler:
Sure, this seems pretty reasonable.

PR #565, opened 2020-01-31
Add a Row-like interface to any PostgreSQL composite type

alexwl:
It can be useful to be able to get fields of a composite type without the need to derive `FromSql` for a structure (https://github.com/sfackler/rust-postgres/issues/366).

This PR adds a new type `CompositeType` that implements `FromSql` and has the same methods as `Row`.

Example:

```sql
create table users (
  id               serial primary key,
  name             text
);
create table posts (
  id               serial primary key,
  author_id        integer not null references users(id),
  title            text,
  content          text
);
```
```rust
let query = "SELECT users.*,ARRAY_REMOVE(ARRAY_AGG(posts),NULL) as posts FROM users LEFT JOIN posts ON posts.author_id = users.id GROUP BY users.id";
let stmt = client.prepare(query).await?;
let rows: Vec<Row> = client.query(&stmt, &[]).await?;
for user_with_posts in rows {
    let user_name = user_with_posts.get::<_,&str>("name");
    // ...
    let user_posts = user_with_posts.get::<_,Vec<CompositeType>>("posts");
    for post in user_posts {
        let post_title = post.get::<_,&str>("title");
        let post_content = post.get::<_,&str>("content");
        // ...
    }
}
```
 

The binary format of a composite type is different from the format of a `DataRow` (there is an OID before each length/value pair), so I added a type parameter to `DataRowRanges` to make `FallibleIterator` for ranges usable for both cases:
https://github.com/alexwl/rust-postgres/blob/aa4b5547e3bf13e112ae90dd919b0ac70d67ebaa/postgres-protocol/src/message/backend.rs#L550-L574

### DIFF
--- a/postgres-protocol/src/message/backend.rs
+++ b/postgres-protocol/src/message/backend.rs
@@ -6,6 +6,7 @@ use fallible_iterator::FallibleIterator;
 use memchr::memchr;
 use std::cmp;
 use std::io::{self, Read};
+use std::marker::PhantomData;
 use std::ops::Range;
 use std::str;
 
@@ -531,11 +532,12 @@ pub struct DataRowBody {
 
 impl DataRowBody {
     #[inline]
-    pub fn ranges(&self) -> DataRowRanges<'_> {
+    pub fn ranges(&self) -> DataRowRanges<'_, DataRow> {
         DataRowRanges {
             buf: &self.storage,
             len: self.storage.len(),
             remaining: self.len,
+            phantom: PhantomData,
         }
     }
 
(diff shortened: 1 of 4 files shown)


postgres-protocol/src/message/backend.rs
```diff
@@ -531,11 +532,12 @@ pub struct DataRowBody {
 
 impl DataRowBody {
     #[inline]
-    pub fn ranges(&self) -> DataRowRanges<'_> {
+    pub fn ranges(&self) -> DataRowRanges<'_, DataRow> {
```
sfackler: This module is specifically for parsing the backend protocol messages. I'd rather have composite support added in the types module.

tokio-postgres/src/row.rs
```diff
@@ -175,6 +182,134 @@ impl Row {
...
+
+fn composite_type_fields(type_: &Type) -> &[Field] {
+    match type_.kind() {
+        Kind::Composite(ref fields) => fields,
+        _ => unreachable!(),
+    }
+}
+
+impl<'a> CompositeType<'a> {
+    pub(crate) fn new(
```
sfackler: It seems like this can just be inlined in to the FromSql implementation.

alexwl:
I moved the code from `CompositeType::new` to `FromSql::from_sql` and moved the iterator over the fields of a composite type to `postgres-protocol/src/types/mod.rs`.

runfalk:
Is there something blocking this? It is something I'm quite interested in myself.

alexwl:
I'm also still interested in this. I can update the PR if necessary.

mpalmer:
I've rebased this PR against current `master`, and it applied successfully (`patch` seemed to find the right offsets).  I'm now using it in a project, and so far it seems to working great, so I'd be :+1: for merge.  One thing that seemed a bit odd was how `CompositeType` was in `tokio-postgres`, rather than, say, `postgres-types`, as I can't see any async-specific code in the implementation.

phillipleblanc:
If you need this before its merged, you can also just copy the structs from this PR into your codebase to use it without needing to fork this library. An example of this can be found here: https://github.com/spiceai/spiceai/blob/trunk/crates/arrow_sql_gen/src/postgres/composite.rs

PR #566, opened 2020-02-02
Expose async notifications in sync API

### LINKED ISSUE #404: How should notifications be handled in the sync API?

benesch:
@sfackler not sure if there are additional complexities here, but this callback is sufficient for my notification-receiving needs with the sync client. Consider this more of a proposal than a ready-to-merge PR!

----

Expose a Config::connect_with method in the sync API that allows the
user to provide a callback function that is invoked for every
asynchronous message or error received by the underlying connection.

Fix #404.

### DIFF
--- a/postgres/src/config.rs
+++ b/postgres/src/config.rs
@@ -2,8 +2,9 @@
 //!
 //! Requires the `runtime` Cargo feature (enabled by default).
 
-use crate::Client;
-use futures::FutureExt;
+use crate::{AsyncMessage, Client};
+use futures::future;
+use futures::stream::{self, StreamExt};
 use log::error;
 use std::fmt;
 use std::path::Path;
@@ -88,9 +89,9 @@ use tokio_postgres::{Error, Socket};
 /// ```not_rust
 /// postgresql:///mydb?user=user&host=/var/lib/postgresql
 /// ```
-#[derive(Clone)]
 pub struct Config {
     config: tokio_postgres::Config,
+    notification_callback: Box<dyn NotificationCallback>,
 }
 
 impl fmt::Debug for Config {
@@ -98,6 +99,15 @@ impl fmt::Debug for Config {
         fmt.debug_struct("Config")
             .field("config", &self.config)
             .finish()
+    }
+}
+
+impl Clone for Config {
+    fn clone(&self) -> Config {
+        Config {
+            config: self.config.clone(),
+            notification_callback: self.notification_callback.box_clone(),
+        }
     }
 }
 
@@ -112,6 +122,7 @@ impl Config {
     pub fn new() -> Config {
         Config {
             config: tokio_postgres::Config::new(),
+            notification_callback: Box::new(|_| ()),
         }
     }
 
@@ -308,6 +319,23 @@ impl Config {
         self.config.get_channel_binding()
     }
 
+    /// Sets the notification callback.
+    ///
+    /// This callback will be invoked with every [`AsyncMessage`]s that is
+    /// received over the connection.
+    pub fn notification_callback<F>(&mut self, f: F) -> &mut Config
+    where
+        F: Fn(AsyncMessage) + Clone + Send + 'static,
+    {
+        self.notification_callback = Box::new(f);
+        self
+    }
+
+    /// Gets the current notification callback.
+    pub fn get_notification_callback(&self) -> &(dyn FnMut(AsyncMessage) + Send) {
+        self.notification_callback.as_inner()
+    }
+
     /// Opens a connection to a PostgreSQL database.
     pub fn connect<T>(&self, tls: T) -> Result<Client, Error>
     where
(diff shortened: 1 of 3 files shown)


postgres/src/config.rs
```diff
@@ -308,32 +309,51 @@ impl Config {
...
     }
 
-    /// Opens a connection to a PostgreSQL database.
-    pub fn connect<T>(&self, tls: T) -> Result<Client, Error>
+    /// Like `connect`, but `notify_fn` will be invoked with any asynchronous
+    /// notifications or connection errors that arise.
+    ///
+    /// `notify_fn` is invoked from an asynchronous runtime and therefore must
+    /// not block.
+    pub fn connect_with<T, F>(&self, tls: T, mut notify_fn: F) -> Result<Client, Error>
```
sfackler: Rather than a separate method, I think it would make more sense to store the callback in the config with a setter.

postgres/src/config.rs
```diff
@@ -308,32 +309,51 @@ impl Config {
...
     where
         T: MakeTlsConnect<Socket> + 'static + Send,
         T::TlsConnect: Send,
         T::Stream: Send,
         <T::TlsConnect as TlsConnect<Socket>>::Future: Send,
+        F: FnMut(Result<AsyncMessage, Error>) + Send + 'static,
```
sfackler: I'd rather it take a `AsyncMessage` rather than a `Result`.

benesch: Sure, done.

postgres/src/config.rs
benesch: How do you propose to store the closure in the config? I see two options:

  * An `fn (AsyncFunction)` function pointer. This is almost certainly too limiting since you can't pass a closure that captures any state. (It's definitely too limiting for my uses.)
  * A `Box<dyn FnMut(AsyncMessage) + Clone + Send + 'static>`. This doesn't work as written, of course, because `Clone` isn't an auto trait, but I think you could make it work by creating a child trait that inherits from `Clone` and `FnMut(AsyncMessage)`? (`Clone` is necessary to preserve the `Clone`ness of `Config`). This is also potentially limiting, if the closure wants to capture a non-`Clone`able type.

sfackler: I'd go with an `Arc<dyn Fn(AsyncMessage) + Sync + Send>`.

benesch: Forcing a `Sync` bound on the closure is tough, though. For example it precludes the test that I wrote which sends the messages over an MPSC (which I imagine is what 90% of these implementations will do) because an MPSC transmitter is not `Sync`.  And putting an MPSC transmitter in a `Mutex` feels really weird, especially because we know that each closure is only going to be invoked by one thread.

I've pushed up an implementation that instead uses a `Box<NotificationCallback>`, where `NotificationCallback` is a custom private trait that's implemented for any `FnMut(AsyncMessage) + Send + Clone`. From the user's perspective the API is

```rust
    pub fn notification_callback<F>(&mut self, f: F) -> &mut Config
    where
        F: Fn(AsyncMessage) + Clone + Send + 'static,
    {
```

which feels quite nice. A bit of ugly boilerplate internally though.

benesch:
@sfackler just wanted to make sure you hadn't missed the latest updates here! No worries if you're busy.

postgres/src/config.rs
```diff
@@ -88,9 +89,9 @@ use tokio_postgres::{Error, Socket};
 /// ```not_rust
 /// postgresql:///mydb?user=user&host=/var/lib/postgresql
 /// ```
-#[derive(Clone)]
 pub struct Config {
     config: tokio_postgres::Config,
+    notification_callback: Box<dyn NotificationCallback>,
```
sfackler: This should just be able to store an `Arc<dyn FnMut(AsyncMessage) + Send>`.

benesch: I tried that. Sadly an `Arc<dyn FnMut(AsyncMessage) + Send>` does not itself implement `Send`, as `Arc` propagates both `Send + Sync` iff its contents are `Send + Sync`. And without the `Send` bound you can't call the callback from a futures context.

I don't see a good way to do this besides this `Box<dyn NotificationCallback>` or the separate `connect_with` method I proposed originally.

My only requirement is to be able to easily use an MPSC transmitter from the notification callback. I don't have particular opinions on how the API should look, so if it'd be easier for you to just hammer on this PR to get into a mergeable shape, please feel free to do so!

sfackler:
Thinking about this more, I wonder if this is actually the right approach for the blocking interface. One problem here is that notifications are only processed while the connection is being actively used (that's not the case in tokio-postgres since the spawned connection will generally always be running).

I think it might actually make some more sense to go back to the API that the old version of the crate had, with an explicit iterator over notifications: https://docs.rs/postgres/0.15.2/postgres/notification/struct.Notifications.html

postgres/src/config.rs
sfackler: Ah right  yeah.

benesch:
> One problem here is that notifications are only processed while the connection is being actively used (that's not the case in tokio-postgres since the spawned connection will generally always be running).

Can you say more about what you mean? By my read of this code

https://github.com/sfackler/rust-postgres/blob/c2a446922c89be16201151d76e7fe9beb203cfc6/postgres/src/config.rs#L359-L365

the connection future is going to be polled automatically and indefinitely by the Tokio runtime, and so the callback will get invoked even if the connection is not seeing active use?

sfackler:
The runtime is single threaded - nothing is happening outside of the calls to block_on.

benesch:
Ah, thanks, that's the piece I was missing. Hmm.

sfackler:
I opened up a PR with the alternate implementation: #588.

PR #568, opened 2020-02-05
Added information on how to use Vec<T> with Client::query_raw

runfalk:
The examples in the rendered documentation looks like this:

```rust
use tokio_postgres::types::ToSql;
use futures::StreamExt;

let params: Vec<String> = vec![
    "first param".into(),
    "second param".into(),
];
let mut it = Box::pin(client.query_raw(
    "SELECT foo FROM bar WHERE biz = $1 AND baz = $2",
    params.iter().map(|p| p as &dyn ToSql),
).await?);

while let Some(row) = it.next().await.transpose()? {
    let foo: i32 = row.get("foo");
    println!("foo: {}", foo);
}
```

```rust
use postgres::types::ToSql;
use fallible_iterator::FallibleIterator;

let params: Vec<String> = vec![
    "first param".into(),
    "second param".into(),
];
let mut it = client.query_raw(
    "SELECT foo FROM bar WHERE biz = $1 AND baz = $2",
    params.iter().map(|p| p as &dyn ToSql),
)?;

while let Some(row) = it.next()? {
    let foo: i32 = row.get("foo");
    println!("foo: {}", foo);
}
```

I tested the code using `cargo test --doc` and rendered it locally using `cargo doc`. I did not know how to run the full suite.

Do you think the `Box::pin` bit is fine or is there a leaner way to do it?

### DIFF
--- a/tokio-postgres/src/client.rs
+++ b/tokio-postgres/src/client.rs
@@ -313,6 +313,34 @@ impl Client {
     /// Panics if the number of parameters provided does not match the number expected.
     ///
     /// [`query`]: #method.query
+    ///
+    /// # Examples
+    ///
+    /// If you have a type like `Vec<T>` where `T: ToSql` Rust will not know how to use it as params. To get around
+    /// this the type must explicitly be converted to `&dyn ToSql`.
+    ///
+    /// ```no_run
+    /// # async fn async_main(client: &tokio_postgres::Client) -> Result<(), tokio_postgres::Error> {
+    /// use tokio_postgres::types::ToSql;
+    /// use futures::{pin_mut, StreamExt};
+    ///
+    /// let params: Vec<String> = vec![
+    ///     "first param".into(),
+    ///     "second param".into(),
+    /// ];
+    /// let mut it = client.query_raw(
+    ///     "SELECT foo FROM bar WHERE biz = $1 AND baz = $2",
+    ///     params.iter().map(|p| p as &dyn ToSql),
+    /// ).await?;
+    ///
+    /// pin_mut!(it);
+    /// while let Some(row) = it.next().await.transpose()? {
+    ///     let foo: i32 = row.get("foo");
+    ///     println!("foo: {}", foo);
+    /// }
+    /// # Ok(())
+    /// # }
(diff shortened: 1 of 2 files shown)


tokio-postgres/src/client.rs
```diff
@@ -313,6 +313,36 @@ impl Client {
...
     ///
     /// [`query`]: #method.query
+    ///
+    /// # Examples
+    ///
+    /// If you have a type like `Vec<T>` where `T: ToSql` Rust will not know how to use it as params. To get around
+    /// this the type must explicitly be converted to `&dyn ToSql`.
+    ///
+    /// ```no_run
+    /// # { #![cfg(feature = "runtime")] // tokio_postgres::connect doesn't exist without "runtime"
```
runfalk: There was some trickery here to make it work with `cargo test --no-default-features --doc`.

sfackler: Instead of that, you can change the signature of `async_main` to pass the `Client` in and avoid having to cfg at all.

tokio-postgres/src/client.rs
```diff
@@ -313,6 +313,36 @@ impl Client {
...
+    /// # async fn async_main() -> Result<(), tokio_postgres::Error> {
+    /// # let (client, _) = tokio_postgres::connect("host=localhost user=postgres", NoTls).await?;
+    /// use tokio_postgres::types::ToSql;
+    /// use futures::StreamExt;
+    ///
+    /// let params: Vec<String> = vec![
+    ///     "first param".into(),
+    ///     "second param".into(),
+    /// ];
+    /// let mut it = Box::pin(client.query_raw(
```
sfackler: I'd probably use pin_mut rather than Box::pin

tokio-postgres/src/client.rs
runfalk: Great suggestion, way less clutter now.

tokio-postgres/src/client.rs
runfalk: Done

sfackler:
Thanks!

PR #576, opened 2020-02-25
Let columns in SimpleQueryRow be visiable

nottxy:
Let columns in `SimpleQueryRow` be visiable


tokio-postgres/src/row.rs
```diff
@@ -253,4 +253,9 @@ impl SimpleQueryRow {
         let buf = self.ranges[idx].clone().map(|r| &self.body.buffer()[r]);
         FromSql::from_sql_nullable(&Type::TEXT, buf).map_err(|e| Error::from_sql(e, idx))
     }
+
+    /// Returns information about the columns returned.
```
sfackler: I'd rather not expose this directly as a `[String]` since there's other information returned that we could provide in the future. Adding a `SimpleColumn` type is a possibility instead.

PR #580, opened 2020-02-28
Add support for time-0.2 types

aloucks:
This is essentially ported from the `chrono` implementation.

### DIFF
--- a/postgres-types/src/lib.rs
+++ b/postgres-types/src/lib.rs
@@ -398,7 +398,7 @@ impl WrongType {
 /// | `chrono::NaiveDate`             | DATE                                |
 /// | `chrono::NaiveTime`             | TIME                                |
 /// | `time::PrimitiveDateTime`       | TIMESTAMP                           |
-/// | `time::OffsetDateTime`          | TIMESTAMP, TIMESTAMP WITH TIME ZONE |
+/// | `time::OffsetDateTime`          | TIMESTAMP WITH TIME ZONE            |
 /// | `time::Date`                    | DATE                                |
 /// | `time::Time`                    | TIME                                |
 /// | `eui48::MacAddress`             | MACADDR                             |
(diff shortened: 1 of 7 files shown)


postgres-types/src/lib.rs
```diff
@@ -391,6 +397,10 @@ impl WrongType {
 /// | `chrono::DateTime<FixedOffset>` | TIMESTAMP WITH TIME ZONE            |
 /// | `chrono::NaiveDate`             | DATE                                |
 /// | `chrono::NaiveTime`             | TIME                                |
+/// | `time::PrimitiveDateTime`       | TIMESTAMP                           |
+/// | `time::OffsetDateTime`          | TIMESTAMP, TIMESTAMP WITH TIME ZONE |
```
sfackler: I think this should only support TIMESTAMP WITH TIME ZONE to align with chrono.

sfackler:
Can you add tests as well?

postgres-types/src/lib.rs
aloucks: Chrono's types embed specific timezones (or rather `UTC` vs `Local`) where as `OffsetDateTime` doesn't have the generic component and is compatible with both `TIMESTAMP` and `TIMESTAMPZ`. I'd prefer to be able to use `OffsetDateTime` with `TIMESTAMP`.

aloucks:
> Can you add tests as well?

Ah! I just found the tests under `tokio-postgres`. I'll update and push a commit shortly.

postgres-types/src/lib.rs
aloucks: > I'd prefer to be able to use OffsetDateTime with TIMESTAMP.

With [cockroach](https://www.cockroachlabs.com/docs/v19.2/timestamp.html#variants) `TIMESTAMP` will always be `UTC`. I'm still looking for clarification on postgres's behavior. I think with the fact that we're explicitly setting the session to `UTC`, it should be OK either way.

aloucks: @sfackler I thought about this some more and came to the same conclusion. `OffsetDateTime` now only supports `TIMESTAMPTZ` in the latest commit. Do you want me to squash these commits?

sfackler:
No need to squash!

sfackler:
Thanks!

PR #584, opened 2020-03-16
Add documentation for crate features

### LINKED ISSUE #544: no feature flag docs?
I can't find any docs on feature flags.  I am familiar with where to look (Cargo.toml) but [others aren't](https://github.com/bikeshedder/deadpool/issues/23)

bikeshedder:
This closes #544

### DIFF
--- a/postgres/src/lib.rs
+++ b/postgres/src/lib.rs
@@ -46,6 +46,20 @@
 //! as an argument. The `NoTls` type in this crate can be used when TLS is not required. Otherwise, the
 //! `postgres-openssl` and `postgres-native-tls` crates provide implementations backed by the `openssl` and `native-tls`
 //! crates, respectively.
+//!
+//! # Features
+//!
+//! The following features can be enabled from `Cargo.toml`:
+//!
+//! | Feature | Description | Extra dependencies | Default |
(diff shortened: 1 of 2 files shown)


postgres/src/lib.rs
```diff
@@ -46,6 +46,20 @@
 //! as an argument. The `NoTls` type in this crate can be used when TLS is not required. Otherwise, the
 //! `postgres-openssl` and `postgres-native-tls` crates provide implementations backed by the `openssl` and `native-tls`
 //! crates, respectively.
+//!
+//! # Features
+//!
+//! The following features can be enabled from `Cargo.toml`:
+//!
+//! | Feature | Description | Extra dependencies | Default |
```
sfackler: Given that we don't have any default features on the postgres crate, let's remove the last column.

sfackler:
Thanks! LGTM other than the one change.

bikeshedder:
I left the last column in because it was simpler to copy and paste it from the `tokio-postgres` crate. (Leaving out the `runtime` feature)

If leaving out that column an extra sentence is needed: `No features are enabled by default.`

sfackler:
Fair enough.

PR #588, opened 2020-03-22
Add a notification API to the blocking client

sfackler:


benesch:
Also just to get the issues linked up, I believe this would close #404.

sfackler:
Yep! For that - the callback-in-Config approach should work just fine I think.

PR #590, opened 2020-04-01
Don't accidentally return early from CopyOutReader

benesch:
The COPY OUT protocol allows sending CopyData packets that have no data.
The (synchronous) CopyOutReader needs to be careful not to return an
empty slice in this case, but instead request more data, since an empty
slice is taken to mean EOF.


sfackler:
IIRC the postgres documentation currently guarantees that each CopyData contains a row of data, but this seems like a good change regardless!

benesch:
Awesome, thanks for the quick review + merge!

PR #600, opened 2020-05-01
Re-add savepoint method to Transaction

nvanbenschoten:
Revives #184.

The rewrite for async/await and Tokio accidentally lost functionality
that allowed users to assign specific names to savepoints when using
nested transactions. This functionality had originally been added
in #184 and had been updated in #374.

This commit revives this functionality using a similar scheme to the
one that existed before. This should allow CockroachDB users to update
to the next patch release of version `0.17`.

### DIFF
--- a/postgres/src/test.rs
+++ b/postgres/src/test.rs
@@ -152,6 +152,57 @@ fn nested_transactions() {
 }
 
 #[test]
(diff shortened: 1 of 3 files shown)


sfackler:
Thanks!

PR #601, opened 2020-05-07
Add support for money type

### LINKED ISSUE #598: postgres-types: Converting Rust type to Postgres Money type
First, thank you for this incredible lib.

Second: I'm new to Rust and spinning up my first backend (DB + Webserver), so please feel free to redirect me other places if I can answer my own questions.

### Problem
While the Postgres `money` type is [specified in postgres-types](https://github.com/sfackler/rust-postgres/blob/master/postgres-types/src/type_gen.rs#L962-L963), it doesn't appear to have `To/FromSql` support. I could implement these myself, but I wonder if they are simple enough to include in the library (if not by default, then perhaps as a feature). 

From what I can tell from the [Postgres docs on Monetary Types](https://www.postgresql.org/docs/9.1/datatype-money.html#DATATYPE-MONEY-TABLE), **it appears as though the `money` type is likely represented as an i64 of cents**. 

#### Rationale:
* Postgres docs list a range of -92233720368547758.08 to +92233720368547758.07. If you take the high number, you'll notice that it is (2^63)-1 / 100. So it seems like a good bet that Postgres is modeling the cents under the covers, but "rendering" in dollars.

#### Proposed Code Change
* Add `MONEY` to the `simple_from!` and `simple_to!` entries for `i64` in src/lib.rs:
```rust
 (Line 563) simple_from!(i64, int8_from_sql, INT8, MONEY);
 (Line 889) simple_to!(i64, int8_to_sql, INT8, MONEY);
```

I made this change to my local cargo lib, and was able to run a query against a `money` Postgres column and serialize it into an i64 struct field successfully. I have not tested the reverse path, going from Rust i64 to PG money column.

Snippets of the struct, impl, and Postgres table below:

```rust
pub struct Account {
    id: Uuid,
    name: String,
    balance: i64,
}

impl From<Row> for Account {
    fn from(row: Row) -> Self {
        Self {
            id: row.get("id"),
            name: row.get("name"),
            balance: row.get("balance"),
        }
    }
}
```
<localhost>=# \d accounts;
                           Table "public.accounts"
 Column  |         Type          | Collation | Nullable |      Default
---------+-----------------------+-----------+----------+--------------------
 id      | uuid                  |           | not null | uuid_generate_v4()
 name    | character varying(40) |           | not null |
 balance | money                 |           | not null | 0
```

#### Open Questions:
* Is the macro the correct approach, or should `money` get it's own explicit impl?
* Is this a common enough conversion to include by default, or should it be a feature?
* From the Postgres docs: "The fractional precision is determined by the database's lc_monetary setting." This seems like it could cause some wackiness.
* Further investigation would have to be done to ensure the conversion from Rust -> PG works as expected.
* Is i64 a good way to model money?

tkbrigham:
Closes #598 

Implements `ToSql` and `FromSql` for the [`postgres_money::Money`](https://docs.rs/postgres_money/0.2.0/postgres_money/struct.Money.html) struct, behind the feature `with-money-0_2`.

### DIFF
--- a/postgres-protocol/src/types/mod.rs
+++ b/postgres-protocol/src/types/mod.rs
@@ -441,6 +441,22 @@ pub fn uuid_from_sql(buf: &[u8]) -> Result<[u8; 16], StdBox<dyn Error + Sync + S
     Ok(out)
 }
 
+/// Serializes a `Money` value.
+#[inline]
+pub fn money_to_sql(v: i64, buf: &mut BytesMut) {
+    buf.put_i64(v);
+}
+
+/// Deserializes a `Money` value.
+#[inline]
+pub fn money_from_sql(mut buf: &[u8]) -> Result<i64, StdBox<dyn Error + Sync + Send>> {
+    let v = buf.read_i64::<BigEndian>()?;
+    if !buf.is_empty() {
+        return Err("invalid buffer size".into());
+    }
+    Ok(v)
+}
+
 /// Serializes an array value.
 #[inline]
 pub fn array_to_sql<T, I, J, F>(
--- a/postgres-types/Cargo.toml
+++ b/postgres-types/Cargo.toml
@@ -16,6 +16,7 @@ with-bit-vec-0_6 = ["bit-vec-06"]
 with-chrono-0_4 = ["chrono-04"]
 with-eui48-0_4 = ["eui48-04"]
 with-geo-types-0_4 = ["geo-types-04"]
+with-money-0_1 = ["money-01"]
 with-serde_json-1 = ["serde-1", "serde_json-1"]
 with-uuid-0_8 = ["uuid-08"]
 with-time-0_2 = ["time-02"]
@@ -30,6 +31,7 @@ bit-vec-06 = { version = "0.6", package = "bit-vec", optional = true }
 chrono-04 = { version = "0.4", package = "chrono", optional = true }
 eui48-04 = { version = "0.4", package = "eui48", optional = true }
 geo-types-04 = { version = "0.4", package = "geo-types", optional = true }
+money-01 = { version = "0.2", package = "postgres_money", optional = true }
 serde-1 = { version = "1.0", package = "serde", optional = true }
 serde_json-1 = { version = "1.0", package = "serde_json", optional = true }
 uuid-08 = { version = "0.8", package = "uuid", optional = true }
--- a/postgres-types/Cargo.toml
+++ b/postgres-types/Cargo.toml
@@ -16,7 +16,7 @@ with-bit-vec-0_6 = ["bit-vec-06"]
 with-chrono-0_4 = ["chrono-04"]
 with-eui48-0_4 = ["eui48-04"]
 with-geo-types-0_4 = ["geo-types-04"]
-with-money-0_1 = ["money-01"]
+with-money-0_2 = ["money-02"]
 with-serde_json-1 = ["serde-1", "serde_json-1"]
 with-uuid-0_8 = ["uuid-08"]
 with-time-0_2 = ["time-02"]
@@ -31,7 +31,7 @@ bit-vec-06 = { version = "0.6", package = "bit-vec", optional = true }
 chrono-04 = { version = "0.4", package = "chrono", optional = true }
 eui48-04 = { version = "0.4", package = "eui48", optional = true }
 geo-types-04 = { version = "0.4", package = "geo-types", optional = true }
-money-01 = { version = "0.2", package = "postgres_money", optional = true }
+money-02 = { version = "0.2", package = "postgres_money", optional = true }
 serde-1 = { version = "1.0", package = "serde", optional = true }
 serde_json-1 = { version = "1.0", package = "serde_json", optional = true }
 uuid-08 = { version = "0.8", package = "uuid", optional = true }
--- a/postgres-types/src/lib.rs
+++ b/postgres-types/src/lib.rs
@@ -196,8 +196,8 @@ mod chrono_04;
 mod eui48_04;
 #[cfg(feature = "with-geo-types-0_4")]
 mod geo_types_04;
-#[cfg(feature = "with-money-0_1")]
-mod money_01;
+#[cfg(feature = "with-money-0_2")]
+mod money_02;
 #[cfg(feature = "with-serde_json-1")]
 mod serde_json_1;
 #[cfg(feature = "with-time-0_2")]
--- a/postgres-types/src/lib.rs
+++ b/postgres-types/src/lib.rs
@@ -196,6 +196,8 @@ mod chrono_04;
 mod eui48_04;
 #[cfg(feature = "with-geo-types-0_4")]
 mod geo_types_04;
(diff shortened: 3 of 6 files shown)


sfackler:
For a postgres-specific crate like postgres_money, these impls would just live directly in that crate and not require a feature over here.

tkbrigham:
@sfackler the `postgres_money` crate is not dependent on the rust-postgres `ToSql` and `FromSql` traits in order to be useful. It is just representation of data that should mirror what Postgres does for the `money` type. In fact, the `postgres_money` crate could potentially be valuable without every communicating with Postgres if you wanted to use it to see how Postgres would parse and represent strings (though this is admittedly farfetched).

The pattern I believe you are suggesting is keep all of the impls in the postgres_money crate, but following this idea leads to the following situation:
* Another crate (`new_postgres_reader`) wants to use `postgres_money`
* They open a pull request to `postgres_money` that implements the traits that `new_postgres_reader` needs to communicate
* `postgres_money` now has an additional dependency on another crate. admittedly easy to feature flag, but still.

This seems to fly in the face of encapsulation, where this crate now has to have knowledge of its consumers, instead of the consumers taking a dependency on this crate. It seems that unless every crate that interfaces with Postgres data directly has the same APIs and signatures, implementing the SQL representation traits for `postgres_money::Money` would make sense to live closer to the Postgres binding crate.

As a concrete example, if we pretend Diesel didn't have the [Money](https://docs.diesel.rs/diesel/pg/types/sql_types/struct.Money.html) type support in place, they could take a dependency on `postgres_money` and implement their bindings between `postgres_money::Money` and their DB interface. Conversely, if we added the binding in the `postgres_money` crate, `postgres_money` would take a dependency on Diesel and vice versa.

So it seems like the SQL-interface layer is better served as a feature within the crates that access Postgres, not within the generic representation of that data.

arekbal:
> They open a pull request to postgres_money that implements the traits that new_postgres_reader needs to send data to PG
Don't blame me if I am wrong(a noob here) but isn't postgres_money meant to be more like sattelite crate - an extension for main postgres crates instead? Not like let-everybody-in club.
> This seems to fly in the face of encapsulation.

sfackler:
> The pattern I believe you are suggesting is keep all of the impls in the postgres_money crate, but that seems to lead to the following situation:

This does not appear to have been a problem in the last 5 or so years of the development of this crate.

sfackler:
You can always have a feature flat on postgres_money to make the dependency on postgres_types optional.

tkbrigham:
If I were someone new coming to the rust-postgres crate and I wanted support for both the `uuid` and `money` column types, I would find it odd that one was natively supported (behind a feature) and I'd have to find my own implementation for the other. 

If the goal is to have developers choose their own implementation because `rust-postgres` doesn't support it natively, then implementing the ToSql and FromSql elsewhere seems reasonable, just not what I had in mind.

sfackler:
I don't understand - you have to use an external implementation of the `Uuid` type just as you would for the `Money` type. The features just add the impl for crates where it doesn't make sense for them to support it natively.

tkbrigham:
I think the drivers for my thought process are basically three things, broken down into user concerns (devs using the rust-postgres crate) and maintainer concerns (e.g. you for rust-postgres and me for postgres_money):

### User Concerns
#### Discoverability
As a developer, I decide to us PG as a backend. I discover this crate, and pull it in as a dep. I know I have some tables that will depend on `money`, so I look through `rust-postgres` to see how I can interface with this type. 

##### Possible outcomes:
* If `money` is a supported feature in `rust-postgres`, I simply enable it, rebuild and am on my way.
* If `money` is not supported, possibly `rust-postgres` suggests/documents the `postgres_money` crate, the dev can go out and import this crate themselves with the appropriate features. (From my perspective, why would `rust-postgres` suggest a crate but not implement it behind a feature?)
* If `money` is not supported and not suggested/documented, the developer is left with the task of finding their own implementation/solution

#### Robust Postgres Support
This crate says `PostgreSQL support for Rust`. There is a gap though - the Postgres `money` type. If this crate is supporting Postgres, why should support for this native Postgres feature not live inside this crate? If the `money` support is behind a feature flag, it's hard for me to understand what the drawbacks of collocating money support with a postgres crate are.

You mentioned this hasn't been a concern with other crates - can you give me examples of other crates that implement Postgres native types and also implement rust-postgres ToSql and FromSql impls?

### Maintainer Concerns
#### Maintainability
If your concern is to avoid adding more dependencies to the rust-postgres footprint, or concerns with being the one responsible to maintain the ToSql and FromSql implementations for the `postgres_money` crate, that seems like a fine reason not to want to include this dependency. If that is the case, I'm happy to commit to handling bugfixes and feature requests for this bit of functionality.

sfackler:
> * If `money` is a supported feature in `rust-postgres`, I simply enable it, rebuild and am on my way.

No, you enable the feature on the postgres dependency and add the dependency on the postgres_money crate. How is that significantly different from adding a dependency on the postgres_money crate and (maybe) enabling a feature on the postgres_money crate?

tkbrigham:
> How is that significantly different from adding a dependency on the postgres_money crate and (maybe) enabling a feature on the postgres_money crate?

Because of discoverability. The developer has already found the `rust-postgres` crate, so enabling a feature within that crate is a matter of reading the docs. 

On the other hand, if support for `money` is something that `rust-postgres` doesn't know about, then the developer must go do research and determine 1) whether or not there are `money` implementations for `rust-postgres` and 2) spend time evaluating which implementation they want to use.

This is all in addition to the fact that `rust-postgres` technically has a feature gap, which is support for Postgres' native `money` type. It seems like if `rust-postgres` is intended as a fully-featured driver for Postgres, support for the `money` type would be welcome.

I'm having a hard time understanding your motivations and responses from the feedback you are providing. Perhaps you could share what your goals are for suggesting this live outside of rust-postgres, or what you are optimizing for generally? If you do, I could help address those concerns or understand why my suggested approach is not preferable.

sfackler:
> The developer has already found the `rust-postgres` crate, so enabling a feature within that crate is a matter of reading the docs.

The docs can point people towards postgres-money.

> I'm having a hard time understanding your motivations and responses from the feedback you are providing.

I want to minimize the number of feature flags exposed by the postgres crate. If a library is built specifically for postgres, like postgres-array, or postgres-range, or postgres-money, then that's great since a feature flag won't be necessary for those types.

I want to avoid having types with nontrivial APIs (like ranges or arrays or money) in this crate. That allows those APIs to have more freedom to iterate through e.g. breaking changes. I do not want to cut breaking releases of the core postgres crates very often since each break pushes work downstream to update.

tkbrigham:
@sfackler understood. The minimization of feature flags is definitely not what I'd expected, so that is good to know.

I will close this PR. I'd probably lean towards keeping the issue open, but I'll defer to you on that.

nixpulvis:
Both `uuid` (for example) and `money` are built in types to Postgres, as a user I would fully expect to enable support for them in the same way with this crate (assuming they aren't enabled by default). If there's an issue of too many feature flags, perhaps that could be helped by making groups of feature flags? What is the problem with having a lot of feature flags?

Issue #585, opened 2020-03-16
Allow deriving Type for custom Enums


oberien:
I'm using COPY IN with FORMAT binary with custom enums. The `BinaryCopyInWriter::new` function requires a list of Types for serialization. However, getting the type of a custom enum on which `ToSql` is derived seems to be non-trivial.

Assume the following enum:
```rust
#[derive(Debug, ToSql)]
pub enum Foo {
    VariantA,
    VariantB,
    VariantC,
}
```
Currently, I need to manually write out the type (and keep it synchronized by hand) like this:
```rust
Type::new(
    "Foo".to_string(),
    0, // Oid, not relevant
    Kind::Enum(vec![
        "VariantA".to_string(),
        "VariantB".to_string(),
        "VariantC".to_string(),
    ]),
    "".to_string(), // schema, not relevant
)
```

I see 3 possible solutions:
1. Allow `Type::TEXT` as Type for enums, because that's what they are serialized as anyways.
2. Create a new derivable trait `GetType` (or some other name):
    ```rust
    impl GetType for Foo {
        fn type() -> Type {
            Type::new(...)
        }
    }
    ```
    * Another idea was to create an associated constant. However `Type::new` would need to be a const-fn and `vec![]` would need to be const as well, which isn't the case. So that option is not possible (yet).
3. Add another associated constant to `Type`, which can be used as a placeholder for custom enums, similar to what `anyenum` is for binary functions. I don't know enough about the postgres ABI, but maybe just using `anyenum` here would already be possible?

Solutions (1) and (3) are only focusing on enums. However, the same should probably also be possible for custom struct types.

sfackler:
A better solution IMO would be to just prepare a one-off query to get the real Type from the server:

```rust
client.prepare("SELECT $1::myenum")?.params()[0].clone()
```

sfackler:
The OID of custom types is set by the database backend when they are defined. There is no way to determine it at compile time.

seanlinsley:
Is the OID actually needed? It seems to work from my and the OP's testing.

sfackler:
Oh sorry I misremembered. It just relies on the `Type`s to handle the data conversions.

If you don't care about that, you can just pass in an arbitrary type and have your `ToSql` implementation ignore that argument.

Alternatively you can always write your own binary copy stream.

sfackler:
[...]

PR #604, opened 2020-05-21
Fix enum name check.

moose-consulting:
I think there is an issue with checking the name derived using `ToSql` on Enums and the case-insensitivity of the Postgres type names. I ran into issues when I derived `ToSql` and then tried to use `prepared_typed` to insert the Enum values. The types I used in `prepare_typed` were fetched from the server using the suggestion from #585. The original type definition used to create the type in Postgres had a casing that matched the Enum name in rust, but the type name returned from the query was lowercase. I know this can probably be fixed with `#[postgres(name="foo")]`, but I think it is better to handle this by default. This patch may not cover all possible scenarios that we want to match.

### DIFF
--- a/postgres-derive/src/tosql.rs
+++ b/postgres-derive/src/tosql.rs
@@ -10,8 +10,7 @@ use crate::overrides::Overrides;
 
 pub fn expand_derive_tosql(input: DeriveInput) -> Result<TokenStream, Error> {
     let overrides = Overrides::extract(&input.attrs)?;
-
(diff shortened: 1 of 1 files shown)


sfackler:
That is what the `#[postgres(name = "...")]` attribute is for: https://docs.rs/postgres-types/0.1.1/postgres_types/#naming

moose-consulting:
It seems strange to require the extra macro by default.

PR #605, opened 2020-05-25
Add support for geo-types=0.5

jakubadamw:
Support for geo-types=0.4 (via the `with-geo-types_04` feature) has been preserved for convenience.


postgres-types/src/geo_types_04_or_05.rs
```diff
@@ -1,6 +1,11 @@
+#[cfg(all(feature = "with-geo-types-0_4", not(feature = "with-geo-types-0_5")))]
```
sfackler, line 1: The purpose of naming the features `geo-types-04` is so that multiple versions can coexist at the same time.

jakubadamw: @sfackler, okay! I have extracted out a new module. 🙂

sfackler:
Thanks!

PR #606, opened 2020-05-26
Fix clippy lint warnings regarding indirect structural matches

jpgneves:
With `#[warn(future_incompatible)]`, one gets a warning in the `accepts!` macro due an indirect structural match (see https://github.com/rust-lang/rust/issues/62411).

E.g. today one gets:

```
warning: to use a constant of type `std::sync::Arc` in a pattern, `std::sync::Arc` must be annotated with `#[derive(PartialEq, Eq)]`
   --> postgres-types/src/lib.rs:148:19
    |
148 |                 $($crate::Type::$expected)|+ => true,
    |                   ^^^^^^^^^^^^^^^^^^^^^^^
...
944 |     accepts!(INET);
    |     --------------- in this macro invocation
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #62411 <https://github.com/rust-lang/rust/issues/62411>
    = note: this warning originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

This PR implements option number 2 in the fixes suggested in the above tracking issue, and enables the `future_incompatible` lint to ensure we conform.

### DIFF
--- a/postgres-types/src/lib.rs
+++ b/postgres-types/src/lib.rs
@@ -106,7 +106,7 @@
 //! }
 //! ```
 #![doc(html_root_url = "https://docs.rs/postgres-types/0.1")]
-#![warn(clippy::all, rust_2018_idioms, missing_docs)]
+#![warn(clippy::all, future_incompatible, rust_2018_idioms, missing_docs)]
 
 use fallible_iterator::FallibleIterator;
 use postgres_protocol::types::{self, ArrayDimension};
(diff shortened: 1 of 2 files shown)


sfackler:
I believe this lint is allow by default because the handling is not sufficiently precise in cases just like this. This is something that should be fixed in the compiler IMO.

PR #630, opened 2020-07-08
Move SqlState and DbError to postgres-protocol

### LINKED ISSUE #594: Could SqlState live in postgres-protocol?
The auto-generated `SqlState` is super useful for us (https://github.com/MaterializeInc/materialize/pull/2595), but it's a bit unfortunate to have to import all of `tokio_postgres` or `postgres` to pull it in. Do you think we could move it to `postgres_protocol`, and re-export in `tokio_postgres`?

benesch:
They seem to belong there a bit better. The one possibly weird change is
that DbError::parse is now public. Happy to try to address that if it's
a problem.

Fix #594.


sfackler:
My general intuition for postgres-protocol is that it contains the very raw low-level protocol parsing logic and nothing else. Things like `Error` and  `Display` impls doesn't seem to match that.

benesch:
Ok, gotcha. What about moving just `SqlState` then? I think that would align well with your stated categorization.

sfackler:
Sure, that seems reasonable.

PR #634, opened 2020-07-17
Add Debug + Clone to structs in postgres-native-tls

RAnders00:
Straight-forward, I found these to be missing when combining the TLS implementation with a `bb8` connection pool, which requires the `TlsConnector` to be `Clone`.


RAnders00:
Sorry, I had some types mixed up and jumped to conclusions too quickly given the compile errors I got. I inserted `TlsConnector` in a place that expected `MakeTlsConnector`, which caused me to intially thing the `Clone` was simply missing.

PR #636, opened 2020-07-22
Configuration to disable internal stmt cache

pimeys:
This enables usage with pgBouncer's transaction mode. Typically when using the transaction mode, a client gets a new connection from pgBouncer for every new transaction. It's quite useful and allows one to use prepared statements in this mode with the following workflow:

```sql
-- start a new transaction
BEGIN
-- deallocate all stored statements from the server to prevent
-- collisions
DEALLOCATE ALL
-- run the queries here
-- ..
-- ..
COMMIT -- or ROLLBACK
```

Now in a case where the query uses custom types such as enums, what tokio-postgres does is it fetches the type info for the given type, stores the info to the cache and also caches the statements for fetching the info to the client. Now when we have two tables with different custom types in both of them, we can imagine the following workflow:

```rust
// first query
client.simple_query("BEGIN")?;
client.simple_query("DEALLOCATE ALL")?;
let stmt = client.prepare("SELECT \"public\".\"User\".\"id\", \"public\".\"User\".\"userType\" FROM \"public\".\"User\" WHERE 1=1 OFFSET $1")?;
dbg!(client.query(&stmt, &[&0i64])?);
client.simple_query("COMMIT")?;

// second query
client.simple_query("BEGIN")?;
client.simple_query("DEALLOCATE ALL")?;
let stmt = client.prepare("SELECT \"public\".\"Work\".\"id\", \"public\".\"Work\".\"workType\" FROM \"public\".\"Work\" WHERE 1=1 OFFSET $1")?;
dbg!(client.query(&stmt, &[&0i64])?);
client.simple_query("COMMIT")?;
```

The `userType` and `workType` are both enums, and the preparing of the second query will give an error `prepared statement "s1" does not exist`, where `s1` is the query to the `pg_catalog` for the type info for the second time, using the cached statement.

The change here gives an extra flag for the client to disable caching of statements.


postgres/src/config.rs
```diff
@@ -307,6 +307,21 @@ impl Config {
         self.config.get_channel_binding()
     }
 
+    /// When enabled, the client skips all internal cached for statements,
```
tomhoule, line 310: nitpick: I think you meant "skips all internal caching".

tokio-postgres/src/config.rs
```diff
@@ -387,6 +389,21 @@ impl Config {
         self.channel_binding
     }
 
+    /// When enabled, the client skips all internal cached for statements,
```
tomhoule, line 392: nitpick: same typo as above

sfackler:
Rather than having a separate mode that has to be enabled, it seems like it could be cleaner to just have some logic to detect if the typeinfo query fails because the statement isn't live and re-prepare it at that point.

pimeys:
This would trigger an extra roundtrip though? Could we detect the pgBouncer mode when connecting? Or I was also thinking of just querying all the info at the beginning and using if from the `HashMap`...

pimeys:
Could we maybe solve this at some point? I'd like to find a way to either:
a) Completely disable the caches
b) Query all type information in the beginning, and use it through the lifetime of the connection.

The first adds roundtrips to every query using enums and other custom stuff, the latter will then break whenever a new type is introduced and the connection is running with old data. We have users using pgbouncer that cannot cache statements, so whatever the master branch is doing right now will give quite interesting errors to them.

pimeys:
Also, right now when using our custom branch of tokio-postgres, how rust-decimal has tokio-postgres as a dependency forces us to use a custom branch of rust-decimal, and therefore a custom branch of every other crate that uses rust-decimal. You understand it is quite a nightmare to maintain right now! :)

benthecarman:
Any updates here?

PR #645, opened 2020-08-01
Give a more helpful message on error

jyn514:
Before:

```
database error: ERROR: insert or update on table "owner_rels" violates foreign key constraint "owner_rels_cid_fkey"
```

After:

```
database error: ERROR: insert or update on table "owner_rels" violates foreign key constraint "owner_rels_cid_fkey"
DETAIL:  Key (cid)=(4) is not present in table "releases".
```

Warning: this is completely untested except for `cargo check`, I didn't see existing tests for `Display` and docs.rs is still stuck on 0.15.

### DIFF
--- a/tokio-postgres/src/error/mod.rs
+++ b/tokio-postgres/src/error/mod.rs
@@ -311,10 +311,10 @@ impl fmt::Display for DbError {
     fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(fmt, "{}: {}", self.severity, self.message)?;
         if let Some(detail) = &self.detail {
-            write!(fmt, "DETAIL: {}", detail)?;
+            write!(fmt, "\nDETAIL: {}", detail)?;
         }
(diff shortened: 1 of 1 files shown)


tokio-postgres/src/error/mod.rs
```diff
@@ -309,7 +309,14 @@ impl DbError {
 
 impl fmt::Display for DbError {
     fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(fmt, "{}: {}", self.severity, self.message)
+        write!(fmt, "{}: {}", self.severity, self.message)?;
+        if let Some(detail) = &self.detail {
+            write!(fmt, "DETAIL: {}", detail)?;
```
sfackler, line 314: Shouldn't there be some whitespace before DETAIL and HINT?

jyn514: Good catch, I added a newline before each.

sfackler:
Thanks!

PR #655, opened 2020-09-16
Permit configuring the notice callback

benesch:
Right now the behavior is hardcoded to log any received notices at the
info level. Add a `notice_callback` configuration option that permits
installing an arbitrary callback to handle any received notices.

As discussed in #588.

### DIFF
--- a/postgres/src/config.rs
+++ b/postgres/src/config.rs
@@ -4,13 +4,16 @@
 
 use crate::connection::Connection;
 use crate::Client;
+use log::info;
 use std::fmt;
 use std::path::Path;
 use std::str::FromStr;
+use std::sync::Arc;
 use std::time::Duration;
 use tokio::runtime;
 #[doc(inline)]
 pub use tokio_postgres::config::{ChannelBinding, Host, SslMode, TargetSessionAttrs};
+use tokio_postgres::error::DbError;
 use tokio_postgres::tls::{MakeTlsConnect, TlsConnect};
 use tokio_postgres::{Error, Socket};
 
@@ -90,6 +93,7 @@ use tokio_postgres::{Error, Socket};
 #[derive(Clone)]
 pub struct Config {
     config: tokio_postgres::Config,
+    notice_callback: Arc<dyn Fn(DbError) + Send + Sync>,
 }
 
 impl fmt::Debug for Config {
@@ -109,9 +113,7 @@ impl Default for Config {
 impl Config {
     /// Creates a new configuration.
     pub fn new() -> Config {
-        Config {
-            config: tokio_postgres::Config::new(),
-        }
+        tokio_postgres::Config::new().into()
     }
 
     /// Sets the user to authenticate with.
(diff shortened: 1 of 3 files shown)


postgres/src/config.rs
```diff
@@ -307,6 +316,30 @@ impl Config {
...
+    /// the same structure as errors, but they are not "errors" per-se.
+    ///
+    /// Notices are distinct from notifications, which are instead accessible
+    /// via the [`Notifications`] API.
+    ///
+    /// [`AsyncMessage::Notice`]: tokio_postgres::AsyncMessage::Notice
+    /// [`Notifications`]: crate::Notifications
+    pub fn notice_callback<F>(&mut self, f: F) -> &mut Config
+    where
+        F: Fn(DbError) + Clone + Send + 'static,
```
sfackler, line 332: I would just store this in an Arc internally rather than forcing the closure to be Clone.

postgres/src/config.rs
```diff
@@ -307,6 +316,30 @@ impl Config {
...
+    {
+        self.notice_callback = Box::new(f);
+        self
+    }
+
+    /// Gets the current notice callback.
+    pub fn get_notice_callback(&self) -> &(dyn FnMut(DbError) + Send) {
```
sfackler, line 339: This doesn't really seem to be that useful, right?

postgres/src/config.rs
sfackler: That would also avoid the need for the NoticeCallback trait.

postgres/src/config.rs
benesch: Nope. Added just for consistency with the other accessors I added in #535. I don't particularly care about this, though, so happy to remove it if you prefer.

postgres/src/config.rs
benesch: Ack, done.

benesch:
Thanks for the initial review, @sfackler! This should be ready for another look.

postgres/src/config.rs
sfackler: Yeah, let's remove it for now then.

postgres/src/config.rs
```diff
@@ -90,6 +93,7 @@ use tokio_postgres::{Error, Socket};
 #[derive(Clone)]
 pub struct Config {
     config: tokio_postgres::Config,
+    notice_callback: Arc<dyn Fn(DbError) + Send>,
```
sfackler, line 96: This does mean that `Config` will no longer be `Sync` - do you think it'll be a problem to require the closure to be `Sync + Send` instead?

sfackler: Oh right, mpsc channels aren't sync :/ I guess we'll just do this and see if it breaks anyone then :(

benesch:
Since bounded channels are Sync—I only learned that recently—I’m actually
fine to require the Sync bound. Want me to do that? Don’t want to cause you
BC trouble over this.

sfackler:
It's definitely safer to require a Sync bound, at least for now.

postgres/src/config.rs
benesch: Done.

benesch:
> It's definitely safer to require a Sync bound, at least for now.

You bet. Done.

sfackler:
Thanks!

PR #659, opened 2020-09-25
Disable chrono's "oldtime" feature to drop time v0.1

benesch:
The latest version of chrono, v0.4.16, permits removing its dependency
on the long-deprecated time v0.1 crate by disabling the "oldtime"
feature.

For backwards compatibility, chrono is leaving the "oldtime" feature on
by default, so disabling the "oldtime" feature requires disabling all
default features and then re-enabling the one default feature ("clock")
that rust-postgres needs.

Note that this change does not cause any backwards-compatibility
problems for users of rust-postgres. The "oldtime" feature controls only
whether `time_v01::Duration` and `chrono::Duration` are the same type
and does not affect any of the APIs used by rust-postgres.


sfackler:
Thanks!

PR #664, opened 2020-10-15
Immutable borrow of `Client` in transactions

P-E-Meunier:
This (fairly large) PR changes the implementation of transactions in order to avoid taking mutable borrows.

The idea is to introduce a new kind of message from the frontend, namely `connection::RequestMessages::Transaction(…)` containing just a channel receiver. When receiving this message, the backend stores that receiver, and then takes its messages from that receiver only, until the sender is dropped.

This guarantees that only messages from that transaction are sent between `BEGIN` and `COMMIT`/`ROLLBACK`, which was achieved before by having the `Transaction` type take a mutable borrow on the `Client`. However, that previous solution is not convenient when multiple Tokio tasks share a reference to the client (which I think is a pretty common case in a web server), since mutexes could result in a deadlock in that case (as the task that can unlock the mutex is on the same thread as the waiting task), and refcells in a panic.

### DIFF
--- a/tokio-postgres/src/client.rs
+++ b/tokio-postgres/src/client.rs
@@ -53,12 +53,24 @@ impl Responses {
     }
 }
 
-struct State {
+pub(crate) struct State {
     typeinfo: Option<Statement>,
     typeinfo_composite: Option<Statement>,
     typeinfo_enum: Option<Statement>,
     types: HashMap<Oid, Type>,
     buf: BytesMut,
+}
+
+impl State {
+    pub(crate) fn new() -> Self {
+         State {
+             typeinfo: None,
+             typeinfo_composite: None,
+             typeinfo_enum: None,
+             types: HashMap::new(),
+             buf: BytesMut::new(),
+         }
+    }
 }
 
 pub struct InnerClient {
@@ -67,6 +79,14 @@ pub struct InnerClient {
 }
 
 impl InnerClient {
+
+    pub fn new(sender: mpsc::UnboundedSender<Request>) -> Self {
+        InnerClient {
+            sender,
+            state: Mutex::new(State::new()),
+        }
+    }
+
     pub fn send(&self, messages: RequestMessages) -> Result<Responses, Error> {
         let (sender, receiver) = mpsc::channel(1);
         let request = Request { messages, sender };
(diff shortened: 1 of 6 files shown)


P-E-Meunier:
By the way, I couldn't run the tests as I don't know how to run them. You might want to add a README in the tests directory as to how to do that, and how to avoid problems with the other databases I might have on my system.

sfackler:
> However, that previous solution is not convenient when multiple Tokio tasks share a reference to the client

This is a pattern I would discourage for exactly this reason. Connections are stateful objects that should not be used concurrently by multiple tasks. A connection pool is a more appropriate approach.

sfackler:
> since mutexes could result in a deadlock in that case (as the task that can unlock the mutex is on the same thread as the waiting task)

An async mutex would cover this issue, but that's basically just a connection pool with size 1.

P-E-Meunier:
> A connection pool is a more appropriate approach.

If I never used transactions, what would the problem be?

After looking at the code, it seems to me that the sequence of commands sent to the Postgres socket in the pooled and non-pooled cases is the same, is that not the case?

Moreover, connection pools and async mutexes seem to add complexity and dependencies. Also, this PR doesn't break connection pools.

sfackler:
If you never used transactions, you don't need this PR right?

Both the futures and tokio crates provide async mutexes. No additional dependencies are required.

P-E-Meunier:
> If you never used transactions, you don't need this PR right?

That's right, but my question was more general: when you say connections are stateful objects, is there something I'm misunderstanding, or are transactions the only stateful thing in connections?

> Both the futures and tokio crates provide async mutexes. No additional dependencies are required.

Right, but connection pools are not.

sfackler:
There are various other bits of connection specific state like isolation levels, time zones, string encodings, etc.

> Right, but connection pools are not.

Okay, but if you're already okay with using a single connection for all of your tasks, then you don't need a pool at all.

P-E-Meunier:
> Okay, but if you're already okay with using a single connection for all of your tasks, then you don't need a pool at all.

Does that mean that having more than one connection per thread can have any performance benefit?

sfackler:
If you want to be able to have multiple tasks running queries at the same time, yes.

PR #679, opened 2020-10-20
Get for text results when querying

### LINKED ISSUE #389: Support the `ltree` data type
The result should be just `String`.

https://www.postgresql.org/docs/9.1/ltree.html

EPashkin:
Fix #389

### DIFF
--- a/tokio-postgres/src/client.rs
+++ b/tokio-postgres/src/client.rs
@@ -226,6 +226,25 @@ impl Client {
             .await
     }
 
+    /// Same as `query` but returns text results instead of binary
+    ///
+    /// We introduce an additional method rather than parameters not to break all
+    /// the conversions and marshalling supplied by `postgres-types` and feature crates and
+    /// inasmuch as this allows us to maintain API compatibility with upstream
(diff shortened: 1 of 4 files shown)


tokio-postgres/src/client.rs
```diff
@@ -226,6 +226,25 @@ impl Client {
...
+    /// Same as `query` but returns text results instead of binary
+    ///
+    /// We introduce an additional method rather than parameters not to break all
+    /// the conversions and marshalling supplied by `postgres-types` and feature crates and
+    /// inasmuch as this allows us to maintain API compatibility with upstream
+    pub async fn query_with_text_results<T>(
+        &self,
+        statement: &T,
+        params: &[&(dyn ToSql + Sync)],
+    ) -> Result<Vec<Row>, Error>
```
sfackler, line 238: This shouldn't use the `Row` type since `FromSql` will not work with almost any of the types returned in text mode.

EPashkin:
@sfackler Sorry, as you stated, no good solution for this issue.
I close MR.

PR #684, opened 2020-10-27
better api for passing parameters to `*_raw` methods

### LINKED ISSUE #683: limitation in how parameters are passed to `*_raw` methods and `ToSql`
I was trying to implement inserting a lot of rows in batches and I think I found a limitation in the way parameters are passed to the `*_raw` methods. Here's some code to explain what I was trying to do (the code is incomplete, I've added just enough to ilustrate the issue):

```rust
struct Data {
    id: u16,
    data: serde_json::Value,
}

async fn insert_data(cnn:&tokio_postgres::Client, xs:&[Data]) -> Result<()> {
    let sql = ... // sql to insert xs.len() rows
    let params = ... // create some iterator from xs
    cnn.query_raw(sql, prams).await?;
    Ok(())
}
```

Next I tried to implement an interator that returns all the parameters for all rows from `xs`. So I created the following:

```rust
struct DataParams<'a> {
    xs:&'a [Data]
}

impl<'a> Iterator for DataParams<'a> {
    type Item = &'a dyn tokio_postgres::types::ToSql;
    ...
}
```
But this doesn't work because `u16` and `serde_json::Value` don't implement `ToSql` and I need to convert or wrap them in a type that does implement it, so i cannot return an element with lifetime `'a`. Then I had the following idea:

```rust
enum DataParam<'a> {    
    Id(i32),
    Data(tokio_postgres::types::Json(&'a serde_json::Value)),
}
impl<'a> Iterator for DataParams<'a> {
    type Item = DataParam<'a>;
    ...
}
impl<'a> tokio_postgres::types::ToSql for DataParam<'a> {
    ...
}
```

but also ends up not working because `query_raw` needs an iterator that returns references to `dyn tokio_postgres::types::ToSql`, and I don't think that can be achieved with this strategy.

As a workaround I ended up creating a `Vec<DataParam<'a>>` and creating the iterator from it, but i started looking at the rust-postgres code to see if `query_raw` could be changed to receive an iterator to some type that implements `ToSql`. After some investigation I think that can be done but it requires adding:

```rust
impl ToSql for &dyn ToSql {
    fn to_sql(&self, ty: &Type, w: &mut BytesMut) -> Result<IsNull, Box<dyn Error + Sync + Send>> {
        (*self).to_sql_checked(ty, w)
    }

    fn accepts(_ty: &Type) -> bool {
        true
    }

    to_sql_checked!();
}
```

that feels a bit wrong.

Is this something that's worth investigating? or maybe I missed something, or the workaround of using a `Vec` is good enough?

vemoo:
This enables passing iterators of `&dyn ToSql` or `T: ToSql` as the `params` argument to `*_raw` methods.

Adds the a helper trait `BorrowToSql` that is only implemented for both `&dyn ToSql` and `T: ToSql`. This trait shouldn't be implemented by users.

Maybe there's a better name for the trait, I chose it because I works as `Borrow<dyn ToSql>` if that was possible.

Fixes #683

### DIFF
--- a/postgres-types/src/lib.rs
+++ b/postgres-types/src/lib.rs
@@ -108,6 +108,7 @@
 #![doc(html_root_url = "https://docs.rs/postgres-types/0.1")]
 #![warn(clippy::all, rust_2018_idioms, missing_docs)]
 
+use crate::sealed::Sealed;
 use fallible_iterator::FallibleIterator;
 use postgres_protocol::types::{self, ArrayDimension};
 use std::any::type_name;
@@ -952,17 +953,28 @@ fn downcast(len: usize) -> Result<i32, Box<dyn Error + Sync + Send>> {
     }
 }
 
-/// A helper trait to be able create a parameters iterator from `&dyn ToSql` or `T: ToSql`
-pub trait BorrowToSql {
+mod sealed {
+    pub trait Sealed {}
+}
+
+/// A helper trait used internally by Rust-Postgres
+/// to be able create a parameters iterator from `&dyn ToSql` or `T: ToSql`.
+///
+/// /// This cannot be implemented outside of this crate.
+pub trait BorrowToSql: sealed::Sealed {
     /// Get a reference to a `ToSql` trait object
     fn borrow_to_sql(&self) -> &dyn ToSql;
 }
+
+impl Sealed for &dyn ToSql {}
 
 impl BorrowToSql for &dyn ToSql {
     fn borrow_to_sql(&self) -> &dyn ToSql {
         *self
     }
 }
+
+impl<T: ToSql> Sealed for T {}
 
 impl<T: ToSql> BorrowToSql for T {
     fn borrow_to_sql(&self) -> &dyn ToSql {
--- a/postgres-types/src/lib.rs
+++ b/postgres-types/src/lib.rs
@@ -951,3 +951,21 @@ fn downcast(len: usize) -> Result<i32, Box<dyn Error + Sync + Send>> {
         Ok(len as i32)
     }
 }
+
+/// A helper trait to be able create a parameters iterator from `&dyn ToSql` or `T: ToSql`
+pub trait BorrowToSql {
+    /// Get a reference to a `ToSql` trait object
+    fn borrow_to_sql(&self) -> &dyn ToSql;
+}
+
+impl BorrowToSql for &dyn ToSql {
+    fn borrow_to_sql(&self) -> &dyn ToSql {
+        *self
+    }
+}
+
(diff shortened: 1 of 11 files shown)


postgres-types/src/lib.rs
```diff
@@ -951,3 +951,21 @@ fn downcast(len: usize) -> Result<i32, Box<dyn Error + Sync + Send>> {
         Ok(len as i32)
     }
 }
+
+/// A helper trait to be able create a parameters iterator from `&dyn ToSql` or `T: ToSql`
+pub trait BorrowToSql {
```
sfackler, line 956: It's probably best to make this opaque and seal it like we do for `RowIndex` just in case.

postgres/src/client.rs
```diff
@@ -263,10 +263,11 @@ impl Client {
     /// # Ok(())
```
sfackler, line 263: You should be able to update the example above to remove the `.map(|p| ...)` bit.

postgres-types/src/lib.rs
```diff
@@ -951,3 +952,32 @@ fn downcast(len: usize) -> Result<i32, Box<dyn Error + Sync + Send>> {
...
 }
+
+mod sealed {
+    pub trait Sealed {}
+}
+
+/// A helper trait used internally by Rust-Postgres
+/// to be able create a parameters iterator from `&dyn ToSql` or `T: ToSql`.
+///
+/// /// This cannot be implemented outside of this crate.
```
sfackler, line 963:
```suggestion
/// This cannot be implemented outside of this crate.
```

Stargateur:
this make impossible to use `query_raw` with mixed slice type.

Stargateur:
no I confirm this make the api way more limited to use

Stargateur:
in the end I need to not use `query_raw` or implement a small enum that wrap my two type that will be them converted to dynamic dispatch... we finally do a static dynamic dispatch !

Stargateur:
is there any actual test that prove this can be used with `&dyn ToSql` ?

Stargateur:
seem this is what is needed:
```rust
let params: [&dyn ToSql; 2] = &[&last_one, &(MAX as i64)];
        client
            .query_raw(&statement, params)
            .await
            .context(PostgreSQLError)?
```

my real problem seem to be unrelated to this, nevermind I need to dig more.

Stargateur:
still doesn't work I run into `Sync` issue

Stargateur:
I end up copied https://github.com/sfackler/rust-postgres/blob/ff7d88f9013e4b2f2575993e72609d4a5efa6cca/tokio-postgres/src/lib.rs#L248 and it's work. I have literally zero clue why.

PR #696, opened 2020-11-15
WIP replication support

jeff-davis:
This is a rough draft to implement some of the ideas I had to improve on #652.

This is not meant to necessarily replace that PR, it's just to throw some code behind some of the ideas I expressed on that #652.


Martichou:
That's an amazing work that you've done ! 

Did you tested your implementation ? During my tests, it works fine for exactly one minute.
After some seconds, I'm getting a `keepalive.reply() == 1` and some time after, a `connection error: Error { kind: Closed, cause: None }`.

I'm using the "main" from without your sources, I don't ask for any support, just to know if you've encountered the same issue or if it's something else from my side.
> [Logs](https://gist.github.com/Martichou/464eed20cc0a17910473ee4191bf52a7)

jeff-davis:
Hi,

It looks like your code is not sending the keepalive reply, so the server
is simply closing the connection. That's understandable because I don't
currently document and test this case; I need to improve the PR at my next
chance.

Please check your server logs and see if there is any protocol error or
anything else. If so, it's a different problem and please tell me.

Regards,
     Jeff

petrosagg:
Hey @jeff-davis! Thanks for all this great work!

Over at materialize we're working on integrating with postgres replication directly and I've starting working on your PR. I've cleaned up all fixups in your commit log, rebased on top of master, added support for running simple queries over the replication protocol and added support for decoding the output of the `pgoutput `replication plugin.

If you're ok with it I can open a new PR and take over the upstreaming effort, there are a few more things to be done like adding support for replication in the sync client, tests, etc that I'm going to get to in the next couple of weeks.

You can see the current state of the work here: https://github.com/sfackler/rust-postgres/compare/master...petrosagg:replication

Let me know what you think!

jeff-davis:
Thank you for helping!

I hate to nag, but do you know a way to ping sfackler to see if he can
update the image so that it supports replication connections and logical
replication, that way it will pass in CI? And more generally, I haven't
gotten much feedback on this PR from him, so I don't know whether it's on
track to be merged with a bit more work, or whether I'm going in the wrong
direction. He probably just hasn't had a chance to review it.

But if the work is wanted, I'm happy to continue hacking on it. I learned a
lot about the postgres replication protocol, and I patched postgres in an
area that was not properly supporting pipelining. I can review your work
probably next week.

jeff-davis:
> Hey @jeff-davis! Thanks for all this great work!

Hi!
 
> Over at materialize we're working on integrating with postgres replication directly and I've starting working on your PR. I've cleaned up all fixups in your commit log, rebased on top of master, added support for running simple queries over the replication protocol and added support for decoding the output of the `pgoutput `replication plugin.

Sounds interesting.
 
> If you're ok with it I can open a new PR and take over the upstreaming effort, there are a few more things to be done like adding support for replication in the sync client, tests, etc that I'm going to get to in the next couple of weeks.

If you have the cycles to do so that would be much appreciated.
 
> You can see the current state of the work here: [master...petrosagg:replication](https://github.com/sfackler/rust-postgres/compare/master...petrosagg:replication)
> 
> Let me know what you think!

I think you broke logical decoding, e.g. wal2json. This part is a bit confusing in PG-land, so let me try to explain:

There are two types of replication connections: physical replication and logical replication. A physical replication connection is for physical replication, but a logical replication connection may be for **either** logical *replication* or logical *decoding*. "Logical replication" is a special case of logical decoding that uses the logical replication messages in the payload. But if you are doing some other kind of logical decoding, e.g. wal2json, then it uses an arbitrary payload (e.g. json) and doesn't use the logical replication messages at all.

Also, the arbitrary simple queries are only allowed over a logical replication connection. It's possible we may want to use the type system to enforce this, or we could just let the backend throw an error if the client tries to send an arbitrary query over a physical replication connection.

petrosagg:
> If you have the cycles to do so that would be much appreciated.

Awesome! I'll open a new PR soon.

> I think you broke logical decoding, e.g. wal2json. This part is a bit confusing in PG-land, so let me try to explain:

You're right, this was just me being lazy for a PoC I wanted to build with `pgoutput`. I was meaning to separate the plugin implementation out under a trait so that this project can support a couple official ones and let users define their own decoding implementations. I just added that to my branch now so `wal2json` is usable again :) See commits https://github.com/petrosagg/rust-postgres/commit/d46ada07a3e6d1d68ed0857630992a2b6b0014fe and https://github.com/petrosagg/rust-postgres/commit/48acadbdffb69c9f520404ed3d2886805015b154

> but a logical replication connection may be for either logical replication or logical decoding.

I don't think that's quite right. Logical replication is also doing logical decoding, it's just that the plugin used (pgoutput) is shipped with postgres itelf. Here is the [relevant documentation](https://www.postgresql.org/docs/13/logical-replication-architecture.html) (quote: *The walsender process starts logical decoding (described in Chapter 48) of the WAL and loads the standard logical decoding plugin (pgoutput)*)

> Also, the arbitrary simple queries are only allowed over a logical replication connection.

Since there is only one type of connection for any decoding I think you can issue queries regardless. I haven't tested this though so if there is a relevant part of the docs please send it over!

jeff-davis:
> Awesome! I'll open a new PR soon.

Great! 

> You're right, this was just me being lazy for a PoC I wanted to build with `pgoutput`. I was meaning to separate the plugin implementation out under a trait so that this project can support a couple official ones and let users define their own decoding implementations. I just added that to my branch now so `wal2json` is usable again :) See commits [petrosagg@d46ada0](https://github.com/petrosagg/rust-postgres/commit/d46ada07a3e6d1d68ed0857630992a2b6b0014fe) and [petrosagg@48acadb](https://github.com/petrosagg/rust-postgres/commit/48acadbdffb69c9f520404ed3d2886805015b154)

That looks like a good direction. FWIW I would focus on getting the basic replication stuff in (with a payload of bytes), but I like where you are taking it. It would be good if @sfackler could comment on the overall approach and how he'd like to see replication support land. rust-postgres is a widely-used crate and I'm guessing that he wants to be careful about breaking changes in the future.

> I don't think that's quite right. Logical replication is also doing logical decoding, it's just that the plugin used (pgoutput) is shipped with postgres itelf.

Yes. I was trying to get that across, but I think I just added to the confusion. Sorry.

> > Also, the arbitrary simple queries are only allowed over a logical replication connection.
> 
> Since there is only one type of connection for any decoding I think you can issue queries regardless. I haven't tested this though so if there is a relevant part of the docs please send it over!

You can issue queries in logical replication, but not physical. Here's a link to the source:

https://github.com/postgres/postgres/blob/master/src/backend/replication/walsender.c#L1571

Non-replication SQL commands are allowed if it's logical replication/decoding, but it will throw an error if it's a physical replication connection.

Maybe an error there is fine; it doesn't break the protocol.

videni:
@jeff-davis , what is the lastest status?  I also need this to implement the CDC  feature for a rust app, instead of using the Debezium suite which is tool complex for our use case.

tokio-postgres/tests/test/replication.rs
```diff
@@ -0,0 +1,230 @@
...
+
+    let _nrows = sclient
+        .execute("insert into test_logical_replication values(1)", &[])
+        .await
+        .unwrap();
+
+    let mut got_begin = false;
+    let mut got_insert = false;
+    let mut got_commit = false;
+    while let Some(replication_message) = logical_stream.next().await {
```
bendavidsonku, line 122: @jeff-davis I've been pouring over this PR the past couple of days trying to understand everything and there's one thing I'm not quite connecting. Forgive me, as I'm somewhat new to Rust. Isn't `logical_stream.next()` going to wait indefinitely until there is some copy data to consume from the server? 

The reason I ask is because, if we look at [pg_recvlogical](https://github.com/postgres/postgres/blob/master/src/bin/pg_basebackup/pg_recvlogical.c), we don't want to wait indefinitely on copy data because there may be a need to send standby data to postgres (e.g. if we've waited some long amount of time in between receiving copy data) so that postgres doesn't kill our replication connection. 

Apologies if I'm missing something obvious here.

Venryx: I am not experienced in this subject, but from what I observed while I was [integrating the branch](https://github.com/sfackler/rust-postgres/issues/116#issuecomment-1018486358) by @petrosagg, is that postgres sends out "keepalive" messages every 20 seconds or so.

Interestingly, check [this doc page](https://www.postgresql.org/docs/10/protocol-replication.html) about those keepalive messages; the last bit in the "Primary keepalive message (B)" structure tells the client if postgres detects that the client hasn't communicated in a while, and is nearing the point where it will close the connection.

So what I have [my code do](https://github.com/debate-map/app/blob/afc6467b6c6c961f7bcc7b7f901f0ff5cd79d440/Packages/app-server-rs/src/pgclient.rs#L66) at the moment, is just wait for a keepalive message with that "timeout imminent" flag, and then immediately send a response to the server saying "I am still connected". Note that postgres is actually pretty lenient; it gives another 10 or 15 seconds before it actually closes the connection, which is enough time for this "reactionary" approach to be reliable (during the several hours I've tested it anyway).

From what I can tell, this "notify postgres we're still online" feature is not present in this pull-request's code -- so I guess that's something @petrosagg is thinking can be handled by higher level modules/crates. (it might be worth having at least a warning about this behavior in the base `rust-postgres` library however, eg. if it detects a disconnect, and it can see that it was due to the client not sending any responses during the timeout period)

petrosagg: @Venryx unfortunately it is not enough to wait for postgres to send your a keepalive with that bit set in order to send a keepalive, you actually need to proactively be sending them in a timely manner. This is exactly what we do in materialize. 

Check out this code https://github.com/MaterializeInc/materialize/blob/main/src/dataflow/src/source/postgres.rs#L303-L329

and this thread https://www.postgresql.org/message-id/CAMsr+YE2dSfHVr7iEv1GSPZihitWX-PMkD9QALEGcTYa+sdsgg@mail.gmail.com

Venryx: > @Venryx unfortunately it is not enough to wait for postgres to send your a keepalive with that bit set in order to send a keepalive, you actually need to proactively be sending them in a timely manner. This is exactly what we do in materialize.

Ah, interesting; thank you for mentioning. (At my project's current scale, I think it will be a long time before there are transactions of a large enough size for queuing to cause problems with the keepalive responder; but it's still a good thing to fix, to not cause issues for other developers [or myself] down the road.)

guerinoni:
What is the current status? I'm interested in this for switching from libpq :) and searching something with `pgoutput`

azlev:
Hello. Can we revive this PR? Maybe split in smaller parts to other people like me to get involved?

PR #702, opened 2020-11-29
Add escape_identifier() and escape_literal().

### LINKED ISSUE #575: Provide function to escape strings provided to Postgres?
I have a program to help instantiate my database, which internally runs:

```
CREATE USER foo WITH PASSWORD user-input-here
```

Since prepared statements (and so $1, $2..) do not work in this context, is there a way that I can escape the password string provided by the user before passing it to the above?

Might this be a useful piece of functionality to provide from this library? (I appreciate that its use should be discouraged).

jeff-davis:
Fix #575 

These functions are generally useful, but also useful specifically for PR #696 which needs to include arbitrary identifiers in a SQL string.

### DIFF
--- a/postgres-protocol/src/escape/mod.rs
+++ b/postgres-protocol/src/escape/mod.rs
@@ -1,0 +1,93 @@
+//! Provides functions for escaping literals and identifiers for use
+//! in SQL queries.
+//!
+//! Prefer parameterized queries where possible; see
+//! [`Client::query`](crate::Client::query). Do not escape parameters.
+
+#[cfg(test)]
+mod test;
+
+/// Escape a literal and surround result with single quotes. Not
+/// recommended in most cases.
+///
+/// If input contains backslashes, result will be of the form `
+/// E'...'` so it is safe to use regardless of the setting of
+/// standard_conforming_strings.
+pub fn escape_literal(input: &str) -> String {
+    escape_internal(input, false)
+}
+
+/// Escape an identifier and surround result with double quotes.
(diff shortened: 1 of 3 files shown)


postgres-protocol/src/escape/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+#[cfg(test)]
+mod test;
+
+/// Escape a literal and surround result with single quotes. Not
+/// recommended in most cases.
+///
+/// If input contains backslashes, result will be of the form `
+/// E'...'` so it is safe to use regardless of the setting of
+/// standard_conforming_strings.
+pub fn escape_literal(input: &str) -> String {
```
sfackler, line 16: Can we have these return a `Cow<str>` so the no escaping needed case doesn't need to reallocate?

jeff-davis: `escape_literal()` follows the semantics of `PQescapeLiteral()`, which always allocates because the result must be surrounded by quote marks. This is different from `PQescapeStringConn()` which has some messy history that I don't think we need to follow because we should be using parameters most of the time anyway. In those cases where we still need it (like utility commands or replication commands), either form will suffice.

We could potentially optimize `escape_identifier()`, but it would deviate from libpq. I don't see this as likely to be a hot path, so the trade-off doesn't seem great to me.

sfackler: Ah yeah I missed that quotes were always added. The behavior here makes sense in that case.

PR #708, opened 2020-12-01
Fix tokio features for runtime feature in the docs

fanatid:


tokio-postgres/src/lib.rs
```diff
@@ -103,7 +103,7 @@
 //!
 //! | Feature | Description | Extra dependencies | Default |
 //! | ------- | ----------- | ------------------ | ------- |
-//! | `runtime` | Enable convenience API for the connection process based on the `tokio` crate. | [tokio](https://crates.io/crates/tokio) 0.2 with the features `dns`, `net` and `time` | yes |
+//! | `runtime` | Enable convenience API for the connection process based on the `tokio` crate. | [tokio](https://crates.io/crates/tokio) 0.3 with the `io-util` feature | yes |
```
sfackler, line 106: The `io-util` feature is always enabled. `runtime` turns on `net` and `time`: https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/Cargo.toml#L28

fanatid: Oh, I had seen it but then lost it from my eyes when edited file :man_facepalming: 
`dns` feature still should be removed in the docs?

sfackler: Yeah, the dns feature doesn't exist anymore.

sfackler:
Thanks!

PR #710, opened 2020-12-11
Include message object in UnexpectedMessage error

jeff-davis:
Make it easier to debug unexpected messages in the protocol flow.

### DIFF
--- a/postgres-protocol/src/message/backend.rs
+++ b/postgres-protocol/src/message/backend.rs
@@ -73,6 +73,7 @@ impl Header {
 
 /// An enum representing Postgres backend messages.
 #[non_exhaustive]
+#[derive(Debug, PartialEq)]
 pub enum Message {
     AuthenticationCleartextPassword,
     AuthenticationGss,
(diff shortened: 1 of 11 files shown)


sfackler:
I'm a bit worried about large and/or sensitive messages getting stored in the error and propagated around. How about just storing the message tag? That should be enough for a developer to figure out which unexpected message came in without having those issues.

jeff-davis:
Ideally, the `{:?}` of the message would be the enum variant name, but I don't see a simple way to add that kind of reflection without a macro.

I could add the macro, or if you don't tihnk it's worthwhile, I'm OK dropping this patch. My replication client code is working pretty well after I sorted out some under-specified areas of the protocol.

sfackler:
An enum would work fine as well, I'd just prefer it to be a separate enum than the one that actually stores the data.

I personally just use Wireshark to see what's going on when I hit a protocol error.

jeff-davis:
Yeah, it doesn't seem worth the trouble to fix this. Closing.

PR #711, opened 2020-12-11
Support connection validation with timeout

zach-com:
This pull request arose from https://github.com/sfackler/r2d2/issues/116.

### DIFF
--- a/postgres/src/client.rs
+++ b/postgres/src/client.rs
@@ -4,6 +4,7 @@ use crate::{
     ToStatement, Transaction, TransactionBuilder,
 };
 use std::task::Poll;
+use std::time::Duration;
 use tokio_postgres::tls::{MakeTlsConnect, TlsConnect};
 use tokio_postgres::types::{BorrowToSql, ToSql, Type};
 use tokio_postgres::{Error, Row, SimpleQueryMessage, Socket};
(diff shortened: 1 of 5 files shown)


tokio-postgres/src/client.rs
```diff
@@ -450,6 +450,15 @@ impl Client {
         self.simple_query_raw(query).await?.try_collect().await
     }
 
+    /// Validates connection, timing out after specified duration.
+    pub async fn is_valid(&self, timeout: Duration) -> Result<(), Error> {
```
sfackler, line 454: Since timeouts are so easy in the async world, I'd prefer to only expose this API in the blocking client.

postgres/src/client.rs
```diff
@@ -413,6 +413,11 @@ impl Client {
         self.connection.block_on(self.client.simple_query(query))
     }
 
+    /// Validates connection, timing out after specified duration.
+    pub fn is_valid(&mut self, timeout: std::time::Duration) -> Result<(), Error> {
```
sfackler, line 417: Nit: import `Duration` with a `use std::time::Duration` up above.

tokio-postgres/src/client.rs
zach-com: Does Rust allow a module to expose functions to the workspace but not beyond? Otherwise, we would have to introduce `async` in the `postgres` crate, right?

sfackler: It does not, but that's not a problem. There's already async code in the postgres crate in the block_on calls.

tokio-postgres/src/bind.rs
```diff
@@ -20,8 +20,9 @@ where
     I: IntoIterator<Item = P>,
     I::IntoIter: ExactSizeIterator,
 {
+    type BytesResult = Result<bytes::Bytes, Error>;
```
sfackler, line 23: What is this change for?

tokio-postgres/src/error/mod.rs
```diff
@@ -492,3 +495,9 @@ impl Error {
         Error::new(Kind::Connect, Some(Box::new(e)))
     }
 }
+
+impl From<Elapsed> for Error {
```
sfackler, line 499: I'd rather have a `#[doc(hidden)]` explicit constructor rather than a From impl.

tokio-postgres/src/bind.rs
zach-com: After the `impl From` block was introduced, the compiler required the type hint. Replacing the `impl From` block will make this change unnecessary.

sfackler:
Thanks!

PR #717, opened 2020-12-24
deps: upgrade to tokio v1.0 ecosystem

benesch:
@sfackler this is still waiting on releases of both tokio-openssl and tokio-native-tls that include the Tokio v1.0 bump, but I got tests for tokio-postgres and postgres passing, so figured I'd put this up.

### DIFF
--- a/postgres-openssl/src/lib.rs
+++ b/postgres-openssl/src/lib.rs
@@ -48,8 +48,10 @@ use openssl::hash::MessageDigest;
 use openssl::nid::Nid;
 #[cfg(feature = "runtime")]
 use openssl::ssl::SslConnector;
-use openssl::ssl::{ConnectConfiguration, SslRef};
-use std::fmt::Debug;
+use openssl::ssl::{self, ConnectConfiguration, SslRef};
+use openssl::x509::X509VerifyResult;
+use std::error::Error;
+use std::fmt::{self, Debug};
 use std::future::Future;
 use std::io;
 use std::pin::Pin;
@@ -57,7 +59,7 @@ use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
 use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
-use tokio_openssl::{HandshakeError, SslStream};
+use tokio_openssl::SslStream;
 use tokio_postgres::tls;
 #[cfg(feature = "runtime")]
 use tokio_postgres::tls::MakeTlsConnect;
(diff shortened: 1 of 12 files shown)


benesch:
Supersedes #688 and #716.

sfackler:
tokio-openssl is updated.

tokio-postgres/src/binary_copy.rs
```diff
@@ -153,7 +153,7 @@ impl Stream for BinaryCopyOutStream {
             Some(header) => header.has_oids,
             None => {
                 check_remaining(&chunk, HEADER_LEN)?;
-                if &chunk.bytes()[..MAGIC.len()] != MAGIC {
+                if &chunk.chunk()[..MAGIC.len()] != MAGIC {
```
sfackler, line 156: You should be able to just do `if !chunk.starts_with(&MAGIC)` here I think.

postgres/src/copy_out_reader.rs
```diff
@@ -46,7 +46,7 @@ impl BufRead for CopyOutReader<'_> {
             };
         }
 
-        Ok(self.cur.bytes())
+        Ok(self.cur.chunk())
```
sfackler, line 49: I think this can just be `Ok(&self.cur)`.

benesch: Done.

tokio-postgres/src/binary_copy.rs
benesch: `chunk` is a `Cursor<Bytes>` which unfortunately does not implement `AsRef<[u8]>`. Switched to `!chunk.chunk().starts_with(&MAGIC)`. 

The trait method `Buf::chunk` is not documented to return the whole slice, but the implementation of `Buf` for `Cursor<Bytes>` does make that guarantee. If relying on that is scary, we could instead do `chunk.read_bytes(MAGIC.len()) != MAGIC`, which is safe for any implementation of `Buf`.

sfackler: Oh - we should just remove the Cursor layer entirely since Bytes implements Buf directly now.

benesch: I tried that, but we need access to the cursor position: https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/src/binary_copy.rs#L201

Could always do some pointer arithmetic to compute the position, but didn’t seem like an obvious win.

benesch:
All the TLS related crates have new releases now, so this is ready to go pending whatever review feedback you might have, @sfackler. As always feel free to just take over the PR.

tokio-postgres/src/binary_copy.rs
sfackler: Ah right - should be fine as-is then.

postgres-openssl/src/lib.rs
```diff
@@ -134,13 +134,15 @@ where
...
+    type Error = Error;
     #[allow(clippy::type_complexity)]
-    type Future = Pin<Box<dyn Future<Output = Result<TlsStream<S>, HandshakeError<S>>> + Send>>;
+    type Future = Pin<Box<dyn Future<Output = Result<TlsStream<S>, Error>> + Send>>;
 
     fn connect(self, stream: S) -> Self::Future {
         let future = async move {
-            let stream = tokio_openssl::connect(self.ssl, &self.domain, stream).await?;
+            let ssl = self.ssl.into_ssl(&self.domain)?;
+            let mut stream = Box::pin(SslStream::new(ssl, stream)?);
```
sfackler, line 144: Since `S` is `Unpin`, `SslStream<S>` is as well so there's no need to use `Box::pin` here. You should be able to just do `Pin::new(&mut stream).connect().await;`

postgres-openssl/src/lib.rs
```diff
@@ -134,13 +134,15 @@ where
...
+            stream.as_mut().connect().await?;
```
sfackler, line 145: I think it's worth doing something just a little bit more complicated to get better diagnostics after a certificate validation failure: https://github.com/sfackler/hyper-openssl/blob/master/src/lib.rs#L237-L240

postgres-openssl/src/lib.rs
```diff
@@ -134,13 +134,15 @@ where
     S: AsyncRead + AsyncWrite + Unpin + Debug + 'static + Sync + Send,
```
sfackler, line 134: You can remove the `Debug + 'static + Sync` bounds from here now.

sfackler:
Thanks! Looks good other than those few nits.

I think I'm probably just going to immediately cut a release with these dependency upgrades to get that out the door, and then start looking into cleaning things up for a 1.0 release from there.

postgres-openssl/src/lib.rs
benesch: Ah yeah, thanks, that's much nicer. I missed the `Unpin` bound on `S`.

postgres-openssl/src/lib.rs
benesch: Ok, `Debug` and `Sync` are gone. Compiler still insists on `'static` though.

postgres-openssl/src/lib.rs
benesch: SGTM! I just shamelessly copy/pasted that code into here.

PR #726, opened 2021-01-07
PgLsn type.

jeff-davis:
This type will be useful for implementing the replication protocol.

### DIFF
--- a/postgres-types/src/pg_lsn.rs
+++ b/postgres-types/src/pg_lsn.rs
@@ -1,0 +1,79 @@
+//! Log Sequence Number (LSN) type for PostgreSQL Write-Ahead Log
+//! (WAL), also known as the transaction log.
+
+use bytes::BytesMut;
+use postgres_protocol::types;
+use std::error::Error;
+use std::fmt;
+use std::str::FromStr;
+
+use crate::{FromSql, IsNull, ToSql, Type};
+
+/// Postgres `PG_LSN` type.
+#[derive(Clone, Copy, Eq, Ord, PartialEq, PartialOrd)]
+pub struct PgLsn(u64);
+
+/// Error parsing LSN.
+#[derive(Debug)]
+pub struct ParseLsnError(());
+
+impl From<u64> for PgLsn {
+    fn from(lsn_u64: u64) -> Self {
+        PgLsn(lsn_u64)
+    }
+}
+
+impl From<PgLsn> for u64 {
+    fn from(lsn: PgLsn) -> u64 {
+        lsn.0
+    }
+}
+
+impl FromStr for PgLsn {
+    type Err = ParseLsnError;
+
+    fn from_str(lsn_str: &str) -> Result<Self, Self::Err> {
+        let split: Vec<&str> = lsn_str.split('/').collect();
+        if split.len() == 2 {
+            let (hi, lo) = (
+                u64::from_str_radix(split[0], 16).map_err(|_| ParseLsnError(()))?,
+                u64::from_str_radix(split[1], 16).map_err(|_| ParseLsnError(()))?,
+            );
+            Ok(PgLsn((hi << 32) | lo))
+        } else {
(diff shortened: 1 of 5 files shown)


postgres-types/src/pg_lsn.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+}
+
+impl From<PgLsn> for u64 {
+    fn from(lsn: PgLsn) -> u64 {
+        lsn.0
+    }
+}
+
+impl From<&str> for PgLsn {
```
sfackler, line 27: This should probably be a FromStr to avoid the unwraps.

postgres-types/src/pg_lsn.rs
```diff
@@ -0,0 +1,69 @@
...
+        assert_eq!(split.len(), 2);
+        let (hi, lo) = (
+            u64::from_str_radix(split[0], 16).unwrap(),
+            u64::from_str_radix(split[1], 16).unwrap(),
+        );
+        PgLsn((hi << 32) | lo)
+    }
+}
+
+impl From<PgLsn> for String {
```
sfackler, line 39: And this can be a `Display` impl instead.

postgres-types/src/pg_lsn.rs
```diff
@@ -0,0 +1,69 @@
...
+    fn from(lsn: PgLsn) -> String {
+        format!("{:X}/{:X}", lsn.0 >> 32, lsn.0 & 0x00000000ffffffff)
+    }
+}
+
+impl fmt::Debug for PgLsn {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("PgLsn").field(&String::from(*self)).finish()
```
sfackler, line 47: The field here can use the display impl when that's implemented.

postgres-types/src/pg_lsn.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+}
+
+impl From<&str> for PgLsn {
+    fn from(lsn_str: &str) -> Self {
+        let split: Vec<&str> = lsn_str.split('/').collect();
+        assert_eq!(split.len(), 2);
+        let (hi, lo) = (
+            u64::from_str_radix(split[0], 16).unwrap(),
+            u64::from_str_radix(split[1], 16).unwrap(),
```
sfackler, line 33: What do these two parts represent? Should there be accessors for them on the type?

jeff-davis: There isn't much practical meaning there. WAL files are normally 16MB, so the low 24 bits are normally the offset within a WAL file. But that's not guaranteed at a protocol level, and a compile-time switch can change the WAL file size.

I think there was a historical idea of a "log file #" (first 4 bytes) and an "offset" (last 4 bytes), but those never really made a lot of sense because they didn't correspond to the WAL file size, so you always needed both to determine the file name.

postgres-types/src/pg_lsn.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use crate::{FromSql, IsNull, ToSql, Type};
+
+/// Postgres `PG_LSN` type.
+#[derive(Clone, Copy, Eq, Ord, PartialEq, PartialOrd)]
+pub struct PgLsn(u64);
+
+/// Error parsing LSN.
+#[derive(Debug)]
+pub struct ParseLsnError;
```
sfackler, line 18: Can you add a private field to this type (e.g. `pub struct ParseLsnError(());` so it can't be constructed externally?

sfackler:
LGTM other than the one last nit!

postgres-types/src/pg_lsn.rs
sfackler: Ah cool, so this is just there to match how Postgres renders the value.

sfackler:
Thanks!

PR #727, opened 2021-01-08
postgres-protocol: use RustCrypto md-5 crate

benesch:
Swap the md5 crate for the md-5 crate.  Despite the latter's somewhat
more suspicious name, it is part of the wider RustCrypto ecosystem, and
shares code with the sha2 crate that postgres-protocol already uses.


sfackler:
Thanks!

PR #728, opened 2021-01-08
Add password module for hashing on the client side.

jeff-davis:
Hashing a password on the client side is useful so that you can set
a user's password without ever sending it in plain text to the
server. This avoids leaking passwords in the log or elsewhere.

### DIFF
--- a/postgres-protocol/src/password/mod.rs
+++ b/postgres-protocol/src/password/mod.rs
@@ -1,0 +1,104 @@
+//! Functions to encrypt a password in the client.
+//!
+//! This is intended to be used by client applications that wish to
+//! send commands like `ALTER USER joe PASSWORD 'pwd'`. The password
+//! need not be sent in cleartext if it is encrypted on the client
+//! side. This is good because it ensures the cleartext password won't
+//! end up in logs pg_stat displays, etc.
+
+use crate::authentication::sasl;
+use hmac::{Hmac, Mac, NewMac};
+use md5::Md5;
+use rand::RngCore;
+use sha2::digest::FixedOutput;
+use sha2::{Digest, Sha256};
+
+#[cfg(test)]
+mod test;
+
+const SCRAM_DEFAULT_ITERATIONS: u32 = 4096;
+const SCRAM_DEFAULT_SALT_LEN: usize = 16;
+
+/// Hash password using SCRAM-SHA-256 with a randomly-generated
+/// salt.
+///
+/// The client may assume the returned string doesn't contain any
+/// special characters that would require escaping in an SQL command.
+pub fn scram_sha_256(password: &[u8]) -> String {
+    let mut salt: [u8; SCRAM_DEFAULT_SALT_LEN] = [0; SCRAM_DEFAULT_SALT_LEN];
+    let mut rng = rand::thread_rng();
+    rng.fill_bytes(&mut salt);
+    scram_sha_256_salt(password, salt)
+}
+
+// Internal implementation of scram_sha_256 with a caller-provided
+// salt. This is useful for testing.
+pub(crate) fn scram_sha_256_salt(password: &[u8], salt: [u8; SCRAM_DEFAULT_SALT_LEN]) -> String {
+    // Prepare the password, per [RFC
+    // 4013](https://tools.ietf.org/html/rfc4013), if possible.
+    //
+    // Postgres treats passwords as byte strings (without embedded NUL
+    // bytes), but SASL expects passwords to be valid UTF-8.
+    //
+    // Follow the behavior of libpq's PQencryptPasswordConn(), and
+    // also the backend. If the password is not valid UTF-8, or if it
+    // contains prohibited characters (such as non-ASCII whitespace),
+    // just skip the SASLprep step and use the original byte
+    // sequence.
+    let prepared: Vec<u8> = match std::str::from_utf8(password) {
+        Ok(password_str) => {
+            match stringprep::saslprep(password_str) {
+                Ok(p) => p.into_owned().into_bytes(),
+                // contains invalid characters; skip saslprep
+                Err(_) => Vec::from(password),
+            }
+        }
+        // not valid UTF-8; skip saslprep
+        Err(_) => Vec::from(password),
+    };
+
+    // salt password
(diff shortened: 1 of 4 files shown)


jeff-davis:
I feel like this could be organized better to separate out the scram code from the sasl code. But I noticed that there's also a scram crate, but it doesn't seem to provide exactly what we need. Do you think we should refactor the code we have, or try to improve the scram crate and then depend on that?

sfackler:
I remember looking at the scram crate when implementing the SASL support, but the API didn't work out for what we needed here. Can't remember the details though unfortunately.

postgres-protocol/src/misc/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+
+#[cfg(test)]
+mod test;
+
+const SCRAM_DEFAULT_ITERATIONS: u32 = 4096;
+const SCRAM_DEFAULT_SALT_LEN: usize = 16;
+
+/// Choose the password encryption algorithm.
+#[derive(Debug)]
+pub enum PasswordAlgorithm {
```
sfackler, line 16: This should be `#[non_exhaustive]` so we can add more variants in the future.

postgres-protocol/src/misc/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+        PasswordAlgorithm::ScramSha256 => {
+            let prepared_passwd = stringprep::saslprep(passwd)
+                .map_err(|_| EncryptionError)?
+                .into_owned();
+            Ok(scram_sha_256(prepared_passwd.into_bytes().as_slice()))
+        }
+    }
+}
+
+fn scram_sha_256(prepared_passwd: &[u8]) -> String {
```
sfackler, line 57: Yeah, we should be able to deduplicate this code with the SASL implementation somehow I think.

postgres-protocol/src/misc/mod.rs
```diff
@@ -0,0 +1,92 @@
+//! As always, there are some functions that just don't fit anywhere.
```
sfackler, line 1: I'd probably call this module `password` or something like that.

jeff-davis:
I ended up reworking the API; I wasn't very happy with it. MD5 and SCRAM-SHA-256 are different, and it doesn't make a lot of sense to try to put them in the same API, especially when nobody should be using MD5 anyway.

* MD5 will never return an error; SCRAM-SHA-256 will reject passwords with prohibited characters (e.g. `NUL`).
  * arguably, MD5 should also reject `NUL` because it's invalid in a C string (so implicitly rejected by libpq, too), but I don't think we should spend a lot of effort trying to deal with such cases for an algorithm that is outdated anyway
* MD5 uses the username as a salt; SCRAM-SHA-256 generates its own salt randomly
* The randomness in SCRAM-SHA-256 makes it non-deterministic, so we need a deterministic entry point for testing (where the caller provides a salt).

There's another question about how hard we should work to support passwords that are invalid UTF-8 or that contain prohibited characters. libpq falls back to just skip the saslprep step in both cases. Currently, I don't see a way for rust-postgres to handle non-utf8 passwords, anyway (though perhaps it should?), and including prohibited characters in a password seems like an extreme edge case.

postgres-protocol/src/misc/mod.rs
jeff-davis: Resolved by reworking the API.

postgres-protocol/src/password/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+///
+/// This is useful if the caller has access to a better source
+/// of randomness.
+///
+/// An error will be raised if the password contains characters
+/// prohibited by [RFC 4013](https://tools.ietf.org/html/rfc4013).
+///
+/// The client may assume the returned string doesn't contain any
+/// special characters that would require escaping in an SQL command.
+pub fn scram_sha_256_salt(
```
sfackler, line 52: I would probably just make this private for use by the test.

postgres-protocol/src/password/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+use sha2::digest::FixedOutput;
+use sha2::{Digest, Sha256};
+
+#[cfg(test)]
+mod test;
+
+const SCRAM_DEFAULT_ITERATIONS: u32 = 4096;
+const SCRAM_DEFAULT_SALT_LEN: usize = 16;
+
+/// Error encrypting password.
```
sfackler, line 22: Pedantic nit: this isn't really encryption since it's a one-way function.

postgres-protocol/src/password/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/// Encrypt password using SCRAM-SHA-256 with a caller-provided
+/// salt.
+///
+/// This is useful if the caller has access to a better source
+/// of randomness.
+///
+/// An error will be raised if the password contains characters
+/// prohibited by [RFC 4013](https://tools.ietf.org/html/rfc4013).
```
sfackler, line 48: For backwards compatibility, Postgres's version of SCRAM works with arbitrary byte sequences and just skips the saslprep step for invalid ones: https://github.com/sfackler/rust-postgres/blob/master/postgres-protocol/src/authentication/sasl.rs#L20-L23.

So this function should probably just take a `&[u8]` instead of a `&str` to support that.

sfackler:
Splitting into 2 functions seems reasonable to me.

postgres-protocol/src/password/mod.rs
jeff-davis: Done. This answers some of the other problems, as well, because now scram_sha_256 can't return an error. I still left the functions separate, because it seems like slightly better ergonomics.

sfackler: I think you may need to push again?

jeff-davis: Pushed, sorry.

sfackler:
Thanks!

Neustradamus:
@jeff-davis, @sfackler: Thanks a lot for all about SCRAM.

Linked to:
- https://github.com/scram-xmpp/info/issues/1

PR #733, opened 2021-01-14
Added support for `geo-types` 0.7 via `with-geo-types-0_7` feature

michaelkirk:


postgres-types/src/geo_types_06.rs
```diff
@@ -1,6 +1,9 @@
 use bytes::BytesMut;
 use fallible_iterator::FallibleIterator;
+#[cfg(feature = "geo-types-06")]
 use geo_types_06::{Coordinate, LineString, Point, Rect};
+#[cfg(feature = "geo-types-0_7")]
+use geo_types_0_7::{Coordinate, LineString, Point, Rect};
```
michaelkirk, line 6: For the purposes of this crate, the 0.7 API is compatible with 0.6 so I opted to re-use the module.

If you'd prefer that I instead duplicate the implementation into a separate file, let me know.

michaelkirk:
Sorry, I missed some tests. Closing for now.

PR #734, opened 2021-01-25
Added support for `geo-types` 0.7 via `with-geo-types-0_7` feature

michaelkirk:


postgres-types/Cargo.toml
```diff
@@ -15,7 +15,7 @@ derive = ["postgres-derive"]
 with-bit-vec-0_6 = ["bit-vec-06"]
 with-chrono-0_4 = ["chrono-04"]
 with-eui48-0_4 = ["eui48-04"]
-with-geo-types-0_6 = ["geo-types-06"]
+with-geo-types-0_7 = ["geo-types-0_7"]
```
sfackler, line 18: Dropping 0.6 support breaks backwards compatibility. These features were set up so that multiple versions could be supported concurrently.

michaelkirk:
To be clear, your preference is for two mutually exclusive features: one for 0.6 and another for 0.7?

sfackler:
The requirement is for there to be two features, which are not mutually exclusive.

michaelkirk:
Trying to make sure I understand, you want this to work:

cargo test --features "with-geo-types-0_6 with-geo-types-0_7"

sfackler:
Yes. It should be pretty straightforward - you just add parallel structures to the 0.6 stuff that already exists.

tokio-postgres/tests/test/types/mod.rs
```diff
@@ -19,10 +19,10 @@ mod bit_vec_06;
 mod chrono_04;
 #[cfg(feature = "with-eui48-0_4")]
 mod eui48_04;
-#[cfg(feature = "with-geo-types-0_4")]
```
michaelkirk, line 22: I did remove the 0.4 integration from the test module. It seemed unused/vestigial.

michaelkirk (commented):
Ok, I've reworked this to add back the 0.6 integration, so the 0.7 integration is strictly additive.

tokio-postgres/tests/test/types/mod.rs
sfackler: Oops, yeah thanks that was a holdover from a previous release.

sfackler:
Thanks!

Issue #583, opened 2020-03-14
It would be easier to deal with errors if the error kind is made pub


asahaf:
Hi,

Some errors, like closed, column, and row_count,  don't have a cause (None). for these errors, there is no way to know the actual underlayer error as the error kind is not exposed.
It would be easier if the error kind is made public.

Thanks,

hhirtz:
Duplicate of #571?

the-anonymous009:
Any update? This is absolutely necessary to handle error properly when cause is None, especially when connection is closed.

hhirtz:
I mean if it's a matter of removing 4 chars from a file anyone concerned could send a patch.

palfrey:
For reference, [I'd be interested in knowing about the ConfigParse case as I've got a hacky workaround for this right now.](https://github.com/palfrey/wait-for-db/blob/c3fb9dbad94f06f91f83d6ef9e4af0b93ef64083/src/pg.rs#L34)

couchand:
I've run into this from the same place @palfrey seems to: trying to figure out if it's worth retrying the request (c.f. djc/bb8#141).  @sfackler would you be open to a PR simliar to #739 that exposes a method with this information?

One possible difficulty is identifying which kinds of errors should be included.  @palfrey's workaround looks for `Kind::ConfigParse` errors, and @levi-nz's issue implies `Kind::Authentication` should be included.  In a quick review of the enum variants, none stand out to me, but perhaps there are others?

PR #739, opened 2021-02-12
Make error kind public

the-anonymous009:
Sometimes application needs to take different action on different kind of errors but there is no easy way to differentiate errors when cause is None (#583).

### DIFF
--- a/tokio-postgres/src/error/mod.rs
+++ b/tokio-postgres/src/error/mod.rs
@@ -421,6 +421,11 @@ impl Error {
         self.source().and_then(|e| e.downcast_ref::<DbError>())
     }
 
+    /// Determines if the error was associated with closed connection.
+    pub fn is_closed(&self) -> bool {
+        self.0.kind == Kind::Closed
+    }
+
     /// Returns the SQLSTATE error code associated with the error.
     ///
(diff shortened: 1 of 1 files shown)


sfackler:
Which kinds of error in particular do you want to handle differently?

the-anonymous009:
> Which kinds of error in particular do you want to handle differently?

I think in some cases at least Kind::Closed should be handled differently.  For example, if query execution is failed by Kind::Close error the application might want to connect to database few seconds later and try the query again. This behaviour is reasonable for many applications because connection lost is not critical error.  But currently there is no easy way to check if the failure was caused by Kind::Closed. 

Also this can be used to detect connection lost. I know that task from Connection object is terminated as soon as connection is lost, but it would be helpful if connection lost could be detected by error from client method.

the-anonymous009:
What do you think about this issue?

sfackler:
I'd be happy to take a PR adding an `is_closed()` method to `Error`, but I do not want to expose its internal structure. A direct method will be easier to work with for your use case anyway.

the-anonymous009:
Okay I just removed previous commit, synced with latest branch and add is_closed() method.  Could you check this update?

sfackler:
Thanks!

PR #747, opened 2021-03-14
Add parameter and column formatting to query_raw()

### LINKED ISSUE #746: Provide mechanism for letting Postgres itself handle type inference
### Motivation

Postgres itself has great type inference when executing prepared statements without types, e.g.:

```sql
PREPARE coercion_test AS SELECT $1 + 2;
EXECUTE coercion_test('123'); # 125, despite using a string literal
```

Currently, it's unclear how to leverage that type inference when using the `rust-postgres` packages. The ability to leverage that type inference would be useful for supporting Postgres types that don't yet have a robust `ToSql` implementation (e.g. `MONEY`) or for skipping the entire Rust type conversion step where query parameters might not be known ahead of time, or that originate via protocols that can't express the same complex types as Rust (e.g JSON, as mentioned in #668).

###  Prior Art

Similar libraries attempt to leverage this inference as a fallback case where types are not explicitly provided. [`libpq`'s `PQexecParams`](https://www.postgresql.org/docs/current/libpq-exec.html) includes  `paramTypes[]`, which, if left empty or uses a `0` in place of an `Oid` for a param, will infer "a data type for the parameter symbol in the same way it would do for an untyped literal string". [`node-postgres` appears to do a similar thing](https://github.com/brianc/node-postgres/blob/45fa27ea4ae9a9a9cf78b50b325d8da871b1c796/packages/pg-protocol/src/serializer.ts#L123): every type that isn't `null` or a `Buffer` is written directly into the message without any attempt at differentiating between types (as far as I can tell).

### Previous attempts

Trying to implement this myself in userland without messing around with any `rust-postgres` internals, I tried the following:

1. using `prepare_typed` with every param explicitly typed to `Type::TEXT`, as suggested in #668. But this overrides Postgres's inference, rather than handling coercion of string literals into inferred types. Specifically:

    ```rust
    let statement = client.prepare_typed("SELECT $1 + 2", &[Type::TEXT]).await?;
    client.query(&statement, &[&"123"]).await?;
    ```

    ...results in the following error:

    ```
    db error: ERROR: operator does not exist: text + integer
    HINT: No operator matches the given name and argument types. You might need to add explicit type casts.
    ```

2. using an empty array of Types or using a slice of `Type::UNKNOWN` in `prepare_typed`, or using `prepare` or `query` (which calls `prepare` for all non-`Statement`s, it looks like) directly results in a conversion error, e.g.

    ```
    error serializing parameter 0: cannot convert between the Rust type `&&&str` and the Postgres type `int4`
    ```

3. using a `RawValue` that represents a nullable "string literal" and implements `ToSql` such that type-checking by `rust-postgres` is completely bypassed:

    ```rust
   pub struct RawValue<'a> {
     raw: &'a Option<&'a [u8]>
   }

   impl<'a> ToSql for RawValue<'a> {
      fn to_sql(&self, type_: &Type, out: &mut BytesMut) -> Result<IsNull, Error> {
        match self.raw {
          Some(raw) => {
            out.put_slice(raw); // or e.g. <&str as ToSql>::to_sql()
            Ok(IsNull::No)
          },
          None => Ok(IsNull::Yes)
        }
      }

      fn accepts(_: &Type) -> bool {
        true
      }

      fn to_sql_checked(&self, type_: &Type, out: &mut BytesMut) -> Result<IsNull, Error> {
        // YOLO/bypass those type checks
        self.to_sql(type_, out)
      }
    }
    ```
    ...which, predictably, results in various message encoding errors, depending on the specifics of how one attempts to write to the output buffer directly (usually `insufficient data left in message` or `incorrect binary data format in bind parameter 1`). This 👆 is unsurprising, since my understanding is that a) there's no way to prepare an untyped statement, b) preparation is unavoidable as part of the current implementation of `ToStatement` for`str` and c) the current encoding of messages requires alignment of types between those inferred from `prepare` and those written to the output buffer by a `ToSql` implementation (although my understanding of that last point is shaky at best... still navigating my way through the codebase to figure out why/where that association is enforced in the protocol). 

### Next Steps

I'm not an expert on the Postgres message protocol or the internals of this codebase by any means, but here are a few approaches that I thought of to handle this case:

* implement a `Type` for `Oid` `0` (or some equivalent pseudo-type that can be matched on later) so that users can explicitly delegate inference of types during `prepare_typed`
* expose an untyped query method on `Client` (maybe `query_untyped?`)
* cop the `node-postgres` M.O. and delegate type inference down to Postgres by default (<- this is probably not the right approach, since it would probably be a breaking change, but worth mentioning)

It's also entirely possible that it's already simple enough to handle this case using `ToSql` alone, and I'm either misunderstanding how this library works or the Postgres message protocol itself! In which case, it would be nice to document or provide an example for this kind of inference pass-through, as there have already been a few issues in a similar vein (e.g. #668, #722, #377).

Thanks in advance for your help! If it turns out that I'm not misunderstanding the capabilities of Postgres and `rust-postgres` and you'd like to implement this kind of inference as a crate-side feature, then I'd be happy to help out with a PR (with a bit of guidance along the way).

NAlexPear:
## Summary

This PR would close #746 ~by adding a `format(&self)` method to `ToSql`, defaulting to `Format::Binary` for backwards compatibility and ease-of-use~ by adding `param_formats` and `column_formats` to the `query_raw()` method. 

## Notes

~This implementation allocates [two new vectors](https://github.com/sfackler/rust-postgres/compare/master...NAlexPear:parameter-formats?expand=1#diff-75a8b28bb1837403d9e313e5fb5111a321fd4f223e9c150a0bc028313db7fb66R168), which might not be acceptable. Running benchmarks showed no consistent real-world performance differences on my machine, but YMMV. It would be possible to avoid those allocations if `encode_bind` could take an `IntoIterator<Item = &P>` instead of `P`, I think, but that would certainly be a lot more work to get changed.~

### DIFF
--- a/postgres-types/src/lib.rs
+++ b/postgres-types/src/lib.rs
@@ -770,16 +770,12 @@ pub trait ToSql: fmt::Debug {
         ty: &Type,
         out: &mut BytesMut,
     ) -> Result<IsNull, Box<dyn Error + Sync + Send>>;
-
-    /// Specifies the message format of the value
-    fn format(&self) -> Format {
-        Format::Binary
-    }
 }
 
 /// Supported Postgres message format types
 ///
 /// Using Text format in a message assumes a Postgres `SERVER_ENCODING` of `UTF8`
+#[derive(Copy, Clone, Debug)]
 pub enum Format {
     /// Text format (UTF-8)
     Text,
@@ -788,12 +784,18 @@ pub enum Format {
 }
 
 /// Convert from `Format` to the Postgres integer representation of those formats
-impl From<Format> for i16 {
-    fn from(format: Format) -> Self {
+impl From<&Format> for i16 {
+    fn from(format: &Format) -> Self {
         match format {
             Format::Text => 0,
             Format::Binary => 1,
         }
+    }
+}
+
+impl From<Format> for i16 {
+    fn from(format: Format) -> Self {
+        Self::from(&format)
     }
 }
 
--- a/postgres-types/src/lib.rs
+++ b/postgres-types/src/lib.rs
@@ -190,6 +190,7 @@ where
     if !T::accepts(ty) {
         return Err(Box::new(WrongType::new::<T>(ty.clone())));
     }
+
     v.to_sql(ty, out)
 }
 
@@ -741,11 +742,11 @@ pub enum IsNull {
 /// an index offset of 1. **Note:** the impl for arrays only exist when the
 /// Cargo feature `array-impls` is enabled.
 pub trait ToSql: fmt::Debug {
-    /// Converts the value of `self` into the binary format of the specified
+    /// Converts the value of `self` into the `Format` of the specified
     /// Postgres `Type`, appending it to `out`.
     ///
     /// The caller of this method is responsible for ensuring that this type
-    /// is compatible with the Postgres `Type`.
+    /// is compatible with the Postgres `Type` and `Format`.
     ///
     /// The return value indicates if this value should be represented as
     /// `NULL`. If this is the case, implementations **must not** write
@@ -769,6 +770,31 @@ pub trait ToSql: fmt::Debug {
         ty: &Type,
         out: &mut BytesMut,
     ) -> Result<IsNull, Box<dyn Error + Sync + Send>>;
+
+    /// Specifies the message format of the value
+    fn format(&self) -> Format {
+        Format::Binary
+    }
+}
+
+/// Supported Postgres message format types
+///
+/// Using Text format in a message assumes a Postgres `SERVER_ENCODING` of `UTF8`
+pub enum Format {
+    /// Text format (UTF-8)
+    Text,
+    /// Compact, typed binary format
+    Binary,
+}
+
+/// Convert from `Format` to the Postgres integer representation of those formats
+impl From<Format> for i16 {
+    fn from(format: Format) -> Self {
+        match format {
+            Format::Text => 0,
+            Format::Binary => 1,
+        }
+    }
 }
 
 impl<'a, T> ToSql for &'a T
--- a/postgres/src/client.rs
+++ b/postgres/src/client.rs
@@ -6,7 +6,7 @@ use crate::{
 use std::task::Poll;
 use std::time::Duration;
 use tokio_postgres::tls::{MakeTlsConnect, TlsConnect};
-use tokio_postgres::types::{BorrowToSql, ToSql, Type};
+use tokio_postgres::types::{BorrowToSql, Format, ToSql, Type};
 use tokio_postgres::{Error, Row, SimpleQueryMessage, Socket};
 
 /// A synchronous PostgreSQL client.
@@ -221,6 +221,7 @@ impl Client {
     ///
     /// ```no_run
     /// use postgres::{Client, NoTls};
+    /// use postgres::types::Format;
     /// use fallible_iterator::FallibleIterator;
     /// use std::iter;
     ///
@@ -228,7 +229,12 @@ impl Client {
     /// let mut client = Client::connect("host=localhost user=postgres", NoTls)?;
     ///
     /// let baz = true;
-    /// let mut it = client.query_raw("SELECT foo FROM bar WHERE baz = $1", iter::once(baz))?;
+    /// let mut it = client.query_raw(
+    ///     "SELECT foo FROM bar WHERE baz = $1",
+    ///     iter::once(baz),
+    ///     Some(Format::Binary),
+    ///     Some(Format::Binary),
+    /// )?;
     ///
     /// while let Some(row) = it.next()? {
     ///     let foo: i32 = row.get("foo");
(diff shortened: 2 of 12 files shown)


postgres-types/src/lib.rs
```diff
@@ -715,6 +716,31 @@ pub trait ToSql: fmt::Debug {
         ty: &Type,
         out: &mut BytesMut,
     ) -> Result<IsNull, Box<dyn Error + Sync + Send>>;
+
+    /// Specifies the message format of the value
+    fn format(&self) -> Format {
```
sfackler, line 721: I don't know how I feel about exposing this on ToSql, since the same can't really be done for FromSql.

An alternative I've been thinking about is to have a separate query method that operates entirely in "string mode", both for parameters and the resulting rows.

NAlexPear: That's a really good point about the lack of equivalent handle on the returned types. It would be nice to let both parameters and returned types have independently-configurable formats like Postgres itself, but a text-only query variant (`query_text`, perhaps?) would probably cover those cases well-enough.

If we wanted to keep the configurability of each input and output's format, what about adding that configuration to `Statement` instead? Or tracking the configuration of those formats in a `Statement`, generated through `prepare_typed` or some other `prepare_*` method?

NAlexPear:
@sfackler as a possible implementation of a "string mode" query method that lets users specify their own formats for inputs and outputs, I've added `param_formats` and `column_formats` to [`query_raw`](https://github.com/sfackler/rust-postgres/pull/747/files#diff-6d4154ba14a80131d255696a916b1ba537f6080c9edf11289e51f6cc17b9b66dR371).

I decided against a separate method since `query_raw` is supposed to be the "maximally flexible version of `query`", and `query_raw_er` didn't seem like a good road to go down. But if this change breaks your backwards-compatibility goals, then I can pull that work out into a separate `query` method.

Also not sure which check failed on this latest commit, since I don't have access to CircleCI, but let me know what it is and I'll fix it.

sfackler:
```
error[E0061]: this function takes 4 arguments but 2 arguments were supplied
   --> postgres/src/client.rs:276:35
    |
276 |             .block_on(self.client.query_raw(query, params))?;
    |                                   ^^^^^^^^^ -----  ------ supplied 2 arguments
    |                                   |
    |                                   expected 4 arguments

error[E0061]: this function takes 4 arguments but 2 arguments were supplied
   --> postgres/src/client.rs:276:35
    |
276 |             .block_on(self.client.query_raw(query, params))?;
    |                                   ^^^^^^^^^ -----  ------ supplied 2 arguments
    |                                   |
    |                                   expected 4 arguments

error[E0061]: this function takes 4 arguments but 2 arguments were supplied
   --> postgres/src/transaction.rs:114:58
    |
114 |             .block_on(self.transaction.as_ref().unwrap().query_raw(query, params))?;
    |                                                          ^^^^^^^^^ -----  ------ supplied 2 arguments
    |                                                          |
    |                                                          expected 4 arguments

error[E0061]: this function takes 4 arguments but 2 arguments were supplied
   --> postgres/src/transaction.rs:114:58
    |
114 |             .block_on(self.transaction.as_ref().unwrap().query_raw(query, params))?;
    |                                                          ^^^^^^^^^ -----  ------ supplied 2 arguments
    |                                                          |
    |                                                          expected 4 arguments

error: aborting due to 2 previous errors
```

sfackler:
In addition to the fact that it breaks backwards compatibility, I don't think that query_raw is the right thing to modify, since it still works with ToSql/FromSql which are defined to work with binary formats.

NAlexPear:
gah, sorry, forgot about the sync client. Updating!

NAlexPear:
> I don't think that query_raw is the right thing to modify, since it still works with ToSql/FromSql which are defined to work with binary formats.

If one wants to mix-and-match both input and output formats (which I would like to), I don't think this requirement is avoidable. Nor is it a bad thing, I don't think... we can leverage [`to_sql_checked` to make sure that only those `ToSql` implementations that can handle `TEXT` are valid for `Format::Text` params](https://github.com/sfackler/rust-postgres/pull/747/files#diff-75a8b28bb1837403d9e313e5fb5111a321fd4f223e9c150a0bc028313db7fb66R215), and probably ditto for `FromSql` (although I didn't implement that check). `ToSql` and `FromSql` still seem valid to me in this context, since the important part is the handling of `Type::TEXT`.

postgres-types/src/lib.rs
kevlarr: > An alternative I've been thinking about is to have a separate query method that operates entirely in "string mode", both for parameters and the resulting rows.

@sfackler This would be wonderful; is it anything you've thought more about? Passing arbitrary types as strings is particularly useful, since I could pass any number as a string without knowing the exact intended type and Postgres itself can figure it out. Being able to then load those numbers as strings would also be wonderful. My use case might not be typical, though, so I might be the only person wanting this.

NAlexPear:
Closing in favor of #836

PR #752, opened 2021-03-30
Support for physical and logical replication

petrosagg:
This PR adds the replication mode options to the `Config` and a `copy_both_simple` method to the `Client` to support queries that enter the CopyBoth mode. At the time of writing the only query that uses the CopyBoth mode is replication but more could be added in the future.

`copy_both_simple` returns a `CopyBothDuplex` type that implements both `Stream` and `Sink`. Users can split the two halves if they need to using [`futures::StreamExt::split`](https://docs.rs/futures/0.3.14/futures/stream/trait.StreamExt.html#method.split).

Similarly to binary copies, this PR also provides two helper types that can wrap the `CopyBothDuplex` type and offer a higher level view of the data to the user.

* `ReplicationStream`: Decodes the physical replication protocol according to the [specification](https://www.postgresql.org/docs/10/protocol-replication.html). It provides raw byte access to the contents of XLogData messages.
* `LogicalReplicationStream`: Internally wraps `ReplicationStream` and further decodes the contents of the XLogData messages according to the [logical replication specification](https://www.postgresql.org/docs/10/protocol-logicalrep-message-formats.html)

Finally, this PR includes an integration test that sets up a replication connection, starts a CopyBoth stream for logical replication and tests that the correct data is sent and received.


petrosagg:
@jeff-davis here is the new PR! Any comments are very much appreciated :)

postgres-protocol/src/message/backend.rs
```diff
@@ -33,6 +36,24 @@ pub const PARAMETER_DESCRIPTION_TAG: u8 = b't';
 pub const ROW_DESCRIPTION_TAG: u8 = b'T';
 pub const READY_FOR_QUERY_TAG: u8 = b'Z';
 
+// replication message tags
+pub const XLOG_DATA_TAG: u8 = b'w';
+pub const PRIMARY_KEEPALIVE_TAG: u8 = b'k';
+
+pub trait Parse: Sized {
```
sfackler, line 43: Why is a new trait being introduced just to parse replication messages?

postgres-protocol/src/replication/mod.rs
```diff
@@ -0,0 +1,33 @@
+//! This module defines the [DecodingPlugin] trait that can be used to implement the decoding
+//! logic of any logical decoding plugin.
+//!
+//! Two reference implementations are provided:
+//!
+//! * [PgOutput]: Decodes the built-in pgoutput stream
+//! * [Raw]: A dummy plugin that yield each message it receives as bytes
```
sfackler, line 7: This seems like a much higher level API than what's intended to be in the postgres-protocol crate.

tokio-postgres/Cargo.toml
```diff
@@ -45,7 +45,7 @@ futures = "0.3"
 log = "0.4"
 parking_lot = "0.11"
 percent-encoding = "2.0"
-pin-project-lite = "0.2"
+pin-project = "1.0"
```
sfackler, line 48: Why is this change being made?

tokio-postgres/src/client.rs
```diff
@@ -81,6 +84,21 @@ impl InnerClient {
         })
     }
 
+    // Send a message for the existing entry in the pipeline; don't
+    // create a new entry in the pipeline. This is needed for CopyBoth
+    // mode (i.e. streaming replication), where the client may send a
+    // new message that is part of the existing request.
+    pub fn unpipelined_send(&self, messages: RequestMessages) -> Result<(), Error> {
```
sfackler, line 91: Why would CopyBoth need a different API to do this than the existing CopyIn logic?

tokio-postgres/src/config.rs
```diff
@@ -534,6 +557,12 @@ impl fmt::Debug for Config {
             }
         }
 
+        let replication_mode_str = match self.replication_mode {
```
sfackler, line 560: Why not just derive debug for replication mode?

tokio-postgres/src/replication_client.rs
```diff
@@ -0,0 +1,934 @@
+//! Streaming replication support.
```
sfackler, line 1: Rather than having an entirely separate client for this, I would rather just have a `Client::copy_both` method that returns a type implementing `Sink` and `Stream` (like a combination of `CopyInSink` and `CopyOutStream`, and have these more complex interfaces be developed externally initially.

postgres-protocol/src/replication/mod.rs
petrosagg: We can move in different directions here. The first of the implementations, `pgoutput`, is part of the postgres protocol (https://www.postgresql.org/docs/10/protocol-logicalrep-message-formats.html) so I think you would agree that at least that one belongs to this crate.

Now, initially I had it set up such that this was the only mode of logical replication that you could use, but in fact there are many plugins out there, so I added this trait.

What do you think of removing this trait completely and just offering two concrete ways of getting a replication stream? One that gives you the decoded built-in messages of postgres, and one that gives you bytes. And anything that needs to parse the bytes can do so externally without messing with traits of this crate.

postgres-protocol/src/message/backend.rs
petrosagg: This was added to support external plugins using the `DecodingPlugin` trait but I'm happy to remove this. See my other comment on `src/replication/mod.rs` for a counter-proposal that gets rid of that

tokio-postgres/src/replication_client.rs
petrosagg: The initial motivation for the separate client was to prevent certain methods from being available to the user when in that mode and visa-versa. (I'm speaking on behalf of @jeff-davis here so I hope I'm representing it correctly).

I can certainly make this be a `Client::copy_both` on the client and rely on documentation for the users to know that if they create a client in replication mode and attempt to do normal queries they will get runtime errors.

petrosagg:
Thank you for the super quick review! I'm going to try and address your comments today :)

petrosagg:
@sfackler I have addressed all your comments, I think it looks much better now! Here is a summary of what I changed:

* `ReplicationClient` is gone, the normal client can be used to open a replication connection by passing the connection string option
* I added a `copy_both_simple` method on the client that returns a type that is both a Stream and a Sink
* `DecodingPlugin` and `Parse` traits are gone. Instead, two utility types are provided: `ReplicationStream` and `LogicalReplicationStream` that users can use to wrap the raw `CopyBothDuplex` stream returned by the client. This is the same pattern used for the binary copy stuff
* `pin_project_lite` is left intact

I've re-authored all the commits to be reviewable on their own and the PR is now ~1400 lines versus ~2250 before :)

tokio-postgres/src/config.rs
petrosagg: fixed

vitaly-burovoy (changes requested):
A good request!
I hoped @sfackler permits ReplicationClient because it is known at the connection stage there can be only replication flow and simple queries. Still think it is wise to have a special struct for it.

Several nitpickings are below:

tokio-postgres/src/replication_client.rs
petrosagg: fixed

postgres-protocol/src/message/backend.rs
```diff
@@ -776,6 +890,654 @@ impl RowDescriptionBody {
...
+    }
+}
+
+/// The data of an individual column as it appears in the replication stream
+#[derive(Debug)]
+pub enum TupleData {
+    /// Represents a NULL value
+    Null,
+    /// Represents an unchanged TOASTed value (the actual value is not sent).
+    Toast,
```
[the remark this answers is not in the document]
petrosagg: Good idea, changed

postgres-protocol/src/message/backend.rs
```diff
@@ -776,6 +890,654 @@ impl RowDescriptionBody {
...
+        Ok(tuple)
+    }
+}
+
+/// A BEGIN statement
+#[derive(Debug)]
+pub struct BeginBody {
+    final_lsn: u64,
+    timestamp: i64,
+    xid: i32,
```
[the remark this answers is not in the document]
petrosagg: thanks for checking this, fixed!

postgres-protocol/src/message/backend.rs
```diff
@@ -776,6 +890,654 @@ impl RowDescriptionBody {
...
+
+#[non_exhaustive]
+/// A message of the logical replication stream
+#[derive(Debug)]
+pub enum LogicalReplicationMessage {
+    /// A BEGIN statement
+    Begin(BeginBody),
+    /// A BEGIN statement
+    Commit(CommitBody),
+    /// An Origin replication message
```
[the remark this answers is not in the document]
petrosagg: added

tokio-postgres/tests/test/replication.rs
```diff
@@ -0,0 +1,132 @@
...
+    };
+
+    let commit = loop {
+        match stream.next().await {
+            Some(Ok(XLogData(body))) => {
+                if let Commit(commit) = body.into_data() {
+                    break commit;
+                }
+            }
+            Some(Ok(_)) => (),
```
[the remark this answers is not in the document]
petrosagg: There was indeed a bug in the implementation. I fixed the problem and provided an example in the testcase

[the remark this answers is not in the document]
vitaly-burovoy: Thank you for the quick reply!

My not deep digging shows the client does not send the `CopyData` message to the server (I watched it in the wireshark) even if I add `this.stream.flush().await` in the `standby_status_update`/`hot_standby_feedback` functions.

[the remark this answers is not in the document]
petrosagg: yikes, that doesn't sound good. I'll take a look

tokio-postgres/tests/test/replication.rs
```diff
@@ -0,0 +1,149 @@
...
+    assert_eq!(tuple_data.len(), 1);
+    assert!(matches!(tuple_data[0], TupleData::Text(_)));
+    if let TupleData::Text(data) = &tuple_data[0] {
+        assert_eq!(data, &b"42"[..]);
+    }
+
+    // Send a standby status update and require a keep alive response
+    let lsn: PgLsn = lsn.parse().unwrap();
+    let epoch = UNIX_EPOCH + Duration::from_secs(946_684_800);
+    let ts = epoch.elapsed().unwrap().as_micros() as i64;
```
ericseppanen, line 135: What is being computed here? What is the significance of `946_684_800`?

vitaly-burovoy: All stamps in PG are since internal Epoch which is set to 2000-01-01+00.
You can find this mentioned in the protocol description page: https://www.postgresql.org/docs/13/protocol-replication.html
`The client's system clock at the time of transmission, as microseconds since midnight on 2000-01-01.`


The constant `946_684_800` is a unixtime stamp of the PG's Epoch.

tokio-postgres/tests/test/replication.rs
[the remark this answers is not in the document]
petrosagg: The integration test in this PR correctly sends the standby update message, here is the decoded wireshark capture where you see `0x72` (character `'r'`), followed by the rest of the fields:

```
 72 0000000001aafa78 0000000001aafa78 0000000001aafa78 000262fdc867a96701
'r'        write_lsn        flush_lsn        apply_lsn          timestamp
```

[](https://imgur.com/033TZlq.png)

tokio-postgres/tests/test/replication.rs
petrosagg: added a comment to clarify this

petrosagg:
@sfackler I've reworded the description above to aid with reviewing. I hope this is more in line with what you had in mind :)

tokio-postgres/tests/test/replication.rs
ericseppanen: Thanks for the clarification.

Perhaps this code should be inside a public `pg_time_now()` function?

ericseppanen: Actually, something like this would be even better. 

```rust
struct PgTimestamp(u64);
impl From<SystemTime> for PgTimeStamp {
    // ...
}
```

That makes `standby_status_update` a little more foolproof.

ericseppanen: I wrote a first draft of `PgTimestamp` [here](https://github.com/petrosagg/rust-postgres/pull/1/commits/a0d067b66447951d1276a53fb09886539c3fa094).

tokio-postgres/tests/test/replication.rs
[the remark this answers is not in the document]
vitaly-burovoy: I'm sorry, it seems I did not fetch the fresh code. I'll test it soon.

postgres-protocol/src/message/backend.rs
```diff
@@ -340,7 +341,12 @@ impl ReplicationMessage<Bytes> {
             XLOG_DATA_TAG => {
                 let wal_start = buf.read_u64::<BigEndian>()?;
                 let wal_end = buf.read_u64::<BigEndian>()?;
-                let timestamp = buf.read_i64::<BigEndian>()?;
+                let ts = buf.read_i64::<BigEndian>()?;
+                let timestamp = if ts > 0 {
+                    *PG_EPOCH + Duration::from_micros(ts as u64)
+                } else {
+                    *PG_EPOCH - Duration::from_micros(-ts as u64)
+                };
```
ericseppanen, line 349: What should happen if `ts` is negative?

ericseppanen, line 349: Going backward from `PG_EPOCH`: is that really allowed?

Perhaps an attempt to represent time before `PG_EPOCH` should just be a hard error?

petrosagg: Yeah it's ok to send negative values. This is the pg debug log (on 13.2) when sending UNIX_EPOCH:

```
postgres_1  | 2021-04-21 19:02:49.770 UTC [91] DEBUG:  write 0/16A3490 flush 0/16A3490 apply 0/16A3490 (reply requested) reply_time 1970-01-01 00:00:00+00
```

tokio-postgres/tests/test/replication.rs
ericseppanen: The commit "use SystemTime for timestamps at API boundary" resolves this question.

tokio-postgres/src/replication.rs
```diff
@@ -0,0 +1,184 @@
...
+const STANDBY_STATUS_UPDATE_TAG: u8 = b'r';
+const HOT_STANDBY_FEEDBACK_TAG: u8 = b'h';
+
+pin_project! {
+    /// A type which deserializes the postgres replication protocol. This type can be used with
+    /// both physical and logical replication to get access to the byte content of each replication
+    /// message.
+    ///
+    /// The replication *must* be explicitly completed via the `finish` method.
+    pub struct ReplicationStream {
```
ericseppanen, line 24: I'm a little confused by the last line of the description.  What `finish` method?

jeff-davis:
With the new approach, how do you stop a replication stream safely from the client side so that you can execute new commands?

jeff-davis:
I think the smallest useful PR we can make here is:

1. The image updates for `pg_hba.conf` so replication can be tested.
2. Enable specifying the `replication` parameter when connecting.
3. Handle `CopyBoth`. I believe this includes the ability for the client to safely send a `CopyDone` message while the server is still streaming data. For that, I think we need something like the `unpipelined_send` that I introduced in 2d669347.
4. I'd like it if the interface could be made safe from protocol desyncs by mutably borrowing the client connection when returning a CopyBoth stream, like I did in #696, and safely shutting down the copyboth stream when the object is dropped (which I believe requires `PinnedDrop`).

I think the logical replication stuff should be in a separate PR. And a lot of the helper functions for building replication commands and parsing the responses can be in a separate PR as well (or perhaps a separate crate, as @sfackler indicated).

petrosagg:
@jeff-davis I think that makes sense. I'll work on splitting this PR in one that just adds `copy_both` support and another that has the replication decoding.

Regarding your concerns about terminating the stream from the client, I think you're right and that's an omission from my part. I'll take a look into that too. My initial plan for the implementation was to use the same mechanics that are used during a `copy_in` operation which doesn't use PinnedDrop either.

petrosagg:
closing in favour of #778. Once that is merged we can address the replication protocol itself in a separate PR

PR #754, opened 2021-04-01
feat: add method to clear types cache

### LINKED ISSUE #753: Possibility to clear types cache
At the moment the library works in such a way that if we use custom types (domain, enum, composite) in our queries and if we need to upgrade database, we need to reload the server after that.

After much research into the source code, I realized that everything is hidden in InnerClient.state.types. Once a connection exists, the client polls the database and collects the types in a hashmap. Cool, but it's not assumed that the types can actually be dynamic.

I think there are several ways to implement this:
1) Add a setting where InnerClient doesn't cache types at all. Easiest and for most lightweight systems, utilities, this is the way to go!
2) Add the possibility to clear the types cache in the public api.
3) Add the possibility to customize the lifetime of the cache. This is more complicated, it takes time. But it would be perfect! https://docs.rs/cached/0.23.0/cached/

It would be ideal to implement all of this! We couldn't cache simple things.  In other cases, state with cache and clear the cache on error.

pleshevskiy:
This PR solves my problem.

Closes #753

### DIFF
--- a/tokio-postgres/src/client.rs
+++ b/tokio-postgres/src/client.rs
@@ -113,6 +113,10 @@ impl InnerClient {
         self.state.lock().types.insert(oid, type_.clone());
     }
 
+    pub fn clear_types(&self) {
+        self.state.lock().types.clear();
+    }
+
     pub fn with_buf<F, R>(&self, f: F) -> R
     where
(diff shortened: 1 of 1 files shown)


sfackler:
Thanks!

Issue #729, opened 2021-01-11
Support passfile


jeff-davis:
I believe we at least want to support the [`passfile` connection parameter](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-PASSFILE) with the documented [format](https://www.postgresql.org/docs/current/libpq-pgpass.html). This avoids the need to store the password in the source and binary, or to come up with a custom way to read the password from another source.

We might also want to support the [`PGPASSFILE` environment variable](https://www.postgresql.org/docs/current/libpq-envars.html) and/or the default location (`~/.pgpass` on unix).

sfackler:
I think supporting the parameter makes sense. The environment variable question is basically #654.

psztoch:
.pgpass support will be useful. Please add it.

PR #758, opened 2021-04-27
Add support for libpq-compatible password files (`passfile=`).

mattheww:
Previously requested in #729.

Adds one new config field: `passfile`.

Accepts `pgpass=` in connection strings.

Support is only present when the tokio_postgres `runtime` feature is enabled. Adds a dependency on the tokio `fs` feature in that case.

Notes:

I've tried to match libpq's behaviour quite closely.

PostgreSQL docs are at:
https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-PASSFILE
https://www.postgresql.org/docs/current/libpq-pgpass.html

libpq has a default password file location (`~/.pgpass` or `%APPDATA%\postgresql\pgpass.conf`). I haven't imitated that; password-file lookup won't happen unless a passfile is explicitly requested.

On Unix, libpq rejects password files which are group- or world-readable. I haven't implemented that behaviour. I think it would be easy to add, but I'm not sure how to avoid it causing pain for the integration tests.

libpq treats connections to a unix socket in the default directory (whether specified explicitly or implicitly) as connections to 'localhost' for purposes of password-file lookup. I haven't tried to imitate that, because rust-postgres doesn't have a notion of the default unix socket directory.

### DIFF
--- a/tokio-postgres/src/passfile/mod.rs
+++ b/tokio-postgres/src/passfile/mod.rs
@@ -35,7 +35,7 @@ impl<'a> PassfileKey<'a> {
         // This default is applied by the server, rather than our caller, so we have to apply it here too.
         let dbname = dbname.unwrap_or(user);
         PassfileKey {
-            hostname: hostname,
+            hostname,
             port: port_string,
             dbname: dbname.as_bytes(),
             user: user.as_bytes(),
@@ -61,13 +61,9 @@ impl PassfileEntry {
                 if b == b':' {
                     return Ok(value);
                 } else if b == b'\\' {
-                    value.push(match it.next() {
-                        Some(b2) => b2,
-                        // To be consistent with libpq, if the line ends with
-                        // a backslash then the backslash is treated as part
-                        // of the last field's value.
-                        None => b'\\',
-                    })
+                    // To be consistent with libpq, if the line ends with a backslash then the backslash is treated as
+                    // part of the last field's value.
+                    value.push(it.next().unwrap_or(b'\\'))
                 } else {
                     value.push(b)
                 }
--- a/tokio-postgres/src/passfile/mod.rs
+++ b/tokio-postgres/src/passfile/mod.rs
@@ -1,0 +1,192 @@
+//! Support for reading password files.
+//!
+//! Requires the `runtime` Cargo feature.
+
+#[cfg(unix)]
+use std::os::unix::ffi::OsStrExt;
+use std::path::Path;
+
+use tokio::fs::{self, File};
+use tokio::io::BufReader;
+
+use crate::config::{Config, Host};
+
+#[cfg(test)]
+mod test;
+
+/// The data needed to search for a matching passfile entry.
+struct PassfileKey<'a> {
+    hostname: &'a [u8],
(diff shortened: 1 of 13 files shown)


tokio-postgres/src/passfile/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+///  - `config` doesn't have a user set.
+///
+/// If `config` doesn't have a dbname set, matches the dbname field against the user name.
+///
+/// Intended to match libpq's behavior closely.
+///
+/// Unlike libpq, doesn't attempt to convert a `host` set to the default socket path into `"localhost"` for matching
+/// purposes. So passfiles using `localhost` to supply passwords for unix socket connections won't work.
+///
+/// Unlike libpq, doesn't enforce any rules on the passfile's permissions.
```
sfackler, line 164: Why not? This can be accessed through MetadataExt: https://doc.rust-lang.org/std/os/unix/fs/trait.MetadataExt.html#tymethod.mode

tokio-postgres/src/passfile/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+) -> Option<Vec<u8>> {
+    // If we don't have a user, this connection can never work anyway
+    if let Some(user) = config.get_user() {
+        #[cfg(unix)]
+        {
+            let meta = match fs::metadata(&passfile).await {
+                Ok(meta) => meta,
+                Err(_) => return None,
+            };
+            if !meta.is_file() {
```
sfackler, line 179: Is this to check if the file is a symlink or something? Why would we only do that on Unix?

tokio-postgres/src/passfile/mod.rs
mattheww: It's easy to implement, but I'm not sure how to manage the integration tests: git can't directly represent a file with restricted permissions. I could have the tests create the files instead of shipping them in git, or have the tests change the permissions, or have the docker setup scripts install something suitable. What would you advise?

tokio-postgres/src/passfile/mod.rs
mattheww: I'm imitating libpq there (`passwordFromFile` in `fe-connect.c`). It doesn't reject symlinks. I'm not sure what it's for; maybe it's related to the following permissions check.

mattheww:
Sorry, I hadn't run clippy. I've added a commit addressing things it was complaining about.

tokio-postgres/src/passfile/mod.rs
sfackler: I think I'd probably use the tempfile crate to make a temporary directory for the test and then make the passfile with appropriate permissions.

Badreah87:
في الثلاثاء، ٤ مايو، ٢٠٢١ ٢:٤٧ ص Steven Fackler ***@***.***>
كتب:

tokio-postgres/src/passfile/mod.rs
mattheww: Thanks, I've tried it that way and submitted it as #766

Issue #768, opened 2021-05-11
Add more SSL options to Config


uce:
I was wondering whether you would be open to extending `Config` as follows:
* Add variants `VerifyCa` and `VerifyFull` to `SslMode`
* ~Add options `sslcert`, `sslkey`, `sslrootcert`~

For context: We allow users to specify a connection string that we then pass directly to the client and we would like it to behave in a similar way as `psql`. We currently have to provide the above options "out-of-band".

I would understand if this is rejected since it does not affect the client directly, but only how to configure TLS support (which is the responsibility of the client user and not the client in the current design).

If you don't have any concerns with this, feel free to assign it to me.

cc @sfackler

sfackler:
How would those options actually be consumed?

uce:
@sfackler Thanks for asking! The `sslcert`, `sslkey`, `sslrootcert` options will actually have to stay "out-of-band" for us, so I'll ignore those (and update the ticket description).

The goal would be to mimic `psql` behaviour by configuring the `SslConnector` accordingly. For instance, we currently disable cert verification, because that's how `psql` behaves:

```rust
let mut builder = SslConnector::builder(SslMethod::tls_client())?;
// Currently supported sslmodes (disable, prefer, require) don't verify peer certificates
builder.set_verify(SslVerifyMode::NONE);
Ok(MakeTlsConnector::new(builder.build()))
```

If we had the additional modes `verify-ca` and `verify-full`, we would use those to decide when to enable cert verification. Within `rust-postgres`, the new modes would behave like `require`.

What do you think?

sfackler:
But there is no interface to pass that configuration down to the TLS implementation currently.

nyurik:
I was trying to implement same ssl options as available in [psql](https://www.postgresql.org/docs/15/libpq-ssl.html), e.g. `verify-full`. Is there a path forward with this issue?  It seems the issues with #774 were never resolved.

Ideally rust-postgres would handle TLS internally - via rustls by default, or for backward compatibility -- via openssl (feature flags), but at the moment the structure of the lib seems to delegate all TLS configuration to the user, which creates multiple fairly complex and sometimes buggy implementations.

nyurik:
I did another, much simpler take on solving this in #988, hopefully this is more palatable to everyone, while it will unblock all those who want to use this functionality.

WGH-:
[...]

PR #774, opened 2021-05-18
Add sslmode `verify-ca` and `verify-full`

uce:
@sfackler Thanks for the discussion in https://github.com/sfackler/rust-postgres/issues/768. I thought it's best to simply follow up with a proposal for what the proposed change might look like. I'd appreciate your feedback.

As noted in the issue, the general idea here is to mimic Postgres client behavior. I can see that the added modes might also be confusing as it's up to the user to do something useful with them (e.g. use the convenience `from_tls_config` constructor). Do you think this has potential to land in `master`?

I'm happy to drop the last commit 39d5178, because it touches quite a bit of infrastructure for maybe too little added coverage.

cc @petrosagg


tokio-postgres/src/config.rs
```diff
@@ -85,7 +89,9 @@ pub enum Host {
 /// * `options` - Command line options used to configure the server.
 /// * `application_name` - Sets the `application_name` parameter on the server.
 /// * `sslmode` - Controls usage of TLS. If set to `disable`, TLS will not be used. If set to `prefer`, TLS will be used
-///     if available, but not used otherwise. If set to `require`, TLS will be forced to be used. Defaults to `prefer`.
+///     if available, but not used otherwise. If set to `require`, `verify-ca`, or `verify-full`, TLS will be forced to
+///     be used. Defaults to `prefer`. Note that for modes `verify-ca` and `verify-full`, it's up to the user to configure
+///     the SSL stream to respect the desired configuration (e.g. verification of certs, hostname verification).
```
sfackler, line 94: This note seems like a pretty big footgun.

uce: @sfackler Any opinion here? (Sorry to ping you again, please let me know if I should just wait.)

sfackler: I'm not sure how doable it would be to adjust `MakeTlsConnect` and/or `TlsConnect` implementations to apply extra configuration after they were initially constructed.

uce: What about introducing convenience constructors that create `MakeTlsConnect` and/or `TlsConnect` with correct configurations (e.g. like [here](https://github.com/sfackler/rust-postgres/pull/774/files#diff-de463e7f8bf25884492679e414d50bf79214f9c353388b1380d8969551661038R104))? The checks (item 3 in the comment above) would guard against wrong configurations and the constructor would be easy to use, e.g. `make_tls(config: &Config)`.

sfackler: But if a user does not realize they need to manually parse their config string up front and use that to create the connector things will silently not work as expected.

uce: > things will silently not work as expected

In item 3 [above](https://github.com/sfackler/rust-postgres/pull/774#discussion_r635199367), the idea was that it will actually fail (and not be silently ignored). Would that be an acceptable solution?

If yes, I could follow-up with an implementation. If no, I think we should close the PR and the linked issue.

nyurik: @sfackler looking into this more - it is fairly complicated to pre-parse sslmode because the connection string could come in many different forms (e.g. url-like or space separated), so every implementation would have to duplicate the whole config parsing code without any easy way to extend what you already have.

I propose `pg::config::Config::from_str(conn_str)` to support all ssl modes as defined by psql, but `PostgresConnectionManager` will continue to only support `disable/prefer/require`.  If other ssl_mode is given, `connect_tls` will return an error.

This way, the code using this library will have consistent behavior:
* if the connection string has `sslmode=verify-full`, this library will parse it, but `connect_tls` will give a TLS error as `kind=tls` or `kind=authentication` (?)
* if the user of the lib want to support it, they will have to handle it while constructing the tls connector, and will modify config with `config.clone_as_tls_required()` -- which will keep all of the original config vals but replace ssl mode with `required`.

The benefit of this approach is that all existing code will continue working as before, and if the user doesn't handle new ssl modes, they will continue getting an error - its just that that error will happen a bit later in the process.

nyurik:
See also an alternative implementation of this in #988 -- a far smaller implementation with just a few lines of code that should address the same need while maintaining backward compatibility

PR #778, opened 2021-05-25
Support CopyBoth queries and replication mode in config

petrosagg:
This PR only adds support for setting the connection replication mode in the connection configuration and support for `CopyBoth` mode queries. This is the minimum support needed for downstream users to implement replication support on their own.

@jeff-davis After studying the code I think we don't need to add `unpipelined_send` nor worry about protocol desyncs.

The reason we don't need `unpipelined_send` is because just like `CopyIn`, when performing a CopyBoth operation the connection enters a special mode where it consumes FrontendMessages from a sub-receiver . These sub-receivers are the `CopyInReceiver` and `CopyBothReceiver` structs.

The reason we don't need to worry about protocol desyncs is that once the connection enters a sub-mode (with `Copy{In,Both}Receiver`) all other sends to the main command queue are ignored while the subprotocol is running. Only after these are exhausted, which always happens with a `CopyDone`/`CopyFail`/`Sync` message, the system resumes consuming the main queue.

The PR includes two tests that verify the client can resume operation to normal query processing. One that gracefully shuts down the replication stream and one that simply drops the handles.


tokio-postgres/tests/test/copy_both.rs
```diff
@@ -0,0 +1,91 @@
...
+        Row(row) => row.get("consistent_point").unwrap(),
+        _ => panic!("unexpeced query message"),
+    };
+
+    client
+        .simple_query("INSERT INTO foo VALUES ('fde61ce315faac78b17c')")
+        .await
+        .unwrap();
+
+    let query = format!("START_REPLICATION SLOT slot LOGICAL {}", lsn);
```
jeff-davis, line 53: Also test for an error case for `START_REPLICATION`, like an LSN that is way in the future or something. And right after the error, you can try a `close()`/`drop()` to make sure it still shuts down properly.

Can errors happen in the middle of the stream? I suppose so -- for instance, an OOM in `test_decoding`. I don't see an easy way to exercise this path, though, but if you have an idea it might be good to include that and make sure it doesn't desync the protocol.

tokio-postgres/src/copy_both.rs
```diff
@@ -0,0 +1,280 @@
...
+}
+
+pin_project! {
+    /// A duplex stream for consuming streaming replication data.
+    ///
+    /// The stream side *must* be consumed even if not required in order to process the messages
+    /// coming from the server. If it is not, `Sink::close` may hang forever waiting for the stream
+    /// messages to be consumed.
+    ///
+    /// The copy should be explicitly completed via the `Sink::close` method to ensure all data is
```
jeff-davis, line 104: It seems like dropping it is enough?

tokio-postgres/src/copy_both.rs
```diff
@@ -0,0 +1,280 @@
...
+    /// Both directions are closed and we're waiting for the final CommandComplete message
+    Reading,
+}
+
+pin_project! {
+    /// A duplex stream for consuming streaming replication data.
+    ///
+    /// The stream side *must* be consumed even if not required in order to process the messages
+    /// coming from the server. If it is not, `Sink::close` may hang forever waiting for the stream
+    /// messages to be consumed.
```
jeff-davis, line 102: Can you expand on the reasoning here? `close()` or `drop()` should certainly ensure that a `CopyDone` is sent. And if the receiver is closed, that should cause future messages received in this sub-protocol to be discarded, eventually allowing the protocol to resume normal operations. Right?

tokio-postgres/tests/test/copy_both.rs
```diff
@@ -0,0 +1,91 @@
...
+    let (mut sink, stream) = duplex_stream.split();
+
+    // We can only do some basic sanity checking of the raw stream. We filter for messages starting
+    // with 'w', which are XLogData messages, and check that they contain the transaction we expect
+    let actual: Vec<_> = stream
+        .try_filter(|buf| future::ready(buf[0] == b'w'))
+        .take(3)
+        .try_collect()
+        .await
+        .unwrap();
```
jeff-davis, line 68: Is it worth having a test where we know there are additional unconsumed messages, and we close anyway?

tokio-postgres/src/copy_both.rs
```diff
@@ -0,0 +1,280 @@
...
+                }
+                // The server indicated an error, terminate both sides
+                Message::ErrorResponse(error) => {
+                    *this.state = match this.state {
+                        ReadingCopy | Reading => return unexpected,
+                        _ => ReadingCopy,
+                    };
+                    // Indicate to CopyBothReceiver to produce a Sync message instead of CopyDone
+                    let _ = this.error_sender.take().unwrap().send(());
+                    return Poll::Ready(Some(Err(Error::db(error))));
```
jeff-davis, line 153: Does this assume that the error happened during `START_REPLICATION`? What about an error that happened during the stream; shouldn't we still send a `CopyDone`?

jeff-davis (commented):
Looks great. A few questions/concerns.

tokio-postgres/tests/test/copy_both.rs
```diff
@@ -0,0 +1,91 @@
...
+        .copy_both_simple::<bytes::Bytes>(&query)
+        .await
+        .unwrap();
+
+    let (mut sink, stream) = duplex_stream.split();
+
+    // We can only do some basic sanity checking of the raw stream. We filter for messages starting
+    // with 'w', which are XLogData messages, and check that they contain the transaction we expect
+    let actual: Vec<_> = stream
+        .try_filter(|buf| future::ready(buf[0] == b'w'))
```
jeff-davis, line 64: Looks like this is causing a test failure.

tokio-postgres/src/copy_both.rs
```diff
@@ -0,0 +1,280 @@
...
+    type Item = FrontendMessage;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<FrontendMessage>> {
+        if self.done {
+            return Poll::Ready(None);
+        }
+
+        // If an error was received from the backend we have to send a Sync message and not send
+        // any other messages
+        if let Ok(Some(_)) = self.error_receiver.try_recv() {
```
jeff-davis, line 49: This only ever receives `()`, right? Should we get rid of the `_` for clarity?

tokio-postgres/src/client.rs
```diff
@@ -449,6 +451,15 @@ impl Client {
         copy_out::copy_out(self.inner(), statement).await
     }
 
+    /// Executes a CopyBoth query, returning a combined Stream+Sink type to read and write copy
+    /// data.
+    pub async fn copy_both_simple<T>(&self, query: &str) -> Result<CopyBothDuplex<T>, Error>
```
jeff-davis, line 456: After the replication stream, if the timeline is historical, Postgres will send a tuple as a response. So we actually need a function that returns something like `Result<(CopyBothDuplex<T>, Option<SimpleQueryMessage>), Error>` (or maybe `Result<(CopyBothDuplex<T>, Option<Vec<SimpleQueryMessage>>), Error>` in case other commands are added in the future which use `CopyBoth` and return a set).

It's actually very specific to `START_REPLICATION` (and even more specifically, to physical replication), so it might make sense to have a more specific name or at least clarify what it's expecting the command to do. Maybe something like `copy_both_simple_with_result()`?

jeff-davis:
`START_REPLICATION` and `BASE_BACKUP` don't follow a simple pattern. `START_REPLICATION` begins streaming in `CopyBoth` mode, and when the copy is finished, it then sometimes sends a single tuple (when streaming from a timeline other than the current one). `BASE_BACKUP` *first* sends *two* result sets (one tuple in the first set, multiple tuples in the second set), then it sends the copy data in `CopyOut` mode.

Supporting these commands requires some special methods on `Client` that won't really be useful for anything else, so I think we should name them specifically, e.g. `client.start_replication_command()` and `client.base_backup_command()`.

jeff-davis:
```
//! It is recommended that you use a PostgreSQL server patch version                                               
//! of at least: 14.0, 13.2, 12.6, 11.11, 10.16, 9.6.21, or                                                        
//! 9.5.25. Earlier patch levels have a bug that doesn't properly                                                  
//! handle pipelined requests after streaming has stopped.
```

In my PR, I had the above comment. Does that apply to this PR, as well?

jeff-davis:
It would be good to have a way to send Standby Status Updates and Hot Standby Feedback. Not all of this has to be in this PR, though. I'm just commenting on everything necessary to make a full-featured and independent crate for replication.

tokio-postgres/src/copy_both.rs
```diff
@@ -0,0 +1,280 @@
...
+            self.message_receiver.close();
+            self.done = true;
+            let mut buf = BytesMut::new();
+            frontend::sync(&mut buf);
+            return Poll::Ready(Some(FrontendMessage::Raw(buf.freeze())));
+        }
```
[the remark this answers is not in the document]
jeff-davis: Your code is fine as-is. From the protocol flow docs: "[in CopyBoth mode] The frontend should treat receipt of ErrorResponse as terminating the copy in both directions; no CopyDone should be sent in this case."

tokio-postgres/src/client.rs
petrosagg: That's a good point, I'll take a look on how we can expose this to users, ideally in a generic way.

tokio-postgres/src/copy_both.rs
```diff
@@ -0,0 +1,280 @@
...
+    let mut responses = client.send(RequestMessages::CopyBoth(receiver))?;
+
+    message_sender
+        .send(FrontendMessage::Raw(buf))
+        .await
+        .map_err(|_| Error::closed())?;
+
+    match responses.next().await? {
+        Message::CopyBothResponse(_) => {}
+        _ => return Err(Error::unexpected_message()),
```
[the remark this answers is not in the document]
petrosagg: This initial error is returned to the async call of `client.copy_both_simple()`. The error receiver hasn't been engaged yet

tokio-postgres/tests/test/copy_both.rs
petrosagg: It seems to be working now, can you test again?

petrosagg (commented):
@jeff-davis thank you for the thorough review!

Unfortunately last night I realised that the original design had a big hole that allowed de-syncing the protocol. I just push a new version that I think is Correct (tm) now.

When doing a CopyBoth query the architecture looks something like this:

```
                                        |
        <tokio_postgres owned>          |    <userland owned>
                                        |
pg -> Connection -> CopyBothReceiver ---+---> CopyBothDuplex
                                        |          /   \
                                        |         v     v
                                        |      Sink    Stream
```

The original version of the feature handled the state machine of the CopyBoth sub-protocol as part of the Stream implementation of `CopyBothDuplex` and treated `CopyBothReceiver` as a dumb relay of messages into the connection. Therein lies the problem. A user could create a `CopyBothDuplex` and drop it immediately. In that case the dumb `CopyBothReceiver` would unconditionally send a `CopyDone` to the server and finish. But what if the server sent an error in the meantime? There was nothing to handle this fact and no Sync message was being sent.


So this re-work of the feature flips this relationship around. `CopyBothReceiver` contains all the logic to drive the sub-protocol forward and ensures that no matter what the correct messages are being exchanged with the server. The user is free to drop their end at any point.

I've also added a ton of comments and a few diagrams to make reading and reviewing the code easier. If it's not a lot of ask I'd hugely appreciate another round of review, I think we're close!

Next week I plan to compile a modified version of postgres that errors out on purpose to test out the error paths of this PR and also see how we can incorporate getting results back for things like timeline changes etc.

jeff-davis:
Regarding the `future::ready()` error, I meant that it seems to be causing a failure in CI: https://github.com/sfackler/rust-postgres/pull/778/checks?check_run_id=2689467390

tokio-postgres/src/client.rs
jeff-davis: In case you missed my other comment, there's a similar issue for `BASE_BACKUP`, except with `CopyOut` instead. That can be a separate PR, though.

tokio-postgres/tests/test/copy_both.rs
jeff-davis: Failure is in CI: https://github.com/sfackler/rust-postgres/pull/778/checks?check_run_id=2689467390

petrosagg:
@jeff-davis 

 >`START_REPLICATION` and `BASE_BACKUP` don't follow a simple pattern.

It looks like the general pattern is that the database can respond with result sets, copy outs, or copy boths. Based on that I think we need here is to define something like:

```rust
enum ResponsePart {
    RowStream(RowStream),
    SimpleQueryStream(SimpleQueryStream),
    CopyOutStream(CopyOutStream),
    CopyBothStream(CopyBothStream),
}
```

And then provide a method to send a query and get a Stream of `ResponsePart`s back. This will leave the interpretations of the various segments to the user issuing the query or some other higher level downstream library. How does this sound?

> In my PR, I had the above comment. Does that apply to this PR, as well?

hm do you have a link to the patch? I thought pipelined queries were not allowed over a replication connection

tokio-postgres/tests/test/copy_both.rs
petrosagg: Added

jeff-davis:
> And then provide a method to send a query and get a Stream of `ResponsePart`s back. This will leave the interpretations of the various segments to the user issuing the query or some other higher level downstream library. How does this sound?

I like it. Maybe it can even be refactored so that there's one generic entry point, `query_simple_extended()`, that returns a stream of `ResponsePart`s; and the other entry points are just special cases of that one. But the refactoring might be better as a separate PR.

> hm do you have a link to the patch? I thought pipelined queries were not allowed over a replication connection

https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=a58db3aa10e62e4228aa409ba006014fa07a8ca2

Is there a reason you thought pipelined queries would not be allowed in replication mode? And if so, how would you prevent them, since the whole protocol implementation is built around pipelining, without explicitly blocking in certain cases?

petrosagg:
> Is there a reason you thought pipelined queries would not be allowed in replication mode? And if so, how would you prevent them, since the whole protocol implementation is built around pipelining, without explicitly blocking in certain cases?

You're 100% right. I confused pipelined queries with extended query mode which is the one that's not allowed in replication mode. So yeah, your original comment still applies in that users should pay attention to this bug. I added this note in the documentation of the `ReplicationMode` enum and the `Config::replication_mode` function

tokio-postgres/src/copy_both.rs
```diff
@@ -0,0 +1,382 @@
...
+///       `---,     |     ,---'
+///            \    |    /
+///             v   v   v
+///              CopyNone
+///                 |
+///                 v
+///            CopyComplete
+///                 |
+///                 v
+///           CommandComplete
```
jeff-davis, line 37: I don't think this is the right place in the state machine for `ServerError`. If a server error happens in `CopyBoth` mode, that's the odd (and hard-to-test) case where an error is thrown by the server in the middle of streaming. From the docs, it looks like Postgres will just return the `ErrorResponse` and then `ReadyForQuery`, without either of the `CommandComplete` messages.

You can test this by hacking up `test_decoding` to throw an error randomly every 10 records or something.

jeff-davis:
Summary of the remaining issues as I see them:

1. Need an API to execute a query using the simple protocol and get back a stream of `ResponsePart`s (or another API with similar functionality).
1. Change the way errors during a `CopyBoth` stream are handled.
    1. The simplest answer is to close the whole client if this happens. That might be best, because handling the errors will be hard to test and not especially useful. Note: this reasoning does not apply to ordinary errors that happen as a part of the `START_REPLICATION` command, which should be properly handled; only errors that happen after entering `CopyBoth` mode can reasonably cause the client to close the connection.
    1. If we can reasonably test it, handling the errors would be the most proper thing to do, and the client can potentially try again with a new `START_REPLICATION` command without closing the entire connection. Honestly, I don't see a real use case here, but it does seem a bit more proper.
1. It would be nice to offer an API to send Standby Status Updates and Hot Standby Feedback.

After these are done, I'll be able to port my other replication code to a new crate, and if that works, then I think this PR is ready.

petrosagg:
@jeff-davis totally agree with your list of pending things. I think the ResponsePart API will have to be a separate PR though, what do you think?

In the meantime, I have updated this PR and the CI is now green! I also added a test that produces an error before entering CopyBoth mode, one that produces an error after entering CopyBoth, and a regular test that streams the log.

I hit a strange thing with the in-stream error though. According to the documentation the server will ignore frontend messages until a Sync message is sent, but that's not the behaviour I see. The latest version of this PR has the sync sending commented out. If I uncomment it the test fails. Also, I've been looking for where this logic is implemented in postgres' code and I cannot find anything of that sort. The only thing I found was a `ignore_till_sync` flag that is not engaged during replication. I'm not sure if this is a bug of the docs, or if there are other paths that will cause the backend to wait for Sync. If you have any ideas here it would help a lot.

petrosagg:
I think I got to the bottom of this and it looks like the CopyBoth documentation of postgres is wrong. The behavior where the database will wait for a Sync message is also documented in more detail for the `CopyIn` mode:

> In the event of a backend-detected error during copy-in mode (including receipt of a CopyFail message), the backend will issue an ErrorResponse message. If the COPY command was issued via an extended-query message, the backend will now discard frontend messages until a Sync message is received, then it will issue ReadyForQuery and return to normal processing. If the COPY command was issued in a simple Query message, the rest of that message is discarded and ReadyForQuery is issued. In either case, any subsequent CopyData, CopyDone, or CopyFail messages issued by the frontend will simply be dropped.
https://www.postgresql.org/docs/devel/protocol-flow.html#PROTOCOL-COPY

From reading the code and running some experiments the same holds true for CopyBoth. However, since it must be started from a replication connetion which does not support extended queries there is no way for the backend to enter the "wait for Sync" mode.

I have verified this and included a test where the client sends a `CopyDone` while the ErrorResponse is in flight and the connection recovers correctly. I believe I have exercised all the error paths now in the tests :)

tokio-postgres/tests/test/copy_both.rs
petrosagg: All error paths are now exercised :)

jeff-davis:
> I think I got to the bottom of this and it looks like the CopyBoth documentation of postgres is wrong. The behavior where the database will wait for a Sync message is also documented in more detail for the `CopyIn` mode:

Great, thank you for investigating. I posted about this here: https://www.postgresql.org/message-id/c71317d9a1f486de00943cebc3ad1b27cf28d075.camel%40j-davis.com

jeff-davis:
I think this is about ready. It's fine with me to merge this, but I'll need the `ResponsePart`s in place to port the earlier work that I did.

tokio-postgres/tests/test/copy_both.rs
```diff
@@ -0,0 +1,125 @@
...
+        .await
+        .unwrap();
+
+    let mut msgs: Vec<_> = duplex_stream.collect().await;
+    let result = msgs.pop().unwrap();
+    assert_eq!(msgs.len(), 0);
+    assert!(result.unwrap_err().as_db_error().is_some());
+
+    // Ensure we can continue issuing queries
+    assert_eq!(q(&client, "SELECT 1").await[0].get(0), Some("1"));
```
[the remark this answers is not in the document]
petrosagg: I switched this statement to `DROP_REPLICATION_SLOT` and connected using `database=true` to exercise that path too. Hopefully also less confusing :)

tokio-postgres/tests/test/copy_both.rs
```diff
@@ -0,0 +1,125 @@
...
+
+#[tokio::test]
+async fn copy_both_stream_error() {
+    let client = crate::connect("user=postgres replication=database").await;
+
+    q(&client, "CREATE_REPLICATION_SLOT err2 TEMPORARY PHYSICAL").await;
+
+    // This will immediately error out after entering CopyBoth mode
+    let duplex_stream = client
+        .copy_both_simple::<bytes::Bytes>("START_REPLICATION SLOT err2 PHYSICAL FFFF/FFFF")
```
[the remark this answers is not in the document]
petrosagg: I started down the same path but then started looking at all the `ereport()` calls in src/backend/replication to find a condition that can be triggered with a vanilla db and get it into the tests

Thanks for getting this into the mailing list!

tokio-postgres/src/copy_both.rs
```diff
@@ -0,0 +1,345 @@
...
+                        _ => self.unexpected_message(),
+                    };
+                }
+                // The server indicated an error, terminate our side if we haven't already
+                Some(Err(err)) => {
+                    match self.state {
+                        CopyBoth | CopyOut | CopyIn => {
+                            self.sink_receiver.close();
+                            self.buffered_message = Some(Err(err));
+                            self.state = CommandComplete;
```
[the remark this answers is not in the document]
petrosagg: No because we're still waiting for the `ReadyForQuery` message and we also need to loop around and start the sending of the buffered error to the user.

tokio-postgres/src/copy_both.rs
```diff
@@ -0,0 +1,345 @@
...
+                        }
+                        _ => self.unexpected_message(),
+                    };
+                }
+                Some(Ok(Message::ReadyForQuery(_))) | None => {
```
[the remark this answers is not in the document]
petrosagg: Good catch, fixed

petrosagg (commented):
Thanks for all the reviews! The only annoying thing with this current PR is that it doesn't give you access to the resultset returned. I entertained the idea that it `CopyBothDuplex` could also implement Future which would result into the rows of the result set. I'll keep thinking about this a bit more because even if we add support for the `ResponsePart` thing in the future it'd be nice if the `copy_both_simple` method can be adapted to it

jeff-davis:
@sfackler do you have an opinion about the current state of this PR?

The goal for this PR is to implement enough at the protocol level for `CopyBoth` so that we can have a separate crate to:
* provide all the helper code in #696
* implement the logical replication protocol
* support the copy protocol submessages (primary keepalive, HS feedback, standby status update)

The only thing missing in this PR is that the `START_REPLICATION` command sometimes returns a result set in addition to the `CopyBoth` stream, and right now there's not a way to get the result set. A similar issue exists for the `BASE_BACKUP` command, which returns multiple result sets as well as a `CopyOut` stream. I think it's fine to do this in a separate PR because is not directly related to `CopyBoth`, it's more of an issue with our API for the simple query protocol (which is not quite so simple, at least as its used for replication).

tokio-postgres/src/copy_both.rs
```diff
@@ -0,0 +1,347 @@
...
+    fn poll_backend(&mut self, cx: &mut Context<'_>) -> Poll<()> {
+        use CopyBothState::*;
+
+        loop {
+            // Deliver the buffered message (if any) to the user to ensure we can potentially
+            // buffer a new one in response to a server message
+            if self.buffered_message.is_some() {
+                // If the receiver is gone we'll just drop the message
+                let _ = ready!(self.stream_sender.poll_ready(cx));
+                let message = self.buffered_message.take().unwrap();
```
jeff-davis, line 109: Perhaps use an `if ... let` here to avoid the `unwrap()`?

jeff-davis:
Maybe we should mark the `Client` with the `ReplicationMode` and provide a getter so that we know whether we can execute replication commands on a given client?

tokio-postgres/src/lib.rs
```diff
@@ -156,6 +156,7 @@ mod connect_raw;
 mod connect_socket;
 mod connect_tls;
 mod connection;
+mod copy_both;
```
jeff-davis, line 159: You also need a `pub use copy_both::CopyBothDuplex;`

tokio-postgres/src/copy_both.rs
petrosagg: Hm, but if `buffered_message` is None then it is definitely a bug in the library because the outer if checked that it's Some(_). I'll replace `unwrap()` with `expect()` to provide an explanation in case this happens in the future

tokio-postgres/src/lib.rs
petrosagg: fixed

tokio-postgres/src/copy_both.rs
```diff
@@ -0,0 +1,350 @@
...
+            // buffer a new one in response to a server message
+            if self.buffered_message.is_some() {
+                // If the receiver is gone we'll just drop the message
+                let _ = ready!(self.stream_sender.poll_ready(cx));
+                let message = self
+                    .buffered_message
+                    .take()
+                    .expect("already checked in outer if");
+                let _ = self.stream_sender.start_send(message);
+            }
```
jeff-davis, line 114: What I was getting at before was just a style issue, but I believe the following is semantically equivalent and more concise:

```
            if let Some(message) = self.buffered_message.take() {
                // If the receiver is gone we'll just drop the message                                                    
                let _ = ready!(self.stream_sender.poll_ready(cx));
                let _ = self.stream_sender.start_send(message);
            }
```

But I could have made a mistake, so please check that it really does mean the same thing.

petrosagg: I see. This won't work because `ready!()` desugars to:

```rust
            if let Some(message) = self.buffered_message.take() {
                // If the receiver is gone we'll just drop the message                                                    
                let _ = match self.stream_sender.poll_ready(cx) {
                    Poll::Ready(val) => val,
                    Poll::Pending => return Poll::Pending, // XXX message lost in this case
                };
                let _ = self.stream_sender.start_send(message);
            }
```
Which means that if the user is consuming their stream slowly and we get `Pending` back, the value will be dropped because `take()` has already been called.

petrosagg: What could work, but has wonky semantics imo is this:

```rust
            if self.stream_sender.poll_ready(cx).is_ready() {
                if let Some(message) = self.buffered_message.take() {
                    let _ = self.stream_sender.start_send(message);
                }
            }
```

This feels strange to me because technically we're waiting on a stream handle to become ready but then we might not use it

petrosagg: I pushed a version that I think ticks all the boxes :)

Martichou:
Any news on this one? What's still needed ? 
This would be very cool if the whole replication mode can be merged ! As far as I know, this is the only lib in Rust that would support it ! 💯

petrosagg:
hey @sfackler! Is there a chance you could take a look at this PR? It's the minimal support needed in `rust-postgres` to allow higher level crates to be developed independently

videni:
what is the latest status?

petrosagg:
@videni  this PR is ready, passes all tests and we have been using it in production for a long time at @materializeinc, unfortunately from a fork. It's the minimal amount of support needed in rust-postgres to allow for higher level external crates handling replication decoding etc.

@sfackler it would be amazing if we could get this merged and allow third party crates to do the rest of the work. It only implements CopyBoth support in the way you had suggested in an earlier review of yours and doesn't deal with the replication protocol itself at all

yeputons:
I'm wondering if it's possible to consume `Client` in `copy_both_simple` and add an explicit method like `CopyBothDuplex::close()` which returns `Client` back. That way Rust (not the user) guarantees that it's impossible to interleave CopyBoth mode with usual queries. This is much more restrictive than non-mutable reference, but I think it's for a good reason.

Moreover, it will be possible to actually poll all messages from the server until the `CopyDone`/`CommandComplete`/etc message is received in the `close()` method before returning to the usual query mode. I'm not sure, but is seems that it's currently possible to drop `CopyBothDuplex` and then enter a race between next `client.simple_query()` and `CopyBothReceiver`.

Looks like a small change in this PR. Any downsides?

petrosagg:
@yeputons this design was implemented in a previous version of this work by @jeff-davis. Note that even with this PR it is still impossible to interleave CopyBoth mode with normal queries. If you attempt to send a query while you're holding onto a CopyDuplex client, these queries are going to be queued internally and not interfere with the COPY BOTH operation at the protocol level. So at least that hazard is taken care of.

You could maybe argue that by not statically enforcing it a user could end up in a deadlock situation if they send a query and await it while the copy both client is not yet dropped. However, given how much work has been put already in this PR to make it as small a change as possible to get it merged I'm inclined it to leave it as is.

petrosagg:
@sfackler thanks for updating the CI image to pick up configuration automatically! I rebased this PR on top of master, it would be amazing if you could find some time to review it

Martichou:
@sfackler any news on this ? I'm relying on this PR for a project and I think a lot of people would benefits from this.
It would be great (and kind) to review the work that @petrosagg and @jeff-davis did on rust-postgres.

benesch:
@Martichou, you're welcome to use our fork (https://github.com/materializeInc/rust-postgres) for the moment! It's got this PR and #774 together, and we integrate new changes from rust-postgres periodically.

(We would, of course, love to get these PRs upstreamed, but @sfackler's time has seemed quite limited lately.)

Martichou:
@benesch Thanks ! I was doing the same (maintaining a fork), but I'll use yours from now on ;)

manfredcml:
@sfackler It'd be great if you can review this PR by @petrosagg and @jeff-davis. Merging this will be helpful for those who are currently maintaining or using the forks.

imor:
@sfackler would you be interested in replication support? I know this PR has conflicts, but I can clean this one up. Or, if you are not happy with the current design, maybe open a fresh one? We are forced to use fork from MaterializeInc (cc @benesch) in [pg_replicate](https://github.com/supabase/pg_replicate) which is not ideal because it prevents us from publishing to crates.io.

benesch:
Hey @imor! We were excited to see you build `pg_replicate` on top of our fork. We'd be delighted to see this merge upstream so that we could shut down our fork. Unfortunately, we've not been able to get Steven's attention on this PR in the last three years.

petrosagg:
I rebased this PR on top of current `master`

imor:
> Unfortunately, we've not been able to get Steven's attention on this PR in the last three years.

@benesch  That's sad. Were there any concrete reasons given as to why this PR is not accepted? I don't see any comment from Steven in this PR. In the worst case if the status quo doesn't change, would it be too much to ask to publish your fork on crates.io. It will allow downstream users to publish their crates as well. Not ideal, I know, but looking for options here.

benesch:
> That's sad. Were there any concrete reasons given as to why this PR is not accepted? I don't see any comment from Steven in this PR.

Nope, we’ve never heard anything, I’m afraid. I think Steven maintains this project on a volunteer basis though so I certainly don’t blame him. :)

Re a fork: I’d like to avoid having us (@MaterializeInc) on the critical path for new releases of `pg_replicate`. E.g., if there are bug fixes you need merged. We’ve accumulated a few Materialize specific patches in our fork too that might not be broadly interesting. You’re welcome to just publish our fork to crates.io under a crate name you control, though!

petrosagg:
> You’re welcome to just publish our fork to crates.io under a crate name you control, though!

Keep in mind that our fork contains more or less the code in this PR but also a lot more (the entire logical replication decoding logic). The idea was that once this PR was merged then the logical replication protocol would be handled by a separate crate layered on top of tokio-postgres, which would then expose the right functionality (i.e CopyBoth queries). I came up with this layering after we had our fork working to reduce the size of this PR and make more likely to be reviewed and merged. So what it might make sense to publish is a crate of this PR and then we can extract the logical decoding bits into another one. In the event that this PR eventually gets merged we can deprecate the published fork crate but keep the logical replication crate

imor:
> I think Steven maintains this project on a volunteer basis though so I certainly don’t blame him. :)

Oh there's no question of blaming Steven. All I have is gratitude for this great library. I understand how hard can it be to maintain a popular open source project.

Regarding the way forward, I like @petrosagg's suggestion about publishing a separate crate. A very practical solution which avoid fragmentation by a fork. I'm willing to help with this effort so let me know how I can be of use.

benesch:
I sent @sfackler an email asking if he has bandwidth to review/merge this PR. If we don't hear back in a few days, then we can investigate publishing a crate which is just rust-postgres plus this one PR.

petrosagg:
Heads up that I tidied up our fork and brought in all changes from the master branch of this repo plus this PR. The logical replication functionality has been extracted in a separate crate here https://github.com/MaterializeInc/rust-postgres/tree/master/postgres-replication and is the crate that we could in theory publish if this PR ever gets merged or gets published under a different name.

qianyiwen2019:
This has saved me a lot of time, thanks to [petrosagg](https://github.com/petrosagg). 
waiting for the merge.

LucaCappelletti94:
@petrosagg have you already pinged also @paolobarbolini, who seems to be the current org maintainer? This PR is under many many others, so it is possible if not likely that it is out of the radar entirely.

That being said, I am really looking forward to see it merged, and I seem to identify at least a few different seams to split this PR into 3-4 smaller PRs which may be faster to digest. If @paolobarbolini agrees and @petrosagg are okay with it, I will try to proceed in such a manner.

petrosagg:
@LucaCappelletti94 thanks for the ping, I was not aware of the maintainer change. I'd love to see this PR merged as well! I just rebased this PR on top of master and I'll be happy to help with any changes needed if @paolobarbolini would be interested in getting this merged.

PR #780, opened 2021-05-29
expose SimpleQueryRow's column names

### LINKED ISSUE #650: Get all statemet columns
I try to get all statement's columns for logging, but field `columns` is private. I use `simple_query` method because working with `pgbouncer`.

```rust
use postgres::{Client, SimpleQueryMessage, No};
use native_tls::{Certificate, TlsConnector};
use postgres_native_tls::MakeTlsConnector;
use std::fs;

fn main() {
    let mut client = Client::connect("...", NoTls).unwrap();

    let messages = client.simple_query("SELECT foo, bar FROM baz;").unwrap();
    match messages.first() {
        Some(SimpleQueryMessage::Row(msg)) => {
            // field "columns" of struct "tokio_postgres::row::SimpleQueryRow" is private
            println!("{}", msg.columns.join(" ")) // TODO print all columns
        },
        _ => (),
    }
}
```

mpajkowski:
resolves #650.

Added `SimpleColumn` type in order to ensure further extensibility.

### DIFF
--- a/tokio-postgres/src/row.rs
+++ b/tokio-postgres/src/row.rs
@@ -1,6 +1,7 @@
 //! Rows.
 
 use crate::row::sealed::{AsName, Sealed};
+use crate::simple_query::SimpleColumn;
 use crate::statement::Column;
 use crate::types::{FromSql, Type, WrongType};
 use crate::{Error, Statement};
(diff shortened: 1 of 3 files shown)


sfackler:
Thanks!

PR #791, opened 2021-06-17
Add query_first

JojiiOfficial:
Adds a new function `query_first` to `Client` which returns `Some(Row)` if n > 0 and `None` if n = 0 where n is the amount of returned rows by Postgres. It completely ignores all remaining rows.

### DIFF
--- a/tokio-postgres/src/client.rs
+++ b/tokio-postgres/src/client.rs
@@ -302,14 +302,12 @@ impl Client {
         T: ?Sized + ToStatement,
     {
         let stream = self.query_raw(statement, slice_iter(params)).await?;
+
         pin_mut!(stream);
 
-        let row = match stream.try_next().await? {
-            Some(row) => row,
-            None => return Ok(None),
-        };
+        let row = stream.try_next().await?;
 
-        Ok(Some(row))
+        Ok(row)
     }
 
     /// Executes a statement which returns zero or one rows, returning it.
--- a/tokio-postgres/src/client.rs
+++ b/tokio-postgres/src/client.rs
@@ -280,7 +280,39 @@ impl Client {
         Ok(row)
     }
 
-    /// Executes a statements which returns zero or one rows, returning it.
+    /// Executes a statement, returns the first result and completely ignores the remaining rows.
+    /// Returns `None` if no row was returned.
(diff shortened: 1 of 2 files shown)


sfackler:
What is the use case for this? It seems like a less efficient version of tacking a `LIMIT 1` onto the end of the query. The database is forced to send every resulting row of the query to the client, even if you are not actually interested in them.

JojiiOfficial:
`query_opt` and `query_one` do a [check](https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/src/client.rs#L276) for whether a next element is available in order to throw an error in case there are multiple lines returned. This function ignores further rows and can indeed be used with `LIMIT 1`.

sfackler:
I am aware that they perform that check. Why would that mean they couldn't be used with `LIMIT 1`?

JojiiOfficial:
I didn't say they can't? You simply don't need that check in some cases if you don't care about the error. This is what `query_first` aims to fix.

sfackler:
If you don't care about the error because it will never occur, then what is the problem in the first place with using `query_one` or `query_opt`? When would a user decide that they specifically want `query_first` and not `query_opt`?

JojiiOfficial:
If the user wants additional "performance" or better said, if the user doesn't need the `query_opt` or `query_one` to do a second check since he doesn't care about other rows, he uses `query_first`. I don't see the necessity in doing this extra check if I don't care about a potential error (ie. by using a `LIMIT 1`) so I don't want it to do a second check.

sfackler:
What percentage of runtime of query_opt have you measured being spent in that check?

Yukaru-san:
As far as I understand it, Jojii's addition improves the time itself by about  1 ms / 1000 requests.
This isn't much by itself but for some uses, those micro-optimizations are indeed quite useful.
Also, since you wouldn't need to use the ```LIMIT 1``` you don't have to load everything into a Vec first if you want to prevent an error.

That being said, he added a single method some people might find useful and shouldn't be hindering any other functionalities of the lib. So why not just add it?

sfackler:
Because library APIs are not generally trying to become the union of all features anyone has ever found useful.

Issue #799, opened 2021-07-01
postgres_types fails to compile for the ToSql macro, FromSql macro, or postgress attribute (repro repo inclunded)


jrsmith17:
I am currently unable to get Rust to compiler when trying to derive `ToSql` or `FromSql` due to `cannot find derive macro ToSql in this scope` errors. This also applies to the attribute `postgres` with the same scoping error. I have made a reproduction repo here: [https://github.com/jrsmith17/postgres-issue-repro]. I will also post the germane bits below.

Also, as best I can tell, I followed the example from the docs here [https://docs.rs/postgres-types/0.2.1/postgres_types/index.html] and I don't **think** I misunderstood the docs, but feel free to poke holes in that theory.

src/main.rs
```use postgres::*;
use postgres_types::{ToSql, FromSql};

#[derive(Debug, ToSql, FromSql)]
#[postgres(name = "Placeholder")]
pub struct Placeholder {
    pub app_id: String,
    pub url: String
}

fn main() {
    println!("Hello, world!");
}
```

Cargo.toml
```[package]
name = "postgres-issue-repro"
version = "0.1.0"
edition = "2018"

[dependencies]
postgres = "0.19.1"
postgres-types = "0.2.0"
```

I've tried this with versions 0.2.1, 0.2.0, and 0.1.3 of the postgres crate. I am running on Windows 10 with the latest updates (as of this posting). My `rustc --version` returns `rustc 1.53.0 (53cb7b09b 2021-06-17)`. My toolchain is `stable-x86_64-pc-windows-msvc (default)` (though this also fails in nightly for me). I also tried specifying `postgres_types::ToSql` in the derive statement and it had a very similar, if slightly more specific, error.

jrsmith17:
Figured this out. I had misread the docs about enabling the derive feature. This seems easy for others to do. An ideal-and-outside-the-scope-of-this-crate solution would be to provide an error message instructing you to add it to Cargo.toml. A seemingly good-enough bandaid solution would be tweaking the docs to add a Cargo.toml snippet to make it more clear what the intent is. I will throw together a PR in the next few hours after wrapping some day job things.

jrsmith17:
I made the very small PR I mentioned above. Here is the link: [https://github.com/sfackler/rust-postgres/pull/800]

PR #800, opened 2021-07-01
Making Postgres Types Documentation More Explict For Feature Flags

jrsmith17:
I had misread the docs around needing to enable a feature flag in order to access the derive macros for postgress types. If I misread it, I figured that others could as well, so I threw together a quick PR to make it more clear.

This is tied to the issue I posted here: [https://github.com/sfackler/rust-postgres/issues/799]


sfackler:
Thanks!

PR #806, opened 2021-07-13
Add `FromSql` and `ToSql` impls for arrays (guarded behind feature)

### LINKED ISSUE #805: Implement `ToSql` and `FromSql` for fixed size Rust arrays `[T; N]`
Thanks for your work on this library!

I was wondering if there is a specific reason why `ToSql` and `FromSql` is not implemented for fixed size arrays? For example:

```sql
create table videos {
    resolution integer[2]
}
```

With this fixed size array, I would expect to be able to retrieve it like `row.get::<_, [i32; 2]>(0)`. Same for having composite types and using `derive(ToSql)` on a `struct`. 

Right now, postgres arrays can only be converted from/to slices and `Vec<T>`. Maybe support for fixed size Rust arrays was not added because until recently you couldn't implement a trait for arrays of any length? With const generics, this has changed.

I tried to find a related issue here, but couldn't. If there is no reason for the lack of this impl, I can create a PR.

LukasKalbertodt:
Fixes #805 

First I tried using the crate `rustversion` to automatically enable the two impls on newer compilers. However, as far as I understand, it is not possible to conditionally include dependencies based on the compiler version. But I wanted to use [`array-init`](https://crates.io/crates/array-init) as there is no other good way to construct the array element by element. So now I decided to just use a Cargo feature. Some might prefer that anyway because downstream users can declare their requirements more explicitly.

I tested this patch in a project of mine and it seems to work fine. 

Let me know what you think!

### DIFF
--- a/postgres-types/src/lib.rs
+++ b/postgres-types/src/lib.rs
@@ -428,8 +428,10 @@ impl WrongType {
 ///
 /// # Arrays
 ///
-/// `FromSql` is implemented for `Vec<T>` where `T` implements `FromSql`, and
-/// corresponds to one-dimensional Postgres arrays.
+/// `FromSql` is implemented for `Vec<T>` and `[T; N]` where `T` implements
+/// `FromSql`, and corresponds to one-dimensional Postgres arrays. **Note:**
+/// the impl for arrays only exist when the Cargo feature `array-impls` is
+/// enabled.
 pub trait FromSql<'a>: Sized {
     /// Creates a new value of this type from a buffer of data of the specified
     /// Postgres `Type` in its binary format.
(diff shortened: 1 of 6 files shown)


LukasKalbertodt:
For the `cargo test --all-features`, the Rust version in CI has to be 1.51. I assume you set an explicit Rust version in CI to make sure the MSRV is not accidentally bumped. So I changed the version to 1.51 only for the `--all-features` run. If you rather want me to bump it globally, let me know.

sfackler:
Once we're building against 2 rustc versions, I'm thinking it may be better to have a separate minimal "min-version" build that uses 1.46 just to smoke test to allow the main test build to use whatever version is easiest.

LukasKalbertodt:
So, should I just change the version to 1.51 for the full CI script for now?

sfackler:
Yeah let's just do that and I'll set up the other build tonight.

postgres-types/src/lib.rs
```diff
@@ -513,6 +515,47 @@ impl<'a, T: FromSql<'a>> FromSql<'a> for Vec<T> {
...
+
+        let mut values = array.values();
+        let out = array_init::try_array_init(|i| {
+            let v = values
+                .next()?
+                .ok_or_else(|| -> Box<dyn Error + Sync + Send> {
+                    format!("too few elements in array (expected {}, got {})", N, i).into()
+                })?;
+            T::from_sql_nullable(member_type, v)
+        });
```
sfackler, line 539: Nit: I think we'll want to early-exit with a `?` here, or an early inner value decode error will actually end up being reported as an "excess elements in array" error.

LukasKalbertodt: Oh, very good point! Change is amended to previous commit and force pushed.

LukasKalbertodt:
I hope I'm not too pushy, but how about releasing a new version of `postgres-types`? As far as I can see, there have been three `postgres-types` related PRs since the last release and none of those should require a major version bump. Should be a fairly straight forward release? Thanks!

sfackler:
No worries! Will do in a couple hours.

sfackler:
Done!

PR #822, opened 2021-08-13
Add support for Postgres Interval type

tmuerell:
This patch implements `FromSql` for the Postgres Interval type and maps it to chrono Duration type.

### DIFF
--- a/postgres-types/src/chrono_04.rs
+++ b/postgres-types/src/chrono_04.rs
@@ -1,3 +1,4 @@
+use byteorder::{BigEndian, ByteOrder, ReadBytesExt};
 use bytes::BytesMut;
 use chrono_04::{DateTime, Duration, FixedOffset, Local, NaiveDate, NaiveDateTime, NaiveTime, Utc};
 use postgres_protocol::types;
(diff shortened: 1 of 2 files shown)


postgres-types/src/chrono_04.rs
```diff
@@ -153,3 +154,26 @@ impl ToSql for NaiveTime {
...
+    accepts!(INTERVAL);
+}
+
+#[inline]
+fn interval_from_sql(mut buf: &[u8]) -> Result<Duration, Box<dyn std::error::Error + Sync + Send>> {
+    let time = buf.read_i64::<BigEndian>()?;
+    let seconds = Duration::seconds(time / 1000000);
+    let frac = Duration::microseconds(time % 1000000);
+    let day = Duration::days(buf.read_i32::<BigEndian>()?.into());
+    let month = Duration::seconds((buf.read_i32::<BigEndian>()? * 30 * 86400).into());
```
sfackler, line 176: Many months are not 30 days long, and some days are not 86,400 seconds long. That is why Postgres stores those separately. A chrono Duration cannot represent that type properly.

tmuerell: I see what you mean. Below the day level it would work fine, but above it has a slightly different semantic because of the separately stored `day` and `month` values.

I would introduce a similar rust struct to represent the Postgres INTERVAL type for this instead. Ok for you?

sfackler: I think that would be best handled via a separate crate so the design can iterate.

tmuerell: That's ok. I first implement it locally for my project and come up with a generic solution out of this.

PR #824, opened 2021-09-23
Add support for time 0.3

lnicola:


lnicola:
Can someone kick off the CI? I haven't contributed to this crate before.

postgres-types/src/time_03.rs
```diff
@@ -0,0 +1,109 @@
+use bytes::BytesMut;
+use postgres_protocol::types;
+use std::convert::TryFrom;
+use std::error::Error;
+use time_03::{Date, Duration, OffsetDateTime, PrimitiveDateTime, Time, UtcOffset};
+
+use crate::{FromSql, IsNull, ToSql, Type};
+
+#[rustfmt::skip]
```
sfackler, line 9: Why?

postgres-types/src/time_03.rs
```diff
@@ -0,0 +1,109 @@
...
+const fn base() -> PrimitiveDateTime {
+    PrimitiveDateTime::new(Date::__from_ordinal_date_unchecked(2000, 1), Time::MIDNIGHT)
```
sfackler, line 11: This does not seem like a public API?

postgres-types/src/time_03.rs
lnicola: It was copied from `time_02`, which uses a macro. But yeah, no point in having it here.

postgres-types/src/time_03.rs
lnicola: It's partially public. The problem is that we can `unwrap` in a `const` function, and calling that API seems less bad than bringing in a proc macro dependency.

paolobarbolini: The `date!` macro might help here

https://docs.rs/time/0.3.2/time/macros/index.html

lnicola: Removed the `const` and switched to the public API.

@paolobarbolini I don't think that proc macro is worth its weight here.

lnicola:
Can you start the CI again? Unfortunately it needs to be done after every push.

lnicola:
Fixed the tests, they now pass on my computer. I might send another PR to fix some doctests which don't build without the `runtime` feature

sfackler:
Thanks!

PR #825, opened 2021-09-24
Add feature gates for doctests

lnicola:


lnicola:
Quick r? @sfackler

lnicola:
Strange, I'm not sure what's happening:

```rust
//! use native_tls::{Certificate, TlsConnector};
//! # #[cfg(feature = "runtime")] {
//! use postgres_native_tls::MakeTlsConnector;
//! # }
//! use std::fs;
//!
//! # fn main() -> Result<(), Box<dyn std::error::Error>> {
//! # #[cfg(feature = "runtime")] {
//! let cert = fs::read("database_cert.pem")?;
//! let cert = Certificate::from_pem(&cert)?;
//! let connector = TlsConnector::builder()
//!     .add_root_certificate(cert)
//!     .build()?;
//! let connector = MakeTlsConnector::new(connector);
//!
//! let connect_future = tokio_postgres::connect(
//!     "host=localhost user=postgres sslmode=require",
//!     connector,
//! );
//! # }
//!
//! // ...
//! # Ok(())
//! # }
```

yields:

```
error[E0433]: failed to resolve: use of undeclared type `MakeTlsConnector`
   --> src/lib.rs:19:17
    |
17  | let connector = MakeTlsConnector::new(connector);
    |                 ^^^^^^^^^^^^^^^^
    | 
   ::: ~/native-tls-0.2.8/src/lib.rs:458:1
    |
458 | pub struct TlsConnector(imp::TlsConnector);
    | ------------------------------------------- similarly named struct `TlsConnector` defined here
    |
help: a struct with a similar name exists
    |
17  | let connector = TlsConnector::new(connector);
    |                 ^^^^^^^^^^^^
help: consider importing this struct
    |
3   | use postgres_native_tls::MakeTlsConnector;
```

So the feature gate enables the test code, but removes the import.

postgres-native-tls/src/lib.rs
```diff
@@ -4,10 +4,13 @@
 //!
 //! ```no_run
 //! use native_tls::{Certificate, TlsConnector};
+//! # #[cfg(feature = "runtime")] {
```
sfackler, line 7: This is not valid syntax.

jyn514: This is valid, it just doesn't have the expected effect - it's a hidden line which only compiles the block when `feature = "runtime"` is set. But that feature will never be set for doctests, only for the library itself.

sfackler:
```
error: expected item after attributes
 --> src/main.rs:1:1
  |
1 | #[cfg(feature = "runtime")] {
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `playground` due to previous error
```

You cannot define bare blocks at module level.

lnicola: Oh, that was a stupid mistake, thanks.

PR #841, opened 2021-11-23
Fix handling of raw ident fields in derive

### LINKED ISSUE #839: `"__r#type"` is not a valid identifier
I currently have to work with the raw identifier `r#type`, as one database table is using this. I can't change this name, as i'm merely reading out the data from a pre-existing format.

However, when i try to put in a struct format, like so;

```rs
#[derive(Debug, ToSql, FromSql)]
struct DeviceInboxMessageFormat {
    pub messages: Json<HashMap<UserId, HashMap<DeviceId, Value>>>,
    pub sender: UserId,
    pub r#type: String,
}
```

The proc-macro panics on;

```
error: proc-macro derive panicked
  --> adapters/synapse/src/extract.rs:83:24
   |
83 | #[derive(Debug, ToSql, FromSql)]
   |                        ^^^^^^^
   |
   = help: message: `"__r#type"` is not a valid identifier
```

I think this problem will be the same across all raw identifiers.

sfackler:
Closes #839

### DIFF
--- a/postgres-derive/src/composites.rs
+++ b/postgres-derive/src/composites.rs
@@ -14,7 +14,13 @@ impl Field {
 
         let ident = raw.ident.as_ref().unwrap().clone();
         Ok(Field {
-            name: overrides.name.unwrap_or_else(|| ident.to_string()),
+            name: overrides.name.unwrap_or_else(|| {
+                let name = ident.to_string();
+                match name.strip_prefix("r#") {
+                    Some(name) => name.to_string(),
+                    None => name,
+                }
+            }),
             ident,
             type_: raw.ty.clone(),
         })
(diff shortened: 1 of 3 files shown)


postgres-derive/src/composites.rs
```diff
@@ -14,7 +14,13 @@ impl Field {
 
         let ident = raw.ident.as_ref().unwrap().clone();
         Ok(Field {
-            name: overrides.name.unwrap_or_else(|| ident.to_string()),
+            name: overrides.name.unwrap_or_else(|| {
```
sfackler, line 17: @dtolnay am I right in thinking that there's nothing that uses IdentFragment to create a string?

dtolnay: Not that I know of. IdentFragment isn't intended for anything other than using in format_ident.

Issue #843, opened 2021-12-07
[Question] Owned query parameters


elpiel:
It's not that often that I get into this problem, however, if you're making a query you always have to pass `&dyn ToSql + Sync`.
This is not sufficient in some cases, e.g. if you're pushing your parameters into a vec in one function scope to build the parameters list and in another (the caller) you try to use the parameters and pass them to `client.query`

Is there a way we can make query or similar function work with `Box<dyn ToSql>`. I've tried implementing it for `tokio-postgres` but I quickly bumped into some issues with the implementation.

It's also not possible to use something like `Vec<Box<dyn ToSql + Sync + Send>` if you want to send the value to the `fn query` since query is expecting a `ToSql + Sync`

sfackler:
You can use `query_raw` for this: https://docs.rs/tokio-postgres/latest/tokio_postgres/struct.Client.html#method.query_raw

elpiel:
> You can use `query_raw` for this: https://docs.rs/tokio-postgres/latest/tokio_postgres/struct.Client.html#method.query_raw

I managed to take a look at the issue again, however, this approach does not work because `Box<dyn ToSql + Sync>` does not implement `BorrowToSql`

![Screenshot from 2021-12-16 10-00-38](https://user-images.githubusercontent.com/8925621/146331571-b093a103-8fbd-48a4-8047-f0aa40743a28.png)

sfackler:
You can transform the iterator to convert the `Box<dyn ToSql>` to `&dyn ToSql`, but that PR seems like a reasonable thing to add as well!

elpiel:
I was actually getting an error with the iterator too, IIRC because of a `Send/Sync` issue.

PS: I still have no idea how the `slice_iter` magic works

PR #853, opened 2021-12-16
Boxed owned query parameters

elpiel:
Related issue: #843 

This allows usage of `Vec<Box<dyn ToSql + Sync>>` & `Vec<Box<dyn ToSql + Sync + Send>>` to be used with `Client.query_raw`

```rust
use futures::TryStreamExt;

fn build_params( ... ) -> Vec<Box<(dyn ToSql + Sync + Send)>> { ... }

async fn query_database() {
    let stmt = ....;
    let params = build_params( ... );
    let rows: Vec<Row> = client.query_raw(&stmt, params).await?.try_collect().await?;
}
```

### DIFF
--- a/postgres-types/src/lib.rs
+++ b/postgres-types/src/lib.rs
@@ -1044,6 +1044,23 @@ impl BorrowToSql for &dyn ToSql {
     }
 }
 
+impl sealed::Sealed for Box<dyn ToSql + Sync> {}
+
+impl BorrowToSql for Box<dyn ToSql + Sync> {
+    #[inline]
+    fn borrow_to_sql(&self) -> &dyn ToSql {
+        self.as_ref()
+    }
+}
+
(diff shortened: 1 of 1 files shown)


sfackler:
Seems reasonable to me. Anything still need to added to this before switching it from a draft?

elpiel:
> Seems reasonable to me. Anything still need to added to this before switching it from a draft?

Would there be an implication of the Boxed type when calling `query()`?
It would be nice to be able to call the `query()` method with types like `Vec<Box<dyn ToSql>>`

elpiel:
Here's another thing:
Would you consider `impl<T: ToSql> From<T> for Box<dyn ToSql +Send +Sync>` & for the `Box<dyn ToSql + Sync>`?

elpiel:
@sfackler could you merge this PR?

PR #857, opened 2021-12-29
Derive transparent

malobre:
# `#[postgres(transparent)]`

# Acknowledgments
- @nolanderc for the original PR: https://github.com/sfackler/rust-postgres-derive/pull/18

# Example
```rust
#[derive(ToSql, FromSql)]
#[postgres(transparent)]
struct UserId(i32);
```

### DIFF
--- a/postgres-derive-test/src/compile-fail/invalid-transparent.rs
+++ b/postgres-derive-test/src/compile-fail/invalid-transparent.rs
@@ -1,0 +1,35 @@
+use postgres_types::{FromSql, ToSql};
+
+#[derive(ToSql, Debug)]
+#[postgres(transparent)]
+struct ToSqlTransparentStruct {
+    a: i32
+}
+
+#[derive(FromSql, Debug)]
+#[postgres(transparent)]
+struct FromSqlTransparentStruct {
+    a: i32
+}
+
+#[derive(ToSql, Debug)]
+#[postgres(transparent)]
+enum ToSqlTransparentEnum {
+    Foo
+}
+
+#[derive(FromSql, Debug)]
+#[postgres(transparent)]
+enum FromSqlTransparentEnum {
+    Foo
+}
+
+#[derive(ToSql, Debug)]
+#[postgres(transparent)]
+struct ToSqlTransparentTwoFieldTupleStruct(i32, i32);
+
+#[derive(FromSql, Debug)]
+#[postgres(transparent)]
+struct FromSqlTransparentTwoFieldTupleStruct(i32, i32);
+
+fn main() {}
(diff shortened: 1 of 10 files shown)


postgres-derive/src/overrides.rs
```diff
@@ -30,6 +31,13 @@ impl Overrides {
                             return Err(Error::new_spanned(&meta.path, "unknown override"));
                         }
 
+                        if overrides.transparent {
+                            return Err(Error::new_spanned(
```
sfackler, line 35: Might be best to have some error checking at the end instead of having each branch here check.

sfackler:
Nice! Would also be good to have a few quick tests.

postgres-derive/src/overrides.rs
malobre: I moved the check to `fromsql.rs` & `tosql.rs`.

malobre:
I added some basic tests and updated some of the logic.

postgres-derive-test/src/compile-fail/invalid-transparent.rs
```diff
@@ -0,0 +1,27 @@
...
+#[postgres(transparent)]
+enum ToSqlTransparentEnum {
+    Foo
+}
+
+#[derive(FromSql, Debug)]
+#[postgres(transparent)]
+enum FromSqlTransparentEnum {
+    Foo
+}
```
sfackler, line 25: Can you add one more for a two-field tuple struct?

sfackler:
Can you add a little bit of docs to the postgres-types crate?

LTGM otherwise!

postgres-derive/src/fromsql.rs
```diff
@@ -11,9 +11,36 @@ use crate::overrides::Overrides;
 pub fn expand_derive_fromsql(input: DeriveInput) -> Result<TokenStream, Error> {
     let overrides = Overrides::extract(&input.attrs)?;
 
+    if overrides.name.is_some() && overrides.transparent {
```
sfackler, line 14: You should be able to just put this at the bottom of Overrides::extract.

malobre: I did it this way because I did not know where to take a span / or `dyn ToTokens` to construct a `syn::Error` in `Overrides::extract`. I took inspirations from `serde_derive` which does something similar.

malobre:
I added documentation and squashed all the commits into one.

postgres-derive/src/fromsql.rs
sfackler: Got it, thanks.

PR #859, opened 2021-12-31
Boxed owned query parameters w/ impl for From<T>

elpiel:
Extension of #853 that allows creation of a `Box<dyn ToSql + Send + Sync>` from `T: ToSql`

### DIFF
--- a/postgres-types/src/lib.rs
+++ b/postgres-types/src/lib.rs
@@ -1044,6 +1044,23 @@ impl BorrowToSql for &dyn ToSql {
     }
 }
 
+impl sealed::Sealed for Box<dyn ToSql + Sync> {}
+
+impl BorrowToSql for Box<dyn ToSql + Sync> {
+    #[inline]
+    fn borrow_to_sql(&self) -> &dyn ToSql {
+        self.as_ref()
+    }
+}
+
+impl sealed::Sealed for Box<dyn ToSql + Sync + Send> {}
+impl BorrowToSql for Box<dyn ToSql + Sync + Send> {
+    #[inline]
+    fn borrow_to_sql(&self) -> &dyn ToSql {
+        self.as_ref()
+    }
+}
+
 impl sealed::Sealed for &(dyn ToSql + Sync) {}
 
 /// In async contexts it is sometimes necessary to have the additional
--- a/postgres-types/src/lib.rs
+++ b/postgres-types/src/lib.rs
@@ -1085,3 +1085,9 @@ where
         self
     }
 }
+
+impl<'a, T: ToSql + Send + Sync + 'a> From<T> for Box<dyn ToSql + Send + Sync + 'a> {
+    fn from(to_sql: T) -> Self {
+        Box::new(to_sql)
+    }
+}


postgres-types/src/lib.rs
```diff
@@ -1068,3 +1085,9 @@ where
         self
     }
 }
+
+impl<'a, T: ToSql + Send + Sync + 'a> From<T> for Box<dyn ToSql + Send + Sync + 'a> {
+    fn from(to_sql: T) -> Self {
+        Box::new(to_sql)
+    }
+}
```
elpiel, line 1093: We could also add impl for Sync only, although I'm not sure if/when it would be needed:

```suggestion
}

impl<'a, T: ToSql + Sync + 'a> From<T> for Box<dyn ToSql + Sync + 'a> {
    fn from(to_sql: T) -> Self {
        Box::new(to_sql)
    }
}

```

sfackler:
I'm not sure I've ever seen this conversion before for other traits. What's the rationale for it?

elpiel:
> I'm not sure I've ever seen this conversion before for other traits. What's the rationale for it?

As mentioned in the comment, this allows you to build a `Box<dyn ToSql>` from any type and can be used to create a `Vec` of these values that can be passed to the `query_raw()`.

This is necessary in order to build these values dynamically and I couldn't find any other way to do so.
The example of `query_raw` uses only `String`s (1 type) so this issue is not present. 

Example:

```
// These parameters can now come from e.g. another function call and can be mixed types
let params: Vec<Box<dyn ToSql + Sync + Send>> = vec![
    "first param".into(),
    42.into(),
];
let mut it = client.query_raw(
    "SELECT foo FROM bar WHERE biz = $1 AND baz = $2",
    params,
).await?;
```

sfackler:
Can't you just call Box::new yourself?

elpiel:
> Can't you just call Box::new yourself?

Maybe I just don't know how to do the casting since I get a type error:

```
17 |         pub fn get_key(&self, key: AllowedKey) -> Option<Box<dyn ToSql + Sync + Send>> {
   |                                                   ------------------------------------ expected `std::option::Option<Box<(dyn ToSql + Send + Sync + 'static)>>` because of return type
18 |             match key {
19 |                 AllowedKey::CampaignId => self.campaign_id.map(Box::new),
   |                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected trait object `dyn ToSql`, found struct `CampaignId`
   |
   = note: expected enum `std::option::Option<Box<(dyn ToSql + Send + Sync + 'static)>>`
              found enum `std::option::Option<Box<CampaignId>>`
```

I've also tried the following but then I get an error because of `Sized` requirements for `Box::new`
```rust
self.campaign_id.map(|s| Box::new(s as _)),
```

sfackler:
Without seeing the rest of the code I can't be 100% sure, but one of these should work: `self.campaign_id.map(|s| Box::new(s))` or `self.campaign_id.map(|s| Box::new(s) as _)`.

elpiel:
> Without seeing the rest of the code I can't be 100% sure, but one of these should work: `self.campaign_id.map(|s| Box::new(s))` or `self.campaign_id.map(|s| Box::new(s) as _)`.

Here are links to the playground with 3 non-working examples:

- `as _`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=042514743d7600b0b5f1cfcbaed37a51

- `Box::new(value.into())` (this is closest but it complains about `Sized`): https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=042514743d7600b0b5f1cfcbaed37a51

- And just `Box::new(value)`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=097653481061736cf66bddb06135efe1

sfackler:
https://docs.rs/security-framework/latest/security_framework/random/struct.SecRandom.html

elpiel:
> https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=26af9ee9d1ab04b26a506c8b33966769

Good point! Didn't think of this :smile:

PR #867, opened 2022-02-07
Tokio-Postgres Make function from_str public

drconopoima:
Seems to me an oversight there is no public handler for URL config that is publicly available. I tried to import URLParser and use method from_string in my code and they were private.


sfackler:
What happens when you try to compile the code with your change?

drconopoima:
Ok, I see now that it says it's implied. I wasn't able to use it without adding "use std::str::FromStr"

PR #874, opened 2022-02-22
postgres-types: make all vector types' Kind Simple

sploiselle:
Previously, some PG legacy vectors were `Kind::Simple` (e.g. `Inner::TsVector`), while others were `Kind:Array` (e.g. `Inner::Int2Vector`). This PR makes all vectors `Kind::Simple`.

Motivating this change is the fact that the behavior of vector types differs from arrays, e.g. their parsing differs from arrays.

An alternative would be introducing another `Kind` for vectors, which all have similar input/output + casting semantics; glad to take this approach, but given that there were other `Kind::Simple` vectors, assumed this approach sufficed.


sploiselle:
@sfackler ty for the quick response

PR #875, opened 2022-03-03
Add ltree, lquery and ltxtquery support

### LINKED ISSUE #389: Support the `ltree` data type

halfmatthalfcat:
This adds [Postgres ltree, lquery and ltxtquery](https://www.postgresql.org/docs/current/ltree.html) support to rust-postgres. For the uninitiated, ltree ("label tree") is a Postgres datatype that follows a dot-notated search path for a given row. Lquery and ltxtquery allow users to search for rows that match an ltree. There are many query operators to help traverse the tree in a given query.

There is some commentary in #389 on why we **are not** adding `Inner`s for the three datatypes. In summary, because `ltree` is an extension (one that ships with Postgres), it's Oids are not static like many other datatypes. This renders the `Inner` loading by Oid unusable and thus we must match on the `Type` name, like we do for `citext`.

Long term motivations for this are to incorporate these changes into quaint and then eventually into Prisma.

Fixes  #389

### DIFF
--- a/postgres-protocol/src/types/mod.rs
+++ b/postgres-protocol/src/types/mod.rs
@@ -1060,18 +1060,59 @@ impl Inet {
     }
 }
 
-/// Serializes a Postgres l{tree,query,txtquery} string
+/// Serializes a Postgres ltree string
 #[inline]
 pub fn ltree_to_sql(v: &str, buf: &mut BytesMut) {
-    // A version number is prepended to an Ltree string per spec
+    // A version number is prepended to an ltree string per spec
     buf.put_u8(1);
     // Append the rest of the query
     buf.put_slice(v.as_bytes());
 }
 
-/// Deserialize a Postgres l{tree,query,txtquery} string
+/// Deserialize a Postgres ltree string
 #[inline]
 pub fn ltree_from_sql(buf: &[u8]) -> Result<&str, StdBox<dyn Error + Sync + Send>> {
-    // Remove the version number from the front of the string per spec
-    Ok(str::from_utf8(&buf[1..])?)
+    match buf {
+        // Remove the version number from the front of the ltree per spec
+        [1u8, rest @ ..] => Ok(str::from_utf8(rest)?),
+        _ => Err("ltree version 1 only supported".into()),
+    }
+}
+
+/// Serializes a Postgres lquery string
+#[inline]
+pub fn lquery_to_sql(v: &str, buf: &mut BytesMut) {
+    // A version number is prepended to an lquery string per spec
+    buf.put_u8(1);
+    // Append the rest of the query
+    buf.put_slice(v.as_bytes());
+}
+
+/// Deserialize a Postgres lquery string
+#[inline]
+pub fn lquery_from_sql(buf: &[u8]) -> Result<&str, StdBox<dyn Error + Sync + Send>> {
+    match buf {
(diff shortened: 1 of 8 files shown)


postgres-protocol/src/types/mod.rs
```diff
@@ -1059,3 +1059,19 @@ impl Inet {
...
+    buf.put_u8(1);
+    // Append the rest of the query
+    buf.put_slice(v.as_bytes());
+}
+
+/// Deserialize a Postgres l{tree,query,txtquery} string
+#[inline]
+pub fn ltree_from_sql(buf: &[u8]) -> Result<&str, StdBox<dyn Error + Sync + Send>> {
+    // Remove the version number from the front of the string per spec
+    Ok(str::from_utf8(&buf[1..])?)
```
sfackler, line 1076: We should check the version number to ensure we are actually able to parse the format. See the jsonb conversion impls for reference.

postgres-types/src/lib.rs
```diff
@@ -612,14 +613,26 @@ impl<'a> FromSql<'a> for Box<str> {
...
 impl<'a> FromSql<'a> for &'a str {
-    fn from_sql(_: &Type, raw: &'a [u8]) -> Result<&'a str, Box<dyn Error + Sync + Send>> {
-        types::text_from_sql(raw)
+    fn from_sql(ty: &Type, raw: &'a [u8]) -> Result<&'a str, Box<dyn Error + Sync + Send>> {
+        match *ty {
+            ref ty if (
+                ty.name() == "ltree"    ||
+                ty.name() == "lquery"   ||
+                ty.name() == "ltxtquery"
+            ) => types::ltree_from_sql(raw),
```
sfackler, line 622: Are these types guaranteed to use the same encoding? It may be better to expose separate to_sql and from_sql methods in postgres_protocol for each even if they currently do the same thing.

postgres-protocol/src/types/mod.rs
halfmatthalfcat: Should now be checking for correct version number (1) and erroring on any others.

postgres-types/src/lib.rs
halfmatthalfcat: Split out to/from for each type

halfmatthalfcat:
@sfackler I think everything should be sorted. Added more tests around the protocol parsers. Let me know if you see anything else glaring. Thanks!

brassel:
Looking forward to see this!
Thank you @halfmatthalfcat and @sfackler!

halfmatthalfcat:
Hey @sfackler I think since I haven't contributed before I need some assistance kicking off another build. Thanks for your help!

halfmatthalfcat:
Ok @sfackler, sorry about that delay. I didn't know about the `matches!` macro and added `ltree` as an extension in the sql setup file, was running into a race condition during the tests.

Confirmed the latest build is working on my fork: https://github.com/halfmatthalfcat/rust-postgres/actions/runs/1996247393

I rebased as well so we *should* be ready to go on this.

halfmatthalfcat:
Hey @sfackler can we kick the tires on this one more time?

sfackler:
Kicked - sorry for the delay. I really with github would auto-build once you've been approved a single time or something... :(

PR #887, opened 2022-04-19
WIP Allow passing precomputed SCRAM keys via Config

funbringer:
According to https://datatracker.ietf.org/doc/html/rfc5802#section-3, SCRAM protocol explicitly allows client to use a `ClientKey` & `ServerKey` pair instead of a password to perform authentication. This is also useful for proxy implementations which would like to leverage `rust-postgres`. This patch adds the ability to do that.

@sfackler Please consider this a draft. The tests are not there yet, but hopefully this should be enough to start a discussion. We'd love to share this code with the upstream, even if some changes are necessary.

### DIFF
--- a/postgres-protocol/src/authentication/sasl.rs
+++ b/postgres-protocol/src/authentication/sasl.rs
@@ -96,14 +96,32 @@ impl ChannelBinding {
     }
 }
 
+/// A pair of keys for the SCRAM-SHA-256 mechanism.
+/// See <https://datatracker.ietf.org/doc/html/rfc5802#section-3> for details.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct ScramKeys<const N: usize> {
+    /// Used by server to authenticate client.
+    pub client_key: [u8; N],
+    /// Used by client to verify server's signature.
+    pub server_key: [u8; N],
+}
+
+/// Password or keys which were derived from it.
+enum Credentials<const N: usize> {
+    /// A regular password as a vector of bytes.
+    Password(Vec<u8>),
+    /// A precomputed pair of keys.
+    Keys(Box<ScramKeys<N>>),
+}
+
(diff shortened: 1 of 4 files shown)


sfackler:
Does postgres in particular support the use of precomputed keys?

funbringer:
> Does postgres in particular support the use of precomputed keys?

Yes, it does. Once you've computed the `ClientKey`, you may use it as many times as you like, for as long as the authentication info (aka scram secret in postgres terminology, see [`scram_build_secret`](https://github.com/postgres/postgres/blob/5513dc6a304d8bda114004a3b906cc6fde5d6274/src/common/scram-common.c#L195)) stays the same.

sfackler:
I'm not sure I understand how this is useful in a proxy implementation. Is there a concern around the performance cost of 2 HMACs per connection?

kelvich:
I can try to give another explanation =)

So we have 3 parties:

* Client. He knows the raw password.
* Postgres. Stores SCRAM credentials (Iterations+Salt+ServerKey+StoredKey)
* Proxy. Need access to the same SCRAM credentials to accept a connection

So the proxy can implement a server-side part of the SCRAM and accept the connection. But then proxy doesn't have a plaintext password to make a second connection to the Postgres. Fortunately, SCRAM was designed [1] with that use-case in mind: it is enough to have a ClientProof (sent in client-final) and access to StoredKey to authenticate yourself at the SCRAM server.


[1] SCRAM RFC:
```
The mechanism permits the use of a server-authorized proxy without
requiring that proxy to have super-user rights with the back-end
server.
```

sfackler:
Cool, that makes sense. I am a bit hesitant to expose this in the main `tokio-postgres`/`postgres` API though. It seems like a proxy would be operating directly on the raw protocol rather than a higher level client, right?

Issue #912, opened 2022-06-28
Request: remove arbitrary restriction on GenericClient to allow for unit tests to run without a local Postgres instance.


Bjohnson131:
As the title says, removing the seal on GenericClient would allow for Unit Tests to be constructed much more easily.

andrewbaxter:
It seems like this is blocking a better implementation of GenericClient in deadpool_postgres per here: https://github.com/bikeshedder/deadpool/issues/56

shaylew:
I've been investigating adding tracing as a wrapper crate, and not being able to have the tracing-enabled wrapper client implement GenericClient is pretty annoying. 

Not sure if unsealing GenericClient is sufficient, but it'd definitely have helped so far.

PR #913, opened 2022-06-28
removed sealed trait

Bjohnson131:
implementation of https://github.com/sfackler/rust-postgres/issues/912

### DIFF
--- a/postgres/src/generic_client.rs
+++ b/postgres/src/generic_client.rs
@@ -4,14 +4,8 @@ use crate::{
     ToStatement, Transaction,
 };
 
-mod private {
-    pub trait Sealed {}
-}
-
 /// A trait allowing abstraction over connections and transactions.
-///
-/// This trait is "sealed", and cannot be implemented outside of this crate.
-pub trait GenericClient: private::Sealed {
+pub trait GenericClient {
     /// Like `Client::execute`.
     fn execute<T>(&mut self, query: &T, params: &[&(dyn ToSql + Sync)]) -> Result<u64, Error>
     where
@@ -69,8 +63,6 @@ pub trait GenericClient: private::Sealed {
     /// Like `Client::transaction`.
     fn transaction(&mut self) -> Result<Transaction<'_>, Error>;
 }
-
-impl private::Sealed for Client {}
 
 impl GenericClient for Client {
     fn execute<T>(&mut self, query: &T, params: &[&(dyn ToSql + Sync)]) -> Result<u64, Error>
@@ -149,8 +141,6 @@ impl GenericClient for Client {
         self.transaction()
     }
 }
-
-impl private::Sealed for Transaction<'_> {}
 
 impl GenericClient for Transaction<'_> {
     fn execute<T>(&mut self, query: &T, params: &[&(dyn ToSql + Sync)]) -> Result<u64, Error>


sfackler:
Can you show me an example of an implementation of this trait you would make if this were merged?

Bjohnson131:
Sure! give me a bit.

Bjohnson131:
So here's the most basic implementation that I could come up with,  There are some changes that need to be made to the PR to get this to work properly. Name, that the members of a Row be made pub, but I'm sure there's other things I could add in.

```rs 

use postgres::{CopyInWriter, CopyOutReader, RowIter, Transaction};
use tokio_postgres::types::{BorrowToSql, ToSql, Type};
use tokio_postgres::{Error, Row, SimpleQueryMessage, Statement, ToStatement};

struct mock_client {
    should_error: bool,
    execute_result: u64,
    row_result: Row,
}

impl postgres::GenericClient for mock_client {
    fn execute<T>(&mut self, _query: &T, _params: &[&(dyn ToSql + Sync)]) -> Result<u64, Error>
    where
        T: ?Sized + ToStatement,
    {
        if self.should_error {
            return Err(Error::Error());
        }
        return Ok(self.execute_result);
    }

    fn query<T>(&mut self, _query: &T, _params: &[&(dyn ToSql + Sync)]) -> Result<Vec<Row>, Error>
    where
        T: ?Sized + ToStatement,
    {
        if self.should_error {
            return Err(Error::Error());
        }
        return Ok(vec![self.row_result]);
    }

    fn query_one<T>(&mut self, _query: &T, _params: &[&(dyn ToSql + Sync)]) -> Result<Row, Error>
    where
        T: ?Sized + ToStatement,
    {
        if self.should_error {
            return Err(Error::Error());
        }
        // Alternatively: self.row.clone()
        return Ok(Row {
            statement: self.row_result.statement.clone(),
            body: DataRowBody{
                self.row_result.body.storage.clone(),
                self.row_result.body.len.clone(),
            },
            ranges: self.row_result.ranges.clone(),
        });
    }
//[...]
}

```

and so on.


The idea is that since I'm writing unit tests to handle responses from the crate, you can craft those responses yourself as test-cases instead of having a live instance.  this way, we could make our functions use a 'GenericClient' instead of a postgres::client as a parameter, allowing us to pass in a 'mock_client' for our unit tests, and a postgres::client for live code.

sfackler:
I don't think I would want to make types like Row publicly constructible. If you want to do this kind of unit testing, I would recommend creating your own trait hierarchy that would allow you to mock out all of those types together.

Bjohnson131:
So one should make my own data type derived from a row, and mock that, if
I'm understanding correctly?

sfackler:
You'd make a trait that you implement for Row and your own testing type.

sfackler:
How are you going to construct the `Row` your trait implementation has to return?

Bjohnson131:
I suppose there would need to be a change to tokio-postgres. Should I add that to the PR?

sfackler:
What would that change be? If it's a breaking one that's not going to work until I get ready to cut the next major version bump.

Bjohnson131:
give me a few weeks to get a solution that's a drop-in for the existing
code, and maybe some tests. feel free to assign me to the pr. I'll @ you
when I have something ready to review.. Marking as a draft for now.

PR #915, opened 2022-06-29
feat: support [u8; N] as BYTEA

xoac:
There is a implementation for `&[u8]` but lacking implementation of `[u8; N]` what this PR is try to add.
Waiting for feedback


sfackler:
This already exists under the `array-impls` Cargo feature.

xoac:
Could you show me where? There is general implementation for [T; N] but I don't see it could map to [u8; N].

Also if this would be implemented my PR should not compile.
Also my LSP provide `to_sql` method for `[0u32, 0]` but not for `[0u8, 0]`

sfackler:
Oh sorry - I misread!

sfackler:
Could you update the `ToSql` docs as well? LGTM otherwise.

sfackler:
Thanks!

PR #919, opened 2022-07-04
Add FromSql / ToSql for smol_str::SmolStr.

BratSinot:


postgres-types/src/lib.rs
```diff
@@ -629,6 +630,18 @@ impl<'a> FromSql<'a> for Box<str> {
     }
 }
 
+#[cfg(feature = "smol_str")]
+impl<'a> FromSql<'a> for smol_str::SmolStr {
```
sfackler, line 634: These impls should go in their own file like the other optional types.

postgres-types/Cargo.toml
```diff
@@ -48,3 +48,5 @@ uuid-08 = { version = "0.8", package = "uuid", optional = true }
 uuid-1 = { version = "1.0", package = "uuid", optional = true }
 time-02 = { version = "0.2", package = "time", optional = true }
 time-03 = { version = "0.3", package = "time", default-features = false, optional = true }
+
+smol_str = { version = "0.1.23", default-features = false, optional = true }
```
sfackler, line 52: This should be `smol_str-01` = { version = "0.1.23", package = "smol_str", default-features = false, optional = true }`.

sfackler:
Can you also update the docs for FromSql and ToSql?

BratSinot:
> Can you also update the docs for FromSql and ToSql?

Done.

sfackler:
It looks like it needs a `cargo fmt`. If you pull in master the clippy failure should be fixed.

BratSinot:
> It looks like it needs a `cargo fmt`. If you pull in master the clippy failure should be fixed.

Yeah, I saw that clippy warning too =)

sfackler:
Thanks!

Issue #570, opened 2020-02-09
Derive ToSql and FromSql implementations for generic data types


alexwl:
At the moment, it is impossible to derive `FromSql` or `ToSql` for a data type with generic parameters.

```rust
#[derive(FromSql)]
struct Data<'a> {
    id: i32,
    description: &'a str,
}
```

Error: missing lifetime specifier

```rust
#[derive(ToSql)]
struct Data<T: ToSql> {
    id: i32,
    description: String,
    links: Vec<T>,
}
```

Error: wrong number of type arguments

The generated code for `FromSql` and `ToSql` doesn't take into account the fact that a structure may have lifetime or type parameters:

https://github.com/sfackler/rust-postgres/blob/cc9b8232a858099e11f8b9e8632566f9b6103a81/postgres-derive/src/fromsql.rs#L62-L64
https://github.com/sfackler/rust-postgres/blob/cc9b8232a858099e11f8b9e8632566f9b6103a81/postgres-derive/src/tosql.rs#L59

Adding support for lifetimes should be relatively straightforward, not sure about type parameters though.

LukasKalbertodt:
I am also interested in that. In my case, I'm reading a `text[]` column with many elements. Since I have to post-process these texts anyway and discard many elements, it would be a shame having to allocate all those intermediate `String`s. Of course, this is already possible since `FromSql<'a>` is implemented for `&'a str`. Nice! But unfortunately, my array is actually a custom composite type with a `bigint` field and a `text` field. And now it gets too complicated. It would be really nice to just be able to write the first example in the top comment. FWIW: I am using that impl manually inside another `FromSql` impl, and not via `Row::get`.

The linked code snippets are not quite up to date anymore, but there is no support for this yet.

https://github.com/sfackler/rust-postgres/blob/98f5a11bc0a8e451552d8941ffa078c7eb6cd60c/postgres-derive/src/fromsql.rs#L118-L122

https://github.com/sfackler/rust-postgres/blob/98f5a11bc0a8e451552d8941ffa078c7eb6cd60c/postgres-derive/src/fromsql.rs#L236-L246

I wonder what this comment means: why would it not make sense? Is there anything I'm missing?

gleaming-pheasant:
Is there still appetite for this change? In a private project, I've created a derive macro which does exactly this, but includes a bounds helper attribute for creating these mappings, similar to [serde's serde(bound("'de: 'a"))](https://serde.rs/container-attrs.html#bound).

[...]

PR #931, opened 2022-08-09
Derive Generic FromSql/ToSql

matt-duch:
Allows for derivation of FromSql and ToSql for generic types. Borrowed types are allowed for ToSql, but not meaningful to support for FromSql (so not supported). (also noted in #570)

Would you be interested in this change? A test was added. Happy to iterate.

### DIFF
--- a/postgres-derive-test/src/composites.rs
+++ b/postgres-derive-test/src/composites.rs
@@ -1,6 +1,6 @@
-use crate::test_type;
+use crate::{test_type, test_type_asymmetric};
 use postgres::{Client, NoTls};
-use postgres_types::{FromSql, ToSql, WrongType};
+use postgres_types::{FromSql, FromSqlOwned, ToSql, WrongType};
 use std::error::Error;
 
 #[test]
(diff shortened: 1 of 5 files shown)


postgres-derive-test/src/composites.rs
```diff
@@ -238,3 +238,67 @@ fn raw_ident_field() {
 
     test_type(&mut conn, "inventory_item", &[(item, "ROW('foo')")]);
 }
+
+#[test]
+fn generics() {
+    #[derive(FromSql, Debug, PartialEq)]
+    struct InventoryItem<T: FromSqlOwned, U>
+    where
+        U: FromSqlOwned,
```
sfackler, line 247: Are these bounds actually required?

matt-duch: `FromSqlOwned` is required for `#[derive(FromSql)]` which is needed to run the test. Or were you asking about the derived `Debug` and `PartialEq`?

sfackler: Derive logic should normally insert those bounds on the generated impls as necessary - you don't need `where T: Clone` at the struct definition when using `#[derive(Clone)]` for example.

matt-duch: Just to make sure I understand the change request, the idea is add a FromSql<'a> bound on any generic type in the FromSql derive logic?

sfackler: Yep.

matt-duch: So I started implementing this, and ran into an ambiguity (where this differs from `Clone`, which is in the rust prelude). I need to create a path to the FromSql and ToSql traits, however it could be one of a few options:
1. From the postgres_derive crate (`postgres_derive::FromSql<'a, T>`, but I think most uses will not import this crate directly)
2. From the postgres_types crate (`postgres_types::FromSql<'a, T>`)
3. Some other crate that re-exports or renames those types (no way to use a path from a crate name)
4. Assume that it's `use`d somewhere (`FromSql<'a, T>`)
Any suggestions on which assumption to use?

sfackler: Here's how the path is already produced in the existing logic: https://github.com/sfackler/rust-postgres/blob/master/postgres-derive/src/fromsql.rs#L90

matt-duch: Thanks for pointing that out. Based on that, looks like it was option (2). Branch is updated, and tests pass locally.

AngusWaller:
Hope this gets in soon, derive is breaking on any struct that has borrowed or lifetime values

sfackler:
Sorry for the delay - LGTM other than the clippy failures!

matt-duch:
@sfackler  Sorry about that. Fixed, clippy runs locally, and pushed.

sfackler:
Thanks! I will cut a release tomorrow.

PR #933, opened 2022-08-11
i128 support

vporton:
- Untested.

- Contrived use of `NUMERIC` SQL type, fails on numbers with a fractional part.

### DIFF
--- a/postgres-protocol/src/types/mod.rs
+++ b/postgres-protocol/src/types/mod.rs
@@ -132,10 +132,26 @@ pub fn int8_to_sql(v: i64, buf: &mut BytesMut) {
     buf.put_i64(v);
 }
 
+/// Serializes to a `NUMERIC` value.
+#[inline]
+pub fn int16_to_sql(v: i128, buf: &mut BytesMut) {
+    buf.put_i128(v);
+}
+
 /// Deserializes an `INT8` value.
 #[inline]
 pub fn int8_from_sql(mut buf: &[u8]) -> Result<i64, StdBox<dyn Error + Sync + Send>> {
     let v = buf.read_i64::<BigEndian>()?;
+    if !buf.is_empty() {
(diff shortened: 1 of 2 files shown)


vitaly-burovoy:
Why do you think it can work?

Where are checks for fractional part (which should lead to failures)?

sfackler:
NUMERIC is an arbitrary precision decimal type, so I'm not sure this conversion is correct. Minimally it needs to handle the fractional part in some way.

vporton:
> Where are checks for fractional part (which should lead to failures)?

Doesn't the below code from my patch fail in the case of a fractional part?

```rust
    if !buf.is_empty() {
        return Err("invalid buffer size".into());
    }
```

vitaly-burovoy:
No. It is better to read how NUMERIC is sent/receive at PostgreSQL sources[1] and read existing implementations[2].

Tests for unknown (for you) behavior and how data is encoded is «must have».

—
[1] https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/backend/utils/adt/numeric.c;h=920a63b00816c0e665e94edd86a54dc031c1d35d;hb=refs/heads/master

[2] https://github.com/paupino/rust-decimal/blob/master/src/postgres/driver.rs

vitaly-burovoy:
P.S. in PG's sources functions which are used in the binary protocol are `numeric_recv` (line 991) and `numeric_send` (line 1079).

EvanCarroll:
This is not a good idea. 128 bits is 16 bytes. PostgreSQL has a fixed-width 16 byte type: [uuid](https://www.postgresql.org/docs/current/datatype-uuid.html). Ignoring the rendering, if you're going to implement a native store for i128 in PostgreSQL you'd be far better of using this type.

UUID is abused for this already too. 

https://dba.stackexchange.com/a/115316/2639

PR #936, opened 2022-08-14
impl ToSQL/FromSQL on url::Url

nikhiljha:
I have a whole bunch of url::Url types in my structs, so here's a simple impl that renders it out as a TEXT string and back (this is probably the best/most interoperable way to implement it?)

My commit message is horrible so you'll want to squash/merge this if you decide this is something you want.

### DIFF
--- a/postgres-types/Cargo.toml
+++ b/postgres-types/Cargo.toml
@@ -23,6 +23,7 @@ with-geo-types-0_7 = ["geo-types-0_7"]
 with-serde_json-1 = ["serde-1", "serde_json-1"]
 with-uuid-0_8 = ["uuid-08"]
 with-uuid-1 = ["uuid-1"]
+with-url-2 = ["url-2"]
 with-time-0_2 = ["time-02"]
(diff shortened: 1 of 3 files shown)


sfackler:
I'm not sure about adding these impls. I want to avoid to the greatest extent possible conversions that only work for a subset of inputs.

nikhiljha:
My impression was that all URLs (per the URL standard) can be represented as a string. Unless you mean inputs == what's in the database, but doesn't it seem reasonable to expect errors if you set your type to url::Url and then try to read a db column that has non-URLs?

notgull:
I think it would be better to have a conversion to a struct that uses `String` instead of `Url`. Is there a way to do that listlessly within the `url` crate?

sfackler:
Yes, there are many values of strings that can be stored in the database that are not valid URLs.

nikhiljha:
Okay so it looks like the URL crate is mappable to pguri, but I'm guessing you don't want to support postgres extensions from this crate.

In that case, thanks for your time! I'll close this.

PR #937, opened 2022-08-15
Encode Format based on Type

NAlexPear:
#836 has been helpful as a way of pairing parameter types with their encodings, but the `encode_format` method in the `ToSql` trait would be even more useful with the addition of a reference to `Type`. This would enable more granular encoding of parameters based on the types inferred by Postgres during the `prepare` phase, which would be useful when processing payloads from more general type systems like gRPC or JSON.

Since parameter types are already generated by the time `bind()` is called, it's easy to include them as part of the format-generating process. This PR simply adds a `Type` parameter to the `encode_format` method added to `ToSql`, which should not be a breaking change if it can be made before the next release.

### DIFF
--- a/tokio-postgres/src/query.rs
+++ b/tokio-postgres/src/query.rs
@@ -157,11 +157,6 @@ where
     I::IntoIter: ExactSizeIterator,
 {
     let param_types = statement.params();
-    let (param_formats, params): (Vec<_>, Vec<_>) = params
-        .into_iter()
-        .zip(param_types.iter())
-        .map(|(p, ty)| (p.borrow_to_sql().encode_format(ty) as i16, p))
-        .unzip();
     let params = params.into_iter();
 
     assert!(
@@ -170,6 +165,13 @@ where
         param_types.len(),
         params.len()
     );
+
+    let (param_formats, params): (Vec<_>, Vec<_>) = params
+        .zip(param_types.iter())
+        .map(|(p, ty)| (p.borrow_to_sql().encode_format(ty) as i16, p))
+        .unzip();
+
+    let params = params.into_iter();
 
     let mut error_idx = 0;
     let r = frontend::bind(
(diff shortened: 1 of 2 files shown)


tokio-postgres/src/query.rs
```diff
@@ -156,16 +156,18 @@ where
     I: IntoIterator<Item = P>,
     I::IntoIter: ExactSizeIterator,
 {
+    let param_types = statement.params();
     let (param_formats, params): (Vec<_>, Vec<_>) = params
         .into_iter()
-        .map(|p| (p.borrow_to_sql().encode_format() as i16, p))
+        .zip(param_types.iter())
```
sfackler, line 162: A zipped iterator stops as soon as the shortest iterator ends, so the length check below won't work properly. You may need to manually iterate.

NAlexPear: Would it be better (or sufficient) to move the length check before the `zip` to guard against the early-ending case?

sfackler: Oh yeah, we already require ExactSizeIterator so we can just move the assert.

NAlexPear: Great, updated [in this commit](https://github.com/sfackler/rust-postgres/pull/937/commits/4e0aaee8987fde55438d250130234e69023fe857).

tokio-postgres/src/query.rs
```diff
@@ -156,25 +156,29 @@ where
...
+        param_types.len(),
         params.len()
     );
 
+    let (param_formats, params): (Vec<_>, Vec<_>) = params
+        .zip(param_types.iter())
+        .map(|(p, ty)| (p.borrow_to_sql().encode_format(ty) as i16, p))
+        .unzip();
+
+    let params = params.into_iter();
```
sfackler, line 174: This is a no-op

sfackler:
I think you just need to rebase to fix the clippy error.

PR #944, opened 2022-08-26
Support keepalive interval and retries.

PG-kura:
hello. I use tokio-postgres a lot.

## Thing I want to do
I would like to customize interval and retries when keepalive is turned on.
Therefore, I propose to make it possible to specify the keepalive interval and retries in this pull-request.

## Features to add
Previously, the only keepalive parameter was `keepalives_idle`. This time, add `keepalives_interval` and `keepalives_retries`. `keepalives_interval` is available on neither Redox nor Solaris. `keepalives_retries` is available on neither Redox, Solaris nor Windows. This constraint aligns with that of [socket2::TcpKeepalive](https://github.com/rust-lang/socket2/blob/14e99cf989f54478f017a483c6e5905e5f2e55a4/src/lib.rs#L378-L388).

## Implementation
I thought that adding `keepalives_interval` and `keepalives_retries` blindly would increase the number of arguments for the [connect_socket::connect_socket()](https://github.com/sfackler/rust-postgres/blob/54331183eaa6ee57f0dfeb01770b93360544c904/tokio-postgres/src/connect_socket.rs#L12-L18) and make the source code messy.
So I defined a new struct `KeepaliveConfig`. It does not publish outside the crate.
As an alternative, I considered putting `socket2::TcpKeepalive` in `Config`, but I rejected it because `socket2::TcpKeepalive` has no getter method and is incompatible with `Config`.

### DIFF
--- a/tokio-postgres/src/config.rs
+++ b/tokio-postgres/src/config.rs
@@ -101,9 +101,9 @@ pub enum Host {
 /// * `keepalives_idle` - The number of seconds of inactivity after which a keepalive message is sent to the server.
 ///     This option is ignored when connecting with Unix sockets. Defaults to 2 hours.
 /// * `keepalives_interval` - The time interval between TCP keepalive probes.
-///     This option is ignored when connecting with Unix sockets. Available on neither Redox nor Solaris.
+///     This option is ignored when connecting with Unix sockets.
 /// * `keepalives_retries` - The maximum number of TCP keepalive probes that will be sent before dropping a connection.
-///     This option is ignored when connecting with Unix sockets. Available on neither Redox, Solaris nor Windows.
+///     This option is ignored when connecting with Unix sockets.
 /// * `target_session_attrs` - Specifies requirements of the session. If set to `read-write`, the client will check that
 ///     the `transaction_read_write` session parameter is set to `on`. This can be used to connect to the primary server
 ///     in a database cluster as opposed to the secondary read-only mirrors. Defaults to `all`.
@@ -177,9 +177,7 @@ impl Config {
     pub fn new() -> Config {
         let keepalive_config = KeepaliveConfig {
             idle: Duration::from_secs(2 * 60 * 60),
-            #[cfg(not(any(target_os = "redox", target_os = "solaris")))]
             interval: None,
-            #[cfg(not(any(target_os = "redox", target_os = "solaris", target_os = "windows")))]
             retries: None,
         };
         Config {
@@ -373,18 +371,12 @@ impl Config {
     /// On Windows, this sets the value of the tcp_keepalive struct’s keepaliveinterval field.
     ///
     /// This is ignored for Unix domain sockets, or if the `keepalives` option is disabled.
-    ///
-    /// Available on neither Redox nor Solaris.
-    #[cfg(not(any(target_os = "redox", target_os = "solaris")))]
     pub fn keepalives_interval(&mut self, keepalives_interval: Duration) -> &mut Config {
         self.keepalive_config.interval = Some(keepalives_interval);
         self
     }
 
     /// Gets the time interval between TCP keepalive probes.
-    ///
-    /// Available on neither Redox nor Solaris.
-    #[cfg(not(any(target_os = "redox", target_os = "solaris")))]
     pub fn get_keepalives_interval(&self) -> Option<&Duration> {
         self.keepalive_config.interval.as_ref()
     }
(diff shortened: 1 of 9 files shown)


tokio-postgres/src/config.rs
```diff
@@ -347,14 +359,53 @@ impl Config {
...
+        self.keepalive_config.idle
+    }
+
+    /// Sets the time interval between TCP keepalive probes.
+    /// On Windows, this sets the value of the tcp_keepalive struct’s keepaliveinterval field.
+    ///
+    /// This is ignored for Unix domain sockets, or if the `keepalives` option is disabled.
+    ///
+    /// Available on neither Redox nor Solaris.
+    #[cfg(not(any(target_os = "redox", target_os = "solaris")))]
```
sfackler, line 378: I think it'd be preferable to always define these fields and just not do anything with them when unsupported. Otherwise the crate's public API is tied to the platform support in socket2.

PG-kura: I fixed it.

tokio-postgres/src/config.rs
```diff
@@ -347,14 +357,41 @@ impl Config {
...
+    /// On Windows, this sets the value of the tcp_keepalive struct’s keepaliveinterval field.
+    ///
+    /// This is ignored for Unix domain sockets, or if the `keepalives` option is disabled.
+    pub fn keepalives_interval(&mut self, keepalives_interval: Duration) -> &mut Config {
+        self.keepalive_config.interval = Some(keepalives_interval);
+        self
+    }
+
+    /// Gets the time interval between TCP keepalive probes.
+    pub fn get_keepalives_interval(&self) -> Option<&Duration> {
```
sfackler, line 380: Duration is Copy, so this can just return `Option<Duration>` directly.

sfackler:
Looks good to me other than the one comment!

tokio-postgres/src/config.rs
PG-kura: I fixed it.

Added fix commit as existing function `get_connect_timeout()` returns reference as well (Commit 4882ecc2de0b02183f4c69249d59178ec5e11f40).

tokio-postgres/src/config.rs
```diff
@@ -326,8 +336,8 @@ impl Config {
 
     /// Gets the connection timeout, if one has been set with the
     /// `connect_timeout` method.
-    pub fn get_connect_timeout(&self) -> Option<&Duration> {
-        self.connect_timeout.as_ref()
+    pub fn get_connect_timeout(&self) -> Option<Duration> {
+        self.connect_timeout.as_ref().copied()
```
sfackler, line 340:
```suggestion
        self.connect_timeout
```

tokio-postgres/src/config.rs
```diff
@@ -347,14 +357,41 @@ impl Config {
...
+    ///
+    /// This is ignored for Unix domain sockets, or if the `keepalives` option is disabled.
+    pub fn keepalives_interval(&mut self, keepalives_interval: Duration) -> &mut Config {
+        self.keepalive_config.interval = Some(keepalives_interval);
+        self
+    }
+
+    /// Gets the time interval between TCP keepalive probes.
+    pub fn get_keepalives_interval(&self) -> Option<Duration> {
+        self.keepalive_config.interval.as_ref().copied()
```
sfackler, line 381:
```suggestion
        self.keepalive_config.interval
```

tokio-postgres/src/config.rs
```diff
@@ -347,14 +357,41 @@ impl Config {
...
+    ///
+    /// This is ignored for Unix domain sockets, or if the `keepalives` option is disabled.
+    pub fn keepalives_retries(&mut self, keepalives_retries: u32) -> &mut Config {
+        self.keepalive_config.retries = Some(keepalives_retries);
+        self
+    }
+
+    /// Gets the maximum number of TCP keepalive probes that will be sent before dropping a connection.
+    pub fn get_keepalives_retries(&self) -> Option<u32> {
+        self.keepalive_config.retries.as_ref().copied()
```
sfackler, line 394:
```suggestion
        self.keepalive_config.retries
```

tokio-postgres/src/config.rs
```diff
@@ -336,8 +336,8 @@ impl Config {
 
     /// Gets the connection timeout, if one has been set with the
     /// `connect_timeout` method.
-    pub fn get_connect_timeout(&self) -> Option<&Duration> {
-        self.connect_timeout.as_ref()
+    pub fn get_connect_timeout(&self) -> Option<Duration> {
```
sfackler, line 339: This is a breaking change, so we can't do it yet unfortunately.

PG-kura: sorry. I agree.

tokio-postgres/src/config.rs
PG-kura: Fixed in 5551b3e6cdd81c18d5ffb13ee5478fcd0e551475

tokio-postgres/src/config.rs
PG-kura: Fixed in 5551b3e6cdd81c18d5ffb13ee5478fcd0e551475

sfackler:
Thanks!

Issue #943, opened 2022-08-26
Possibility to avoid an extra DNS lookup? 


trungda:
Hi, thanks for the crate! We are using this crate heavily in prod. Nice work! I have a question regarding establishing TLS connection to a Postgres database. We are using the following code snippet to create a new connection to the db server. 
```
let builder = ...;
let tls = postgres_openssl::MakeTlsConnector::new(builder.build());

let (client, connection) = config.connect(tls).await?;
```

We provide the `host` and `port` parameters in the config object. The `host` here could be either a raw ipaddress or the hostname of the db server. In order to use TLS, we have to provide the hostname (since it's what the server cert includes). However, our use case is a bit special, we have the hostname but we also have the ipaddress of the db server upfront. 

I am wondering if we could use the ipaddress + hostname to create a TLS connection somehow without passing the hostname to `config.connect()` since I believe under the hood, [`tls.make_tls_connect`](https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/src/connect.rs#L43)` would make a DNS request to resolve the hostname but we already have the hostname resolved before trying to create the connection. 

One possibility I can think of is to extend the `Config` struct to have an optional `hostname` field just so we don't have to [build hostname from `host`](https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/src/connect.rs#L35)(if provided). If we can do that, we can still pass the "already known" ipaddress to `host` and pass the hostname to the new hostname field. What do you think?

sfackler:
It seems reasonable to follow psql and add the hostaddr field for the IP: https://www.postgresql.org/docs/14/libpq-connect.html#LIBPQ-PARAMKEYWORDS

trungda:
Thanks. I will open a PR. How often do you release to the crate? I am asking to see if we need to pull master in our code or we can wait for the PR to be merged and released.

sfackler:
I generally cut a release whenever people ask about wanting one to pull in a new feature :)

trungda:
Sweet. Let me try to open a today. Thanks @sfackler

trungda:
@sfackler  could you help take a look at https://github.com/sfackler/rust-postgres/pull/945? Thank you.

trungda:
#945 is merged. We can close this now.

PR #945, opened 2022-08-26
Add hostaddr support

trungda:
More details in https://github.com/sfackler/rust-postgres/issues/943. 
Essentially, if we could provide both `host` and `hostaddr`, we can avoid an extra DNS lookup to establish a connection to the DB. Specifically, if both `host` and `hostaddr` are provided, `hostaddr` is going to be used to create the socket, i.e., as a raw IP address. `host` is still needed for TLS validation. 

## Testing
`cargo test` returns 4 failing tests which are already failing in master
```
---- test::scram_user stdout ----
thread 'test::scram_user' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Tls, cause: Some(Error { code: -67843, message: "The certificate was not trusted." }) }', postgres-native-tls/src/test.rs:18:71
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- test::require stdout ----
thread 'test::require' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Tls, cause: Some(Error { code: -67843, message: "The certificate was not trusted." }) }', postgres-native-tls/src/test.rs:18:71

---- test::runtime stdout ----
thread 'test::runtime' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Tls, cause: Some(Error { code: -67843, message: "The certificate was not trusted." }) }', postgres-native-tls/src/test.rs:91:6

---- test::prefer stdout ----
thread 'test::prefer' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Tls, cause: Some(Error { code: -67843, message: "The certificate was not trusted." }) }', postgres-native-tls/src/test.rs:18:71


failures:
    test::prefer
    test::require
    test::runtime
    test::scram_user

test result: FAILED. 0 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
```
I would be happy to make them work again in future PRs.


tokio-postgres/src/config.rs
```diff
@@ -90,6 +91,17 @@ pub enum Host {
 ///     path to the directory containing Unix domain sockets. Otherwise, it is treated as a hostname. Multiple hosts
 ///     can be specified, separated by commas. Each host will be tried in turn when connecting. Required if connecting
 ///     with the `connect` method.
+/// * `hostaddr` - Numeric IP address of host to connect to. This should be in the standard IPv4 address format,
+///     e.g., 172.28.40.9. If your machine supports IPv6, you can also use those addresses.
+///     If this parameter is not specified, the value of `host` will be looked up to find the corresponding IP address,
+///     - or if host specifies an IP address, that value will be used directly.
+///     Using `hostaddr` allows the application to avoid a host name look-up, which might be important in applications
+///     with time constraints. However, a host name is required for verify-full SSL certificate verification.
+///     Note that `host` is always required regardless of whether `hostaddr` is present.
```
sfackler, line 100: Why is this? psql at least will try to use just hostaddr if possible.

tokio-postgres/src/config.rs
```diff
@@ -153,6 +169,7 @@ pub struct Config {
     pub(crate) application_name: Option<String>,
     pub(crate) ssl_mode: SslMode,
     pub(crate) host: Vec<Host>,
+    pub(crate) hostaddr: Vec<String>,
```
sfackler, line 172: I think it'd be good to store these as `IpAddr`s and have the setter take that type as well.

tokio-postgres/src/connect.rs
```diff
@@ -32,18 +41,28 @@ where
...
-
         let tls = tls
             .make_tls_connect(hostname)
             .map_err(|e| Error::tls(e.into()))?;
 
-        match connect_once(host, port, tls, config).await {
+        // If both host and hostaddr are specified, the value of hostaddr is used to to establish the TCP connection.
+        let hostaddr = match host {
+            Host::Tcp(_hostname) => match config.hostaddr.get(i) {
+                Some(hostaddr) if hostaddr.is_empty() => Host::Tcp(hostaddr.clone()),
```
sfackler, line 58: Isn't this checking if the hostaddr string is present but empty?

tokio-postgres/src/config.rs
trungda: Yeah I realized that too but making host optional would be a breaking change. Currently, in connect we iterate over the host vector. It'd be rather a big change. If you feel strongly about it I can try to fiddle with it.

tokio-postgres/src/config.rs
trungda: Sure. Will do.

tokio-postgres/src/connect.rs
trungda: Urgh. Thanks. I made this probably  due to resolving a clippy recommendation from != to !is_empty.

tokio-postgres/src/config.rs
sfackler: How would it be a breaking change? What code that used to work would now fail?

trungda: Sure! Let me play with it. So the length of the two vectors host and hostaddr must be equal or one must be empty?

tokio-postgres/tests/test/main.rs
```diff
@@ -147,6 +147,58 @@ async fn scram_password_ok() {
     connect("user=scram_user password=password dbname=postgres").await;
 }
 
+#[tokio::test]
+async fn host_only_ok() {
+    let _ = tokio_postgres::connect(
```
trungda, line 152: @sfackler I am not sure if these tests should belong here since we only test `connect_raw` here. But I think these tests do have some values.

tokio-postgres/src/config.rs
trungda: Thanks! Let me know what you think in the latest commit https://github.com/sfackler/rust-postgres/pull/945/files. I added a few tests to demonstrate how it works.

tokio-postgres/src/connect.rs
```diff
@@ -15,35 +15,67 @@ pub async fn connect<T>(
...
             .unwrap_or(5432);
 
+        // The value of host is always used as the hostname for TLS validation.
+        // postgres doesn't support TLS over unix sockets, so the choice for Host::Unix variant here doesn't matter
         let hostname = match host {
-            Host::Tcp(host) => host.as_str(),
-            // postgres doesn't support TLS over unix sockets, so the choice here doesn't matter
-            #[cfg(unix)]
-            Host::Unix(_) => "",
+            Some(Host::Tcp(host)) => host.as_str(),
```
sfackler, line 57: This is missing the `None` case. It should probably just stringify the hostaddr in that case.

trungda: Thanks! Sounds reasonable. Originally, I wanted to just fallback to empty string for both Unix and None case. It's better to just use the `hostaddr` as hostname if present. TLS verification is likely going to fail anyway, but still better than an empty string. Fixed in latest commit, thank you!

trungda:
@sfackler could you help take a look at the latest commit? Thank you :)

trungda:
@sfackler  could you help take a look? Thank you :)

trungda:
@sfackler could you help take a look? Thanks :)

trungda:
@sfackler would be great if you could take a look :) We have to use a forked version of rust-postgres now and I am worried it might be diverged from upstream :)

trungda:
@sfackler could you help take a look? Thank you :) !

trungda:
@sfackler happy holidays :) Could you help take a look at this PR? Thank you!

trungda:
@sfackler could you help take a look at this PR :) Thank you!

trungda:
@sfackler could you help take a look :) We have been running this branch in production for a few months now so I am confident that this works.

trungda:
Hi @sfackler , could you help review this PR when you have time? Thank you :)

trungda:
Hi @sfackler , could you help take a look at this PR when you have time :) Thank you!

sfackler:
Sorry for the very long delay! This looks good to me.

trungda:
> Sorry for the very long delay! This looks good to me.

Thank you @sfackler ! Happy to contribute more to the project!

PR #946, opened 2022-08-26
WIP add backend messages serialization and frontend messages deserialization

prx0:
The goal is to allow serialization of backend message into bytes and deserialization of frontend message into rust object that we can handle.
It could be useful for making a proxy or a fake postgresql server. (honeypot)

I've still lot of work here. My approach for backend message serialization is to implement TryFrom for Message in order to convert them into bytes.
I'm open to suggestions if you know a better way to do this of course.

Thank you!

### DIFF
--- a/postgres-protocol/src/message/backend.rs
+++ b/postgres-protocol/src/message/backend.rs
@@ -1,10 +1,11 @@
 #![allow(missing_docs)]
 
 use byteorder::{BigEndian, ByteOrder, ReadBytesExt};
-use bytes::{Bytes, BytesMut};
+use bytes::{Bytes, BytesMut, BufMut, Buf};
 use fallible_iterator::FallibleIterator;
 use memchr::memchr;
 use std::cmp;
+use std::convert::TryInto;
 use std::io::{self, Read};
 use std::ops::Range;
 use std::str;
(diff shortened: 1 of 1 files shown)


zain-kabani:
Hi Max, I actually wanted to implement the same thing as I was inspired by https://github.com/jackc/pgproto3. You can get a lot of pointers for parsing the messages there

prx0:
> Hi Max, I actually wanted to implement the same thing as I was inspired by https://github.com/jackc/pgproto3. You can get a lot of pointers for parsing the messages there

Hello!

Thank you. I use already pgproto3 in my company but I still want to continue this PR because I would like to work with Rust.

I'll check out the source code of pgproto3 to get some help.

Backend serialization is quite easy. But frontend deserialization is not.

I need some motivations and time, but I promess I'll finish this PR. ^^'

zain-kabani:
Hi I've been working on a project to make creating postgres proxies easier in rust. I've just gotten started on it and welcome any contributions to it! https://github.com/zain-kabani/postgres-proto-rs

prx0:
@zain-kabani that's great! I noticed that you've started to implement deserialization for frontend messages. That's what I try to do here. (also be careful you didn't specify any license in your project)

I don't know if it's a good idea to create another crate to add postgres protocol parsing because the rust-postgres crate made by @sfackler do this already. rust-postgres is a dependency of the postgres crate which is massively used. It would be great to add deserialization and serialization into rust-postgres in order to add functionnalities for existing projects based on this crate.

I think it would be awesome to create a framework for making progresql proxies with rust-postgres as dependency. pgcat another inspiring project. 
https://github.com/levkk/pgcat
Not a framework but a tool to manage sharding avec load balancing for postgres databases.

zain-kabani:
Hi Max, yeah I considered adding to this repo too but didn't want to rock the boat too much here and decided to just create a separate project.
Fun fact I'm actually an active contributor to the pgcat project, and built this with pgcat in mind.

Yes those are there temporarily as I still have to implement the logic of encoding/decoding them. (contributions are welcome!)

PR #951, opened 2022-09-21
Futures Sync

Dzordzu:
* Added `Sync` trait requirements
* Implemented `Sync` traits for some structs

Changes allow to use [async_trait_with_sync](https://docs.rs/async-trait-with-sync/latest/async_trait_with_sync/) (allowed returning `Client` and `JoinHandle`)

### DIFF
--- a/postgres-protocol/Cargo.toml
+++ b/postgres-protocol/Cargo.toml
@@ -19,3 +19,6 @@ memchr = "2.0"
 rand = "0.8"
 sha2 = "0.10"
 stringprep = "0.1"
+
+[features]
+sync-only = []
--- a/postgres-protocol/Cargo.toml
+++ b/postgres-protocol/Cargo.toml
@@ -19,6 +19,3 @@ memchr = "2.0"
 rand = "0.8"
 sha2 = "0.10"
 stringprep = "0.1"
-
-[features]
-sync-only = []
--- a/postgres-protocol/src/lib.rs
+++ b/postgres-protocol/src/lib.rs
@@ -12,6 +12,8 @@
 #![doc(html_root_url = "https://docs.rs/postgres-protocol/0.6")]
 #![warn(missing_docs, rust_2018_idioms, clippy::all)]
 
+#![feature(trait_alias)]
+
 use byteorder::{BigEndian, ByteOrder};
 use bytes::{BufMut, BytesMut};
 use std::io;
--- a/postgres-protocol/src/lib.rs
+++ b/postgres-protocol/src/lib.rs
@@ -12,8 +12,6 @@
 #![doc(html_root_url = "https://docs.rs/postgres-protocol/0.6")]
 #![warn(missing_docs, rust_2018_idioms, clippy::all)]
 
-#![feature(trait_alias)]
-
 use byteorder::{BigEndian, ByteOrder};
 use bytes::{BufMut, BytesMut};
 use std::io;
--- a/postgres-protocol/src/message/frontend.rs
+++ b/postgres-protocol/src/message/frontend.rs
@@ -132,14 +132,19 @@ pub fn close(variant: u8, name: &str, buf: &mut BytesMut) -> io::Result<()> {
     })
 }
 
-pub struct CopyData<T> {
+pub struct CopyData<T> 
+where T: Sync
+{
     buf: T,
     len: i32,
 }
 
+
+unsafe impl<T> Sync for CopyData<T> where T: Sync {}
+
 impl<T> CopyData<T>
 where
-    T: Buf,
+    T: Buf + Sync,
 {
     pub fn new(buf: T) -> io::Result<CopyData<T>> {
         let len = buf
--- a/postgres-protocol/src/message/frontend.rs
+++ b/postgres-protocol/src/message/frontend.rs
@@ -132,33 +132,26 @@ pub fn close(variant: u8, name: &str, buf: &mut BytesMut) -> io::Result<()> {
     })
 }
 
-#[cfg(feature = "sync-only")]
 pub struct CopyData<T> 
+{
+    buf: T,
+    len: i32,
+}
+
+
+/// Works the same as CopyData, but forces buffer to be `Sync`
+pub struct CopyDataSync<T> 
 where T: Sync
 {
     buf: T,
     len: i32,
 }
 
-#[cfg(not(feature = "sync-only"))]
-pub struct CopyData<T> 
-{
-    buf: T,
-    len: i32,
-}
 
-
-#[cfg(feature = "sync-only")]
-pub trait CopyDataT = Buf + Sync;
-
-#[cfg(not(feature = "sync-only"))]
-pub trait CopyDataT = Buf;
-
-impl<T> CopyData<T>
-where T: CopyDataT,
-{
-    pub fn new(buf: T) -> io::Result<CopyData<T>> {
-        let len = buf
+macro_rules! CopyDataNew {
+    ($buf: expr, $type: ident) => {
+        {
+        let len = $buf
             .remaining()
             .checked_add(4)
             .and_then(|l| i32::try_from(l).ok())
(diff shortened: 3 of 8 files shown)


postgres-protocol/src/message/frontend.rs
```diff
@@ -132,14 +132,19 @@ pub fn close(variant: u8, name: &str, buf: &mut BytesMut) -> io::Result<()> {
...
-pub struct CopyData<T> {
+pub struct CopyData<T> 
+where T: Sync
+{
     buf: T,
     len: i32,
 }
 
+
+unsafe impl<T> Sync for CopyData<T> where T: Sync {}
```
sfackler, line 143: This is exactly equivalent to the automatic implementation of Sync for CopyData.

postgres-protocol/src/message/frontend.rs
```diff
@@ -132,14 +132,19 @@ pub fn close(variant: u8, name: &str, buf: &mut BytesMut) -> io::Result<()> {
     })
 }
 
-pub struct CopyData<T> {
+pub struct CopyData<T> 
+where T: Sync
```
sfackler, line 136: Why do we want to prohibit the use of non-sync Ts here?

tokio-postgres/src/codec.rs
```diff
@@ -37,6 +37,8 @@ impl FallibleIterator for BackendMessages {
 
 pub struct PostgresCodec;
 
+unsafe impl Sync for PostgresCodec {}
```
sfackler, line 40: PostgresCodec is already Sync.

sfackler:
What is the rationale for needing Sync futures?

postgres-protocol/src/message/frontend.rs
Dzordzu: True, will be fixed

tokio-postgres/src/copy_in.rs
```diff
@@ -138,7 +138,7 @@ where
 
 impl<T> Sink<T> for CopyInSink<T>
 where
-    T: Buf + 'static + Send,
+    T: Buf + 'static + Send + Sync,
```
sfackler, line 141: This is a breaking change.

tokio-postgres/src/copy_in.rs
```diff
@@ -86,7 +86,7 @@ pin_project! {
 
 impl<T> CopyInSink<T>
 where
-    T: Buf + 'static + Send,
+    T: Buf + 'static + Send + Sync,
```
sfackler, line 89: This is a breaking change.

Dzordzu:
> What is the rationale for needing Sync futures?

I'm writing a library, where I want to pass an async function as the argument to another async function. After a little struggle I've managed to get 

```rust
    async fn handle_simple(
        &mut self,
        handler: &'static dyn (Fn(String) -> HandlerResult),
    ) 
```

That could not work, because of the 
```
handler has type `&dyn Fn(std::string::String) -> HandlerResult` which is not `Send`, because `dyn Fn(std::string::String) -> HandlerResult` is not `Sync``
```

So the solution was to make Fn sync. 

```rust
    async fn handle_simple(
        &mut self,
        handler: &'static (dyn (Fn(String) -> HandlerResult) + Send + Sync),
    ) {
```

And, as I was using tokio_postgres, I've encountered `get_type_rec` and  `prepare_rec`. After a quick fix (creating this breaking changes), I've managed to use `tokio-postgres` from within passed function.

**TLDR; When one is using tokio-postgres from async function passed to an async method**

> This is a breaking change.

Yep. It's just a simple (working) concept of what can it look like. I was wondering, if there can be flag, that will simply add `Sync` type restrictions. Unfortunately, type aliases are unstable feature, so it may require some boilerplate

postgres-protocol/Cargo.toml
```diff
@@ -19,3 +19,6 @@ memchr = "2.0"
 rand = "0.8"
 sha2 = "0.10"
 stringprep = "0.1"
+
+[features]
+sync-only = []
```
sfackler, line 24: This is still a back compat break. Any crate that enables this will affect all other consumers of postgres in the crate graph.

postgres-protocol/src/lib.rs
```diff
@@ -12,6 +12,8 @@
 #![doc(html_root_url = "https://docs.rs/postgres-protocol/0.6")]
 #![warn(missing_docs, rust_2018_idioms, clippy::all)]
 
+#![feature(trait_alias)]
```
sfackler, line 15: This crate must build on stable.

sfackler:
What type is `HandlerResult`? as far as I can tell, that is a blocking function passed into an asynchronous method.

You may want to use https://docs.rs/sync_wrapper/latest/sync_wrapper/struct.SyncWrapper.html to wrap the futures in something that implements Sync in your own code.

Dzordzu:
> What type is HandlerResult?

Oh. I forgot to include it's declaration. It's a future

```rust
HandlerResult: Future<Output = (String,i8)> + Sync + Send
```

> You may want to use https://docs.rs/sync_wrapper/latest/sync_wrapper/struct.SyncWrapper.html to wrap the futures in something that implements Sync in your own code.

I was already struggling with it, but I could not manage to use it in my case

postgres-protocol/Cargo.toml
Dzordzu: This is why it's an optional feature. It is intended to be not fully compatible with other versions and has to explicitly defined. 

I cannot see other way, except creating a structures variants and then selecting them via call to the proper method (like `execute`, `execute_sync`). 

The use case for it is also very specific - you need to pass connection with a sync requirement

sfackler: Features are global across all consumers of the library in a crate graph.

Dzordzu: I understand your concerns. I will adjust to your will :)

Dzordzu:
Ok. I've got an idea. As of now, the changes for async require only a little replace in 
* `tokio-postgres/src/prepare.rs` (`Send` + `Sync`)
* `tokio-postgres/src/codec.rs`  (`CopyData` and `Send` + `Sync`)
* `tokio-postgres/src/copy_in.rs` (`CopyData` and `Send` + `Sync`)
* `tokio-postgres/src/copy_out.rs` (`CopyData`)

As it was previously said, the best option would be to make this feature a non breaking change. The easiest way of doing it may be using a codegen to copy `tokio-postgres`. Then in the selected (ex. via preceding comment) lines `Send` with `Send + Sync` and `CopyData` with `CopyDataSync` should be replaced . `CopyDataSync` is a new struct with impl introduced with my previous commit.

This will allow to create a separate crate, let's name `tokio-postgres-sync`. The maintenance  of this solution will be minimal, as everything will be synced with the proper `tokio-postgres` crate. 

@sfackler what do you think? Is this acceptable?

sfackler:
That is a thing you could make, but I am not particularly interested in maintaining. Creating copies of every async library you want to use with `-sync` added to the end does not seem like a particularly sustainable approach, and IMO indicates that the need for `Sync` futures is something to avoid.

Dzordzu:
Yey! I've finally managed to avoid unnecessary code duplication. The hack was (simply) to use 

```rust
#[async_trait]
pub trait SimpleHandling<HandlerResult>
where
    HandlerResult: Future<Output = (String,i8)> + Send
{
    async fn handle_simple(&mut self, handler: impl (Fn(String) -> HandlerResult) + Send + Sync);
}
```

That allowed to use non `Sync` futures. @sfackler sorry, for wasting your time. In the future I'll try to help with real contribution there. 

Also, I'm very grateful for instructions, discussion and code reviews. Thanks!

IoIxD:
This solution is horrible (especially the one the repo's owner linked) and actually led me to have to write unsafe code to do a bitwise copy so that i could use SyncWrapper's `.into_inner()` without rust's borrow checker complaining.

If anybody happens to stumble upon this and is running into a conflict such as this, you should know that [sqlx](https://github.com/launchbadge/sqlx) is an alternative and happens to implement Sync. This is not to downplay the library owner's work (especially since this library has its own upsides), but its certainly a better solution then using SyncWrapper.

Issue #934, opened 2022-08-12
Feature request: `rename_all` container attribute


allan2:
A Postgres type like so
```sql
CREATE TYPE mood AS ENUM ('sad', 'happy');
```

can be represented like this
```rust
#[derive(ToSql, Debug)]
#[postgres(name = "mood")]
enum Mood {
    #[postgres(name = "sad")]
    Sad,
    #[postgres(name = "happy")]
    Happy
}
```

This is a request for a container attribute like Serde's `rename_all` <sup>[[1]](https://serde.rs/container-attrs.html)

```rust
#[derive(ToSql, Debug)]
#[postgres(rename_all = "lowercase")]
enum Mood {
    Sad,
    Happy
}
```

What do you think?

--
#894 related

sfackler:
Seems like a reasonable feature to me!

jakubadamw:
@allan2 @sfackler I made an attempt at an initial implementation in #952. Feedback would be welcome!

allan2:
Closing because merged in #1008 and released in postgres-types 0.2.6. Thanks @JaydenElliott @sfackler!

PR #952, opened 2022-09-24
Add a container-level `rename_all` attribute to `FromSql`/`ToSql`

jakubadamw:
It works both with composite types' fields and enum variants, although arguably it will be the most useful for the latter as that's where the naming conventions between Rust and a typical PostgreSQL schema differ.

Right now it supports the following casing styles, following [serde](https://serde.rs/container-attrs.html#rename_all):

* `"camelCase"`
* `"kebab-case"`
* `"lowercase"`
* `"PascalCase"`
* `"SCREAMING-KEBAB-CASE"`
* `"SCREAMING_SNAKE_CASE"`
* `"snake_case"`
* `"UPPERCASE"`

Addresses #934.


jakubadamw:
So this adds a few dependencies to keep the implementation concise. I, personally, subscribe to the “use as much of the crates.io ecosystem as you can possibly can” philosophy, but I can see how it may not be appropriate for the project. Still, that's what I did in the first version to gather feedback. Which dependency should be got rid of can perhaps be decided later.

allan2:
Since this is a library, I think it would be better to avoid introducing external dependencies to keep compile times down.

@sfackler thoughts?

The [case conversion code in serde_derive](https://github.com/serde-rs/serde/blob/master/serde_derive/src/internals/case.rs) might be helpful.

jakubadamw:
@JaydenElliott hey, not really, no. I'll be happy for someone else to take over this. 🙂

PR #955, opened 2022-10-06
Docs: make sure it's hard to miss that type name has to match

vargad:
Some people (me) doesn't read the full page or just gloss over the naming section. It's a lot of wasted time debugging the issue when the type name is not matching, adding name attribute to the examples makes it really hard to miss.


sfackler:
I think it would be better to actually use the attribute rather than to have it there but commented out.

Issue #882, opened 2022-04-07
Text protocol support ? What are the downsides?


Weakky:
## Context

> (If you wanna get straight to the point, you can skip this context and read the second part)

Hey, so we've been using `tokio-postgres` at [Prisma](https://www.prisma.io) for quite a while now. Very briefly, Prisma is a JavaScript ORM-like with a backend in rust. We use rust for various reasons that aren't relevant to my question here.

The ability for our users to send "raw" queries is an essential escape hatch. Unfortunately, we're seeing quite a lot of our users confused as to why they're unable to query a bunch of PG types that Prisma doesn't officially support, even with raw queries. eg: PostGIS types. They're especially confused because 99.9% of JS ORMs are based on the `node-postgres` driver, which uses the text protocol under the hood and so these queries fetching custom types used to work for them.

The underlying reason, obviously, is that we don't have ToSql/FromSql traits implemented for those types.

This leaves us with two options:
1. Suggest them to `TEXT` cast those types if they wish to query them
    - 🟡 That's an ok workaround but that prevents them from `SELECT *`.
2. Gradually support more types
    - 🔴 We need those escape hatch to work reliably without us intervening, otherwise we'd just officially support those types.

It might be possible to just forward those bytes to the Javascript side and enable users to parse them themselves, but the binary representation makes it much harder for common users. Worse, it seems like the best existing documentation as to how complex types should be deserialized is by reading the PG codebase, which is not amazing from a DX point of view when you're just willing to execute a query.

This brings me to my question 👇🏻 

## Text protocol, what are the downsides?

We've recently investigated the PG protocol in the hope of finding downsides to the text protocol, but we haven't really. The extended query flow documentation says types can individually be requested either as string or as binary ([and this discussion too](https://github.com/sfackler/rust-postgres/issues/21#issuecomment-27279280)).

Now, as the title asks, what are the downsides of using the text protocol, and why does `rust-postgres` prefer using the binary protocol? Are there common PG types that lack a text representation? Does it lead to type ambiguities in some cases? Does it have performance implications?

If there are no clear downsides though, is there any chance you'd be ok having `rust-postgres` support that protocol?

Thanks for your time, cheers

sfackler:
The binary protocol is significantly easier to work with if you want to parse out Rust types, which is what the library is primarily designed around.

The simple_query interface will return strings, but it doesn't support query parameters which may not be sufficient for what you're looking for.

[...]

PR #961, opened 2022-10-31
Add text format result support 

ZENOTME:
## What this PR do 
This PR add the TEXT format support for query result. In addition, this modification is compatitable for previous version. User needn't to change the original query.

## Context 
The rust-postgres only support binary format query result in extended query mode before. But in some cases, binary format of some type(such as multidimensional list) isn't support well. Sometimes, TEXT format result is more convenient. Such as in [sqllogictest](https://github.com/risinglightdb/sqllogictest-rs)(It is a testing framework to verify the correctness of an SQL database), it can test a database using extended query protocol. In this case, TEXT format is more convenient to compare the actual result with the expect result. 

I try to support text format in this PR. To make it compatible with the previous version, I add a result_format flag in client. The query will send message according to the result_format. And add the `get_text ` in row to get the result as TEXT format.

## Some problem may need to solved 
- Row can't guarantee the `get_text` will caused panic or error if user use it in a 'binary format row'. I only add the comment to indicate that user must guarantee use `get_text` in 'text format row'.
- I add the test case but I'm not sure whether enough.

Anyway, I'm glad to make any modification and hope it can be merged into main version. 

related issue: #882

### DIFF
--- a/postgres/src/client.rs
+++ b/postgres/src/client.rs
@@ -46,6 +46,24 @@ impl Client {
         Config::new()
     }
 
+    /// Return the result format of client
+    ///
+    /// true indicates that the client will receive the result in binary format
+    /// false indicates that the client will receive the result in text format
+    pub fn result_format(&self) -> bool {
+        self.client.result_format()
+    }
+
+    /// Set the format of return result.
+    ///
+    /// format
+    ///    true: binary format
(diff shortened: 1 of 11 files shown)


tobyhede:
+1 from me - our use case needs to support the text format.

Issue #421, opened 2019-02-25
Add support for rustls


alexduf:
I've successfully used this library to connect to an AWS RDS postgres instance, and it's fantastic and straightforward, thanks for providing such a nice library!

However I haven't managed to encrypt the connection with TLS yet. It seems like Rustls is the ideal candidate to be embedded in an AWS lambda and is already supported by rusoto, an aws api binding in Rust.

I have found one implementation (untested) by @jbg here: https://github.com/jbg/rust-postgres-rustls/blob/master/src/lib.rs

Being a beginner in Rust I find it hard to understand how to bring a TLS implementation such that the postgres connection is secure, and I think the documentation could benefit from being a bit more verbose on the topic.

Particularly, an example for each of the binding provided (openssl, schannel etc.) with their respective cargo.toml and crates used, with a corresponding snippet. Pretty much like the native-tls example that is provided, including the cargo.toml so that we can see without a doubt how to pass the feature flag.

Again thanks for the library and time spend developing it, I hope a rustls binding would be helpful for everyone.

jbg:
I didn't go any further with the implementation linked above because one of the servers I wanted to connect to had a TLS certificate with a CN that wasn't a valid DNS name, and the other used an IP address subjectAltName. Neither of these are supported by webpki, which rustls uses, and I didn't have time to dive into implementing support at the time. I might circle back around to it, but for now the linked code should probably be considered "untested but possibly a usable basis for a proper implementation".

You might want to check what kind of certificate AWS serves up in case it also has some detail that's not supported by webpki.

dhbradshaw:
rustls support would be really nice.  I tried to connect aws lambda to heroku and found compilation too painful with open and native ssl.  I ended up using diesel instead, which was overkill for my use case but had working ssl because of rustls.

sfackler:
Out of curiosity, how did you find using diesel (which links to libpq in addition to OpenSSL) easier to build than using either openssl or native-tls (which uses openssl)?

dhbradshaw:
Looking back at the code, you're right -- I had mis-remembered.

The original intent was to get a lambda function working.  The function had to hit an api and then hit a database.

I worked on the first part of the job first, which was getting reqwest working.  To do that I finally found that I could skip openssl problems by using rustls .

Sadly, I then found out that I had the same issue (ssl compilation issues) connecting to postgres.

[...]

PR #962, opened 2022-11-01
Merge whole history of tokio-postgres-rustls as a crate

nyurik:
This code does not have any significant changes from the original (archived) code at https://github.com/jbg/tokio-postgres-rustls

Changes:
* move all files to the `/postgres-rustls` dir
* bump `env_logger` dev dependency to 0.9
* add new crate as a member to the main Cargo.toml

TODO:
* either get publishing rights to [tokio-postgres-rustls](https://crates.io/crates/tokio-postgres-rustls) at crates.io, or rename the package. Probably should be done as a separate PR? CC: @jbg

See also #421

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
     "postgres-native-tls",
     "postgres-openssl",
     "postgres-protocol",
+    "postgres-rustls",
     "postgres-types",
     "tokio-postgres",
 ]
--- a/postgres-rustls/.gitignore
+++ b/postgres-rustls/.gitignore
@@ -1,0 +1,3 @@
+/target
+**/*.rs.bk
+Cargo.lock
--- a/postgres-rustls/Cargo.toml
+++ b/postgres-rustls/Cargo.toml
@@ -17,7 +17,6 @@ tokio-postgres = "0.7"
 tokio-rustls = "0.23"
 
 [dev-dependencies]
-env_logger = { version = "0.8", default-features = false }
+env_logger = { version = "0.9", default-features = false }
 tokio = { version = "1", features = ["macros", "rt"] }
 rustls = { version = "0.20", features = ["dangerous_configuration"] }
-
--- a/postgres-rustls/Cargo.toml
+++ b/postgres-rustls/Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "tokio-postgres-rustls"
+description = "Rustls integration for tokio-postgres"
+version = "0.9.0"
+authors = ["Jasper Hugo <jasper@jasperhugo.com>"]
+repository = "https://github.com/jbg/tokio-postgres-rustls"
+edition = "2018"
+license = "MIT"
+readme = "README.md"
+
+[dependencies]
+futures = "0.3"
+ring = "0.16"
+rustls = "0.20"
+tokio = "1"
+tokio-postgres = "0.7"
+tokio-rustls = "0.23"
+
+[dev-dependencies]
+env_logger = { version = "0.8", default-features = false }
+tokio = { version = "1", features = ["macros", "rt"] }
+rustls = { version = "0.20", features = ["dangerous_configuration"] }
+
--- a/postgres-rustls/LICENSE
+++ b/postgres-rustls/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 Jasper Hugo
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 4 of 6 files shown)


jbg:
FYI that repo is only archived because it moved to SourceHut. The crate is still developed and maintained. That being said, I'm happy if it can come under the rust-postgres umbrella.

nyurik:
@jbg thank you so much!  I wasn't sure what has happened to it, and some of us got really concerned when we saw that your repo was archived without any notes, while crates.io still pointed to it. Could you maybe unarchive it, put some big notification at the top with the links to the new location, and re-archive it?

I am not familiar with the sourcehut, and I tried to [find it there](https://sr.ht/projects?search=rustls&sort=recently-updated), but it didn't come up with any results.

I am all for collaborative software development -- we tend to support software until our interest runs out, so ideally there should be a community behind each project rather than a single person - to prevent any kind of disaster or lack of interest from ruining an amazing project.  I am not sure what @sfackler 's plans are for this project - I really hope we can all work on it together, perhaps starting an organization too?  Regardless, you have done a great job maintaining it, so I wouldn't want to take it away from you, but please consider keeping it on github so that all other FOSS devs can easily participate.

Thanks!

jbg:
There is a link at the top of the Github repo (at the right on desktop) to the Sourcehut repo.

I may just move back to Github given that hosting elsewhere causes this confusion 😬

sfackler:
Sorry for the delay - if @jbg's happy maintaining it externally then I think it'd probably be best to keep things as-is.

nyurik:
I'm ok either way, but one thing to note: we urgently need to implement all the tls compatibility with rustls - the upstream team seems to [not want](https://github.com/rustls/rustls/pull/1197) to have an easy path for us to upgrade from openssl to rustls, and thus we would have to do a considerable implementation of everything related to [`sqlmode=...`](https://github.com/rustls/rustls/pull/1197#discussion_r1109170002).

I think it would be best to implement this type of functionality in this project for higher visibility -- keeping postgresql-related things together is easier than trying to hunt for individual pieces.  Of course if @jbg is ok with this transition?

jbg:
@nyurik Can you clarify what functionality you are talking about when you say "all the tls compatibility with rustls"? Any kind of verification (or lack thereof) that you can implement with `rustls`, you can use `tokio-postgres-rustls` with. There is no need for any special support to be implemented in `tokio-postgres-rustls` or `tokio-postgres`, you just pass the appropriate `ClientConfig` to `MakeRustlsConnect::new()`. If you're having a specific problem, please open a related issue.

jbg:
Ah, I see you already have. Linking to https://github.com/jbg/tokio-postgres-rustls/issues/11 here for context and will continue the discussion there.

serprex:
Given
* https://github.com/jbg/tokio-postgres-rustls/pull/14
* https://github.com/jbg/tokio-postgres-rustls/pull/15
* https://github.com/jbg/tokio-postgres-rustls/pull/16
* https://github.com/jbg/tokio-postgres-rustls/pull/17
it may be time to revive this PR. It's maybe too soon now, but would be good to decide when it wouldn't be too soon. I reached out to @jbg a week ago asking if he'd like help maintaining tokio-postgres-rustls

vasilakisfil:
I really think that `tokio-postgres-rustls` should be part of this repo, even if not maintained by @sfackler.

jbg:
I've made a `tokio-postgres-rustls` release (apologies for the delay -- was moving countries!).

paolobarbolini:
Given rustls' popularity, we should have a well maintained integration with rust-postgres. Reiterating https://github.com/rust-postgres/rust-postgres/pull/962#issuecomment-1477114699, if @jbg is happy to maintain it then it's ok for us too. There are 3 pending PRs right now that make it look like it may not be getting much attention though.

As an alternative we could either take on maintenance, and in that case this PR would need to be updated to include all of the latest commits [^1], or the [rustls org](https://github.com/rustls) could.

[^1]: looking at the docs, [`git-filter-repo`](https://github.com/newren/git-filter-repo) sounds like a better tool for doing so, instead of the current merge commit + move everything approach.

kamulos:
`tokio-postgres-rustls` definitely seems to be unmaintained or at the very least maintained at a very slow pace. I just found the `tokio-postgres-rustls-improved` crate, which tries to solve the issue via a fork, but I thnk it would be much better to maintain it alongside this crate. Really hope the licensing difference does not block that

paolobarbolini:
The license is not a problem, worse case scenario we keep it into a separate repo instead of this one. `tokio-postgres-rustls-improved` indeed seems the better client :+1:.

PR #963, opened 2022-11-21
Add idle session timeout error

jkosh44:
This commit adds the idle session timeout error (57P05). See https://www.postgresql.org/docs/current/errcodes-appendix.html


tokio-postgres/src/error/sqlstate.rs
```diff
@@ -1498,6 +1503,7 @@ static SQLSTATE_MAP: phf::Map<&'static str, SqlState> =
         ("55006", SqlState::OBJECT_IN_USE),
         ("42P01", SqlState::UNDEFINED_TABLE),
         ("25P03", SqlState::IDLE_IN_TRANSACTION_SESSION_TIMEOUT),
+        ("57P05", SqlState::IDLE_SESSION_TIMEOUT),
```
jkosh44, line 1506: It didn't seem like there was any specific order here, so I just put this next to a similar timeout error. Let me know if I missed something.

sfackler: This file is generated by the `codegen` crate in this repo. The proper way to update it is to update errcodes.txt, pg_range.dat, and pg_type.dat and `cargo run` the codegen.

jkosh44: Oops, totally missed the comment at the top. I'll do that now.

jkosh44: It looks like there have been quite a few changes since this was last updated. Specifically multi-ranges causes the type gen code to panic becuase it thinks the oid is for than normal range type. I can take a look and see how we can update the code for multi-ranges. Let me know if you have any thoughts/ideas.

jkosh44:
I copied errcodes.txt, pg_range.dat, and pg_type.dat from the 51b5834cd53f0bd068729043b55f7da3ca6bb15f commit of PostgreSQL.

codegen/src/type_gen.rs
```diff
@@ -209,7 +218,11 @@ fn parse_types() -> BTreeMap<u32, Type> {
         }
 
         let element = match &*kind {
-            "R" => range_elements[&oid],
+            "R" => match &*raw_type["typtype"] {
+                "r" => range_elements[&oid],
+                "m" => multi_range_elements[&oid],
```
jkosh44, line 223: You can find the description of this field in the `pg_type` catalog table (https://www.postgresql.org/docs/current/catalog-pg-type.html). The relevant part is:
> r for a range type, or m for a multirange type.

sfackler:
It looks like multirange types are currently mapping to `Kind::Range`, but it would probably make more sense to create a separate `Kind::Multirange`.

You'll also need to run `cargo fmt` on the project after the codegen'd files are recreated.

jkosh44:
> It looks like multirange types are currently mapping to `Kind::Range`, but it would probably make more sense to create a separate `Kind::Multirange`.
> 
> You'll also need to run `cargo fmt` on the project after the codegen'd files are recreated.

Just did both. Let me know if you have notes on the approach.

jkosh44:
@sfackler The test and clippy failures look like they're in parts of the code that I haven't touched. Any ideas what's going on?

sfackler:
Those should be fixed on master now if you merge/rebase.

sfackler:
LGTM otherwise!

jkosh44:
> Those should be fixed on master now if you merge/rebase.

Just rebased from master

sfackler:
Thanks!

PR #972, opened 2022-12-24
Change Example Code Snippet

candywater:
I got confused as I just copy&paste code from example and it does not work.
as I'm new to rust, and code from example does not have a main method, 
I'm considering if rust does not need a main method nowadays (just like c# 10 or something else)
yet the answer is not.
main method just be commented out.
it confused me and take me for several hours to figure it out.

so I think we should make example more useful.


sfackler:
The main function is traditionally hidden in code examples in Rust documentation.

candywater:
got it, thanks. @sfackler

Issue #950, opened 2022-09-13
Expose SimpleQueryRow internals to user


dfwilbanks395:
I have a use case where I'd like to reconstruct a Simple Query message and pass it back to my user. To do this I'd want to make a few changes to the `Client::simple_query()` implementation.

1. Store and return the entire `Field` that is built from a backend `Message::RowDescription` in `SimpleQueryRow` instead of just returning the column name in `SimpleColumn`. I've done this with minor changes to the current interface in [our rust-postgres fork](https://github.com/sfackler/rust-postgres/commit/01fa1b828a130388c48d13b127439f8487fa01da).

2. Ideally, expose `SimpleQueryRow::body` to avoid having to read each value using `SimpleQueryRow::get()` and re-serialize the message to bytes myself.

3. Ideally, create public constructors for `SimpleQueryRow`, `Field`, and `DataRowBody` to allow easier unit testing in my own code. I noticed that a similar question came up [in a previous issue](https://github.com/sfackler/rust-postgres/issues/910). Maybe we could implement this under a compile time feature flag?

Issue #910, opened 2022-06-23
Create `Row` for unit tests.


BratSinot:
Greetings!

Any way to create `Row` for unit-tests?

sfackler:
The only way to create a Row is currently to query it from a Postgres instance.

PR #979, opened 2023-01-05
Introduce RowBuilder to support writing basic unit tests

hunts:
Added RowDescription trait, and let rows to share the same description rather than having a copy in each row (think when there are thousand of them in the result).

Added RowBuilder to support adding stubs of row data in unit tests.

Currently, the library users have no chooice but have to use integration tests for testing Postgres data access code. With the changes in this commit, the `tokio-postgres` lib users can use RowBuilder to create sutbs to verify the deserialization from database result (Rows) to custom stucts in unit tests.

It can also serves as a base for future implementation of certain kind of mocks of the db connection.

Related-to #910 #950


phyllipecesar:
Hi @sfackler , 

I usually create a struct and implement `TryFrom<Row>` for that struct, I rely on integration tests(_which are usually more expensive_) to test end to end.

Do you have any recommendation how to unit test conversions?
Thank you.

deleted:
I, too, need something like this for unit testing.
@sfackler Any thoughts on accepting this or something similar?

owlot:
I also need this, been watching the pr for a while, but seems like it's not going to be merged at all?

deleted:
Nothing to add except that the ability to synthesize Rows / a RowStream in unit testing is still very important to me.

Is this something we can expect to ever see in `rust-postgres` or no?

owlot:
@sfackler can you let us know if you're considering this PR at all or if we have to build around it some other way to be able to get our unit test coverage up.

PR #980, opened 2023-01-08
tokio_postgres: add batch_execute to GenericClient

aDogCalledSpot:
`batch_execute` is in both `tokio_postgres::Client` and `tokio_postgres::Transaction` and is therefore present in the non-tokio `postgres::GenericClient` but is missing from `tokio_postgres::GenericClient`.

### DIFF
--- a/tokio-postgres/src/generic_client.rs
+++ b/tokio-postgres/src/generic_client.rs
@@ -72,7 +72,6 @@ pub trait GenericClient: private::Sealed {
     /// Like `Client::batch_execute`.
     async fn batch_execute(&self, query: &str) -> Result<(), Error>;
 
-
     /// Returns a reference to the underlying `Client`.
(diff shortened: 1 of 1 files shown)


tokio-postgres/src/generic_client.rs
```diff
@@ -69,6 +69,10 @@ pub trait GenericClient: private::Sealed {
     /// Like `Client::transaction`.
     async fn transaction(&mut self) -> Result<Transaction<'_>, Error>;
 
+    /// Like `Client::batch_execute`.
+    async fn batch_execute(&self, query: &str) -> Result<(), Error>;
+
+
```
sfackler, line 75:
```suggestion
```

sfackler:
Thanks! Not sure how we missed that method the first time...

PR #985, opened 2023-01-15
feat: add `rows_affected` to RowStream

skyzh:
Signed-off-by: Alex Chi <iskyzh@gmail.com>

This PR adds a new `RawStream` for query result in extended protocol. In some cases, users won't know if the query is a DML or DDL beforehand. `RawStream` returns an enum of `Row` or `RowsAffected`, so that users can always retrieve the information they need.

Previously, only `RowStream` or `query` can get the row contents, and `execute` can only get number of rows affected.

This is used in https://github.com/risinglightdb/sqllogictest-rs/pull/151, which correctly adds Postgres extended protocol support for our Rust version of sqllogictest.

This PR also eliminates duplicated code for extracting number of rows from CommandCompleted. Now we have a single function `extract_row_affected` for that.

Thanks for review! @sfackler

### DIFF
--- a/tokio-postgres/src/query.rs
+++ b/tokio-postgres/src/query.rs
@@ -7,7 +7,7 @@ use bytes::{Bytes, BytesMut};
 use futures_util::{ready, Stream};
 use log::{debug, log_enabled, Level};
 use pin_project_lite::pin_project;
-use postgres_protocol::message::backend::Message;
+use postgres_protocol::message::backend::{CommandCompleteBody, Message};
 use postgres_protocol::message::frontend;
 use std::fmt;
 use std::marker::PhantomPinned;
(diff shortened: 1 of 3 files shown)


skyzh:
Sounds great to me! I'll add a `row_affected: Option<u64>` inside `RowStream` and allow user to access it.

skyzh:
Updated with `rows_affected` inside `RowStream`.

tokio-postgres/src/query.rs
```diff
@@ -217,12 +226,22 @@ impl Stream for RowStream {
...
                 _ => return Poll::Ready(Some(Err(Error::unexpected_message()))),
             }
         }
     }
 }
+
+impl RowStream {
+    /// Returns the number of rows affected by the query.
+    ///
+    /// This will be `None` if the information is not available yet.
```
sfackler, line 243: This should be more specific - it won't be set until the stream has been exhausted.

skyzh: updated, thanks!

PR #999, opened 2023-02-23
Make incorrect number of paramaters an error

### LINKED ISSUE #998: Should query panic when the number of parameters doesn't match?
The `query` method and its siblings will panic when the number of parameters in the statement doesn't match the number of parameters passed into the method. This seems like a very strange design decission. The method already returns a generic opaque error type, so it should be simple to add a variant to it for this type of error. It's also inconsistent, since a completely malformed query will just return a "soft" error. It a valid query with an incorrent number of parameters worse that a completely invalid query? Furthermore, panics are usually reserved for programmer errors. Since the query gets parsed by the server, a misbehaving server could crash client code.

SvizelPritula:
Modifies the `query` method and its siblings to return a new error type instead of panicking when the number of provided parameters doesn't match the number of parameters in the statement.

Closes #998.

### DIFF
--- a/postgres/src/client.rs
+++ b/postgres/src/client.rs
@@ -57,10 +57,6 @@ impl Client {
     /// repeatedly executed (perhaps with different query parameters), consider preparing the statement up front
     /// with the `prepare` method.
     ///
-    /// # Panics
-    ///
-    /// Panics if the number of parameters provided does not match the number expected.
-    ///
     /// # Example
     ///
     /// ```no_run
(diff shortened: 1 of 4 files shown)


sfackler:
CI should be fixed if you rebase.

SvizelPritula:
I accidentally merged instead of rebasing. Hold on.

SvizelPritula:
Rebase completed, should be ready now.

sfackler:
Thanks!

PR #1000, opened 2023-02-24
export Bytes type through DataRowBody::storage_bytes method.

### LINKED ISSUE #996: Consider expose `Bytes` type trhough postgres-protocol?
For type like [DataRowBody](https://github.com/sfackler/rust-postgres/blob/4bae134f501db5c08b8dd43a44474e515bbb39d3/postgres-protocol/src/message/backend.rs#L528) that is backed by Bytes internally it would be beneficial to expose it directly for zero copy downcasting/slicing without concerning of explicit reference and lifetime.

For example:
```rust
pub struct DataRowBody {
    storage: Bytes,
}

impl DataRowBody {
    fn storage(&self) -> &(impl AsRef<[u8]> + Clone) {
        &self.storage
    }
}

// slice bytes with cheap clone
fn slice(row: DataRowBody) -> Bytes {
    let bytes = (row.storage() as &dyn Any).downcast_ref::<Bytes>().unwrap();
    bytes.slice(1..5)
}

// cheap clone to onwership.
fn cheap_clone(row: DataRowBody) -> impl AsRef<[u8]> + 'static {
    row.storage().clone()
}

```

fakeshadow:
Exposing `Bytes` type can be beneficial as https://github.com/sfackler/rust-postgres/issues/996 suggested.

Regarding high level usage it can be introduced as following pesudo api:
```rust
impl Row {
    fn get2<I, T>(&self, idx: &I) -> Result<T, Error>
    where
        I: RowIndex + fmt::Display,
        T: FromSqlOwned,
    {
        ...
    }
}

trait FromSqlOwned {
  fn from_sql(ty: &Type, raw: &Bytes) -> Result<Self, Box<dyn Error + Sync + Send>>;
}

// new type of Bytes with utf8 encoding check for cheap clone.
struct ByteString(Bytes);

impl FromSqlOwned for ByteString {
  fn from_sql(ty: &Type, raw: &Bytes) -> Result<Self, Box<dyn Error + Sync + Send>> {
    std::str::from_utf8(raw.as_ref())?
    ByteString(raw.clone())  
  }
}
```

Closes #996

### DIFF
--- a/postgres-protocol/src/message/backend.rs
+++ b/postgres-protocol/src/message/backend.rs
@@ -544,6 +544,11 @@ impl DataRowBody {
     pub fn buffer(&self) -> &[u8] {
         &self.storage
     }
+
+    #[inline]
+    pub fn storage_bytes(&self) -> &Bytes {
+        &self.storage
+    }
 }
 
 pub struct DataRowRanges<'a> {


fakeshadow:
Regarding changelog should I add an unreleased session on top?

sfackler:
Nope, no need.

ParkMyCar:
I see the [`fn buffer_bytes(...)`](https://github.com/sfackler/rust-postgres/blob/master/postgres-protocol/src/message/backend.rs#L549) method now exists on `DataRowBody`, has anything like the high level interface been implemented yet? Given a `Row` I don't believe you can access the `DataRowBody` given it's a private field and looking through the docs there don't appear to be any accessors.

My goal is to get zero-copy accessing of a `bytea` column, similar to what's mentioned in #996. More than happy to submit a PR if you'd be open to something like the high level usage proposed here @sfackler?

PR #1007, opened 2023-03-24
wire through knob for TCP user timeout

pH14:
TCP user timeouts are pretty useful, especially in conjunction with the existing connect timeout / keepalive settings. This adds a new config value, allowing us to set a user timeout on the socket similar to the keepalive setting.

### DIFF
--- a/tokio-postgres/Cargo.toml
+++ b/tokio-postgres/Cargo.toml
@@ -55,7 +55,7 @@ pin-project-lite = "0.2"
 phf = "0.11"
 postgres-protocol = { version = "0.6.4", path = "../postgres-protocol" }
 postgres-types = { version = "0.2.4", path = "../postgres-types" }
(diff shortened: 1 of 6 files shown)


sfackler:
Does this setting exists in psql? I would prefer to mirror its configuration options.

pH14:
It should! It's one of the libpq params: https://www.postgresql.org/docs/15/libpq-connect.html#LIBPQ-PARAMKEYWORDS

sfackler:
Oh nice, I totally missed that somehow...

sfackler:
Thanks!

pH14:
🙇 thank you @sfackler!

benesch:
> Oh nice, I totally missed that somehow...

(It's new in v12, so you might've been looking at an old version of the docs.)

PR #1008, opened 2023-03-26
added `rename_all` container attribute for enums and structs

JaydenElliott:
Related issue: #934 

Adding a `rename_all` attribute to a struct or enum will rename the container's fields/variants according to the rule specified. Similar to serde's `rename_all` but for use with `ToSql` and `FromSql`. 

This removes the need to manually rename each field/variant before inserting/selecting from postgres (especially useful with enums, who's naming standards differ from rust).

Supported rules:
- "lowercase"
- "UPPERCASE"
- "PascalCase"
- "camelCase"
- "snake_case"
- "SCREAMING_SNAKE_CASE"
- "kebab-case"
- "SCREAMING-KEBAB-CASE"

No external dependencies introduced. Case conversion code from [serde](https://github.com/serde-rs/serde/blob/master/serde_derive/src/internals/case.rs) was used.

Thanks @jakubadamw for the test cases from #952

### DIFF
--- a/postgres-derive/src/case.rs
+++ b/postgres-derive/src/case.rs
@@ -1,0 +1,158 @@
+#[allow(deprecated, unused_imports)]
+use std::ascii::AsciiExt;
+
+use self::RenameRule::*;
+
+/// The different possible ways to change case of fields in a struct, or variants in an enum.
+#[allow(clippy::enum_variant_names)]
+#[derive(Copy, Clone, PartialEq)]
+pub enum RenameRule {
+    /// Rename direct children to "lowercase" style.
+    LowerCase,
+    /// Rename direct children to "UPPERCASE" style.
+    UpperCase,
+    /// Rename direct children to "PascalCase" style, as typically used for
+    /// enum variants.
+    PascalCase,
+    /// Rename direct children to "camelCase" style.
+    CamelCase,
+    /// Rename direct children to "snake_case" style, as commonly used for
+    /// fields.
+    SnakeCase,
+    /// Rename direct children to "SCREAMING_SNAKE_CASE" style, as commonly
+    /// used for constants.
+    ScreamingSnakeCase,
+    /// Rename direct children to "kebab-case" style.
+    KebabCase,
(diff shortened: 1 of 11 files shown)


postgres-derive/src/case.rs
```diff
@@ -0,0 +1,158 @@
...
+    ("camelCase", CamelCase),
+    ("snake_case", SnakeCase),
+    ("SCREAMING_SNAKE_CASE", ScreamingSnakeCase),
+    ("kebab-case", KebabCase),
+    ("SCREAMING-KEBAB-CASE", ScreamingKebabCase),
+];
+
+impl RenameRule {
+    /// Apply a renaming rule to an enum variant, returning the version expected in the source.
+    pub fn apply_to_variant(&self, variant: &str) -> String {
```
sfackler, line 44: You should be able to use https://docs.rs/heck/latest/heck/ to handle all of this instead of doing it manually.

JaydenElliott: Nw, was trying to keep away from libs but this one is tiny anyway. That's done.

JaydenElliott: fyi, in the commit just pushed, I also added in an `Overrides::extra`  flag argument to give a clearer error msg when trying to use `rename_all` on a field instead of the container.

sfackler:
Can you also update the documentation here? https://github.com/sfackler/rust-postgres/blob/master/postgres-types/src/lib.rs

LGTM otherwise

JaydenElliott:
> Can you also update the documentation here? https://github.com/sfackler/rust-postgres/blob/master/postgres-types/src/lib.rs
> 
> LGTM otherwise

@sfackler  Documentation added, see [commit](https://github.com/sfackler/rust-postgres/pull/1008/commits/f4b181a20180f1853351be53a32865b6209d0ab4)

JaydenElliott:
can this be approved @sfackler ?

MindMayhem:
we waiting for u, @sfackler

sfackler:
Thanks! Sorry for the delay

Issue #1023, opened 2023-04-25
Question regarding enum backward compatibility


Leulz:
First of all, thank you for the work done on this library, it has been very useful!

One thing that I would like to understand is the fact that rust-postgres apparently expects the Rust enums to have all possible variants of a Postgres enum, as seen in [these tests](https://github.com/sfackler/rust-postgres/blob/fc10985f9fdf0903893109bc951fb5891539bf97/postgres-derive-test/src/enums.rs#L72-L104). Is that correct?

If so, then does it mean that there is no backward compatibility? If I run a migration that adds an enum value, for example, I have to always ensure that a new deployment that has that new variant in the Rust enum will be deployed ASAP?

I wanted the behavior to be that as long as I am not trying to serialize/deserialize the value of a new enum that does not exit on Rust's or Postgres's side, it would still work. Is it possible to have this behavior? Or is the implementation inherently not backward compatible?

sfackler:
Yes, the generated implementations currently expect an exact match of both the type name and variant names.

The string representation of the enum variant is what's sent over the wire, so there is no hard requirement that the Rust type is an exact match of the Postgres type, but the check is potentially nice in surfacing mismatches immediately rather than deferring the problem to the first time the bad variant is sent or received.

We could add an option to the code generation to disable the variant comparisons potentially.

PR #1026, opened 2023-04-30
feat(derive): add `#[postgres(allow_mismatch)]`

viniciusth:
As discussed on #1023, this PR implements an override/attribute that allows mismatching Rust and Postgres enums.

## Example

```rust
#[derive(ToSql, FromSql)]
#[postgres(allow_mismatch)]
enum Mood {
    Sad,
    Happy,
    Meh,
}
```

### DIFF
--- a/postgres-derive/src/accepts.rs
+++ b/postgres-derive/src/accepts.rs
@@ -31,31 +31,37 @@ pub fn domain_body(name: &str, field: &syn::Field) -> TokenStream {
     }
 }
 
-pub fn enum_body(name: &str, variants: &[Variant]) -> TokenStream {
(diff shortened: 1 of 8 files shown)


viniciusth:
Hey @sfackler could you please review this PR?

postgres-derive/src/accepts.rs
```diff
@@ -40,6 +40,10 @@ pub fn enum_body(name: &str, variants: &[Variant]) -> TokenStream {
             return false;
         }
 
+        if #allow_mismatch {
+            return true;
+        }
+
```
sfackler, line 46: I'd rather just not generate the match below at all if allow_mismatch is set.

viniciusth: true, fixed!

PR #1029, opened 2023-05-07
allow `BorrowToSql` for non-static `Box<dyn ToSql>`

laptou:
This allows me to use `tokio_postgres::binary_copy::BinaryCopyInWriter` with boxed `ToSql` types that are created by borrowing data from somewhere else (and are therefore not `'static`).


sfackler:
Thanks!

Issue #215, opened 2016-11-08
Feature Request: Optionally disable use of the log crate


pdtilton:
I am interested in using postgres as a logging backend in conjunction with the log crate.  The issue is that anytime such a logger creates a query it results in infinite recursion due to the use of the `debug!` and `info!` log macros in src/lib.rs.  Would it be possible to optionally disable this feature in rust-postgres? I'm fairly new to Rust, so perhaps there is a workaround I am not considering instead.

sfackler:
Haha, that's something I haven't thought of! Seems like a reasonable way to fix this would be to add a feature to disable logging.

pdtilton:
Wow!  That was fast.  I was just looking into approaches to tackle the issue.  Thank you very much, sir!

stevenliebregt:
It seems this no longer exists in the current version, is it possible to bring this back?

sfackler:
I think a pluggable trait would be the preferred way to go here, like what's discussed in #1053.

stevenliebregt:
Is there any progress in that PR, or could I propose a quick feature based solution in the meantime? I'm using this crate in an FFI context with Node, and with large queries it keeps crashing the processes because rust expects stdout to be blocking, and node doesn't. Simply setting the node stdout to blocking isn't a great solution as it might block the node event loop as well.

sfackler:
You could not log to stdout when running in an FFI context with Node.

PR #1053, opened 2023-07-24
Tracing support

Protryon:
This PR:
* Adds tracing support following [OpenTelemetry database semantic conventions](https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/database/)
* Adds no new dependencies (tracing was already added transitively by other dependencies)
* Will have no performance impact if no tracer or logger is setup at a `DEBUG` or `TRACE` level.

Example trace overview on real application (vaultwarden):

direct

Breaking down to find a slow query:

direct

Note that this PR does not add distributed tracing support into the database, as it is (AFAIK) experimental and not well supported -- it would also add some extra dependencies.

### DIFF
--- a/tokio-postgres/Cargo.toml
+++ b/tokio-postgres/Cargo.toml
@@ -58,6 +58,7 @@ postgres-protocol = { version = "0.6.5", path = "../postgres-protocol" }
 postgres-types = { version = "0.2.4", path = "../postgres-types" }
 tokio = { version = "1.27", features = ["io-util"] }
 tokio-util = { version = "0.7", features = ["codec"] }
+tracing = "0.1"
 rand = "0.8.5"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
--- a/tokio-postgres/src/client.rs
+++ b/tokio-postgres/src/client.rs
@@ -9,7 +9,7 @@ use crate::simple_query::SimpleQueryStream;
 #[cfg(feature = "runtime")]
 use crate::tls::MakeTlsConnect;
 use crate::tls::TlsConnect;
-use crate::trace::make_span;
+use crate::trace::{make_span, SpanOperation};
 use crate::types::{Oid, ToSql, Type};
 #[cfg(feature = "runtime")]
 use crate::Socket;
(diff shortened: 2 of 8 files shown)


sfackler:
What are some examples of other database drivers that have adopted these logging conventions?

titanous:
This works great, I ended up making two changes:
1. I set all of the spans to INFO because that is the level that we send to Honeycomb. I think at the very least prepare should be bumped to DEBUG, as trace is almost always too noisy to be useful and prepare spans are useful to me (I try to have a span for every external network call).
2. I added a span for connect, which delivered value as soon as I deployed it. Feel free to pull in this commit and/or modify it. I have already found the sub-span timings that I added to be useful as well. https://github.com/titanous/rust-postgres/commit/7e2e3a2210dedda7a0f70ca9069d3ca1f57901cb

Protryon:
> This works great, I ended up making two changes:
> 
> 1. I set all of the spans to INFO because that is the level that we send to Honeycomb. I think at the very least prepare should be bumped to DEBUG, as trace is almost always too noisy to be useful and prepare spans are useful to me (I try to have a span for every external network call).
> 2. I added a span for connect, which delivered value as soon as I deployed it. Feel free to pull in this commit and/or modify it. I have already found the sub-span timings that I added to be useful as well. [titanous@7e2e3a2](https://github.com/titanous/rust-postgres/commit/7e2e3a2210dedda7a0f70ca9069d3ca1f57901cb)

I merged in that commit and bumped all the trace levels as well. Thanks!

sfackler:
From what I can tell most or all of the libraries at that OTEL reference are external libraires that you can use with a database driver to apply tracing conventions.

I am hesitant to land this for a couple of reasons:

1. People can be using this library as a tracing appender, and this change can potentially cause some undesired infinite logging recursion. cf https://github.com/sfackler/rust-postgres/issues/215
2. This seems like a convention that your particular organization wants to use, e.g. "I set all of the spans to INFO because that is the level that we send to Honeycomb", not something that is uniformly applicable.

I think you'd probably be better served making a separate library that can wrap tokio-postgres and apply the appropriate tracing logic.

titanous:
I think a separate library would be fine, though it seems like a hook system would be necessary to ergonomically log everything we have in this PR. For example, there is currently no good way for a separate library to properly trace connection attempts with all of the relevant metadata, especially with timings for each phase.

sfackler:
That approach seems like it could work in principle - I'd be interested in what the hook trait would look like.

PR #1059, opened 2023-08-19
`tokio-postgres`: Set user to executing process' user by default.

### LINKED ISSUE #1024: Make username default to current login name
This is the way libpq (and other libraries based on it) does this:

* https://github.com/postgres/postgres/blob/REL_15_2/src/interfaces/libpq/fe-connect.c#L5947-L6056

At least some third-party libraries not based on libpq do that as well:

* https://github.com/jackc/pgx/blob/v5.3.1/pgconn/defaults.go#L18-L23
* https://github.com/lib/pq/blob/v1.10.8/connector.go#L103-L112

I think it would make sense to do it here for compatibility with other libraries.

ISibboI:
This mimics the behaviour of libpq and some other libraries (see #1024). This commit uses the `whoami` crate, and thus goes as far as defaulting the user to the executing process' user name on all operating systems.

Closes #1024

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
       - run: docker compose up -d
       - uses: sfackler/actions/rustup@master
         with:
-          version: 1.65.0
+          version: 1.67.0
       - run: echo "version=$(rustc --version)" >> $GITHUB_OUTPUT
         id: rust-version
       - uses: actions/cache@v3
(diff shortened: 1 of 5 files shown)


ISibboI:
How should I go about the time package?

sfackler:
You can bump the version we use in CI up to 1.67: https://github.com/sfackler/rust-postgres/blob/master/.github/workflows/ci.yml#L85

sfackler:
Thanks!

PR #1060, opened 2023-08-22
tokio-postgres: Expose table OID and column ID on Column

glittershark:
Pass through the table OID and column ID from field descriptions for prepared statements to fields on Column, with exposed getters


sfackler:
Covered by #1084

PR #1064, opened 2023-08-31
Transactions aren't executed with "psql -c"

tbjuhasz:
The transaction did not appear to execute when running a set of SQL commands using psql commands with the '-c' option, such as "SELECT * FROM t; BEGIN; UPDATE t SET k = id; END;". This issue was caused by the tokio-postgres crate adding column information to all EmptyQueryResponses for every SQL query. The resolution for this issue was to only add the column information to 'SELECT' queries.


tbjuhasz:
this is the wrong repo.

PR #1066, opened 2023-09-22
Refactor `tokio_postgres::Connection` implementation

50U10FCA7:
Minor changes improving `tokio_postgres::Connection` implementation:
- makes `Connection::stream` polling independent from polling pending requests/responses;
- `Connection::poll_read` now ensures the written messages was flushed, before sending the next ones;
- covers private items with docs.


sfackler:
What is the purpose of these changes?

50U10FCA7:
@sfackler Thanks for the fast reply.

These changes should allow to process something like this more efficiently:
```
let (client, connection) =
        tokio_postgres::connect(..).await?;

let rc_client = Rc::new(client);

tokio::spawn_local(move || {
    rc_client.clone().query(..)
})

tokio::spawn_local(move || {
    rc_client.clone().query(..)
})
```

without blocking `Connection` polling on specific request/response from `Client`.

Also, now `Connection::stream::poll_flush` is waited by `Connection` to return `Poll::Ready` (to ensure the msg was written).

Private docs updated to make code a little bit readable.

sfackler:
> without blocking `Connection` polling on specific request/response from `Client`.

What blocking is happening before this change?



> Also, now `Connection::stream::poll_flush` is waited by `Connection` to return `Poll::Ready` (to ensure the msg was written).

In what context would a message not be written?

50U10FCA7:
@sfackler 

> What blocking is happening before this change?

[`poll_read`](https://github.com/sfackler/rust-postgres/blob/c5ff8cfd86e897b7c197f52684a37a4f17cecb75/tokio-postgres/src/connection.rs#L95) on master branch can process only one response at time, even if socket is already received messages of the next responses:
- receive ([1](https://github.com/sfackler/rust-postgres/blob/c5ff8cfd86e897b7c197f52684a37a4f17cecb75/tokio-postgres/src/connection.rs#L102), [2](https://github.com/sfackler/rust-postgres/blob/c5ff8cfd86e897b7c197f52684a37a4f17cecb75/tokio-postgres/src/connection.rs#L90)) a new `BackendMessage`;
- try to [send](https://github.com/sfackler/rust-postgres/blob/c5ff8cfd86e897b7c197f52684a37a4f17cecb75/tokio-postgres/src/connection.rs#L146) received `BackendMessage` to `Client`;
- receiver is [not ready](https://github.com/sfackler/rust-postgres/blob/c5ff8cfd86e897b7c197f52684a37a4f17cecb75/tokio-postgres/src/connection.rs#L159);
- try to send it in the [next poll](https://github.com/sfackler/rust-postgres/blob/c5ff8cfd86e897b7c197f52684a37a4f17cecb75/tokio-postgres/src/connection.rs#L166) (or delayed message which is always `AsyncMessage` and will poll `Connection` again, because it yields a value).

But there can be a situation when socket already has a full response (all its messages) and some messages of another one. In this case we can try to send a part of the second response within current poll:
- [receive](https://github.com/sfackler/rust-postgres/blob/ed54e9683c3ede49f7b4569b8d2a3058192180ea/tokio-postgres/src/connection.rs#L172) all available `BackendMessage`s;
- try to [send](https://github.com/sfackler/rust-postgres/blob/ed54e9683c3ede49f7b4569b8d2a3058192180ea/tokio-postgres/src/connection.rs#L214) all messages of the pending responses within current poll.

I think this should be more efficient, because we don't have a guarantee that runtime will poll the `Connection` again right after the current poll. And because we not poll the socket again (we are waiting for a `Client`'s receiver), the next time `Connection::poll_read` will be polled by runtime only if:
1. A new `Client`'s request received;
2. When `Client`'s receiver is dropped/able to receive a response.

If we assume that `Response`'s receiver is not polled for a long time (doing more important things in `futures::future::select(client.query(..), another_fut)` for example) and we have another one `Response` that is received by socket and ready to be send - we should do it I think.

> In what context would a message not be written?

In case when [`Framed`](https://github.com/tokio-rs/tokio/blob/74c7a879855cac99ca54f292305d85894b2e0b0c/tokio-util/src/codec/framed_impl.rs#L287) cannot write a frame to a socket (I think its possible if machine has a small window and full frame cannot be written, but not sure it is possible in real-world scenarios).

p.s. Also found that both [`TcpStream`](https://github.com/tokio-rs/tokio/blob/74c7a879855cac99ca54f292305d85894b2e0b0c/tokio/src/net/tcp/stream.rs#L1348) and [`UnixStream`](https://github.com/tokio-rs/tokio/blob/74c7a879855cac99ca54f292305d85894b2e0b0c/tokio/src/net/unix/stream.rs#L981) supports vectored write, but [`tokio_postgres::Socket`](https://github.com/sfackler/rust-postgres/blob/c5ff8cfd86e897b7c197f52684a37a4f17cecb75/tokio-postgres/src/socket.rs#L47) doesn't use it, is there any reason for that?

sfackler:
> * [receive](https://github.com/sfackler/rust-postgres/blob/ed54e9683c3ede49f7b4569b8d2a3058192180ea/tokio-postgres/src/connection.rs#L172) all available `BackendMessage`s;
> * try to [send](https://github.com/sfackler/rust-postgres/blob/ed54e9683c3ede49f7b4569b8d2a3058192180ea/tokio-postgres/src/connection.rs#L214) all messages of the pending responses within current poll.

A sufficiently fast server could cause that logic to OOM by giving it an unbounded amount of messages to buffer.

> I think this should be more efficient, because we don't have a guarantee that runtime will poll the `Connection` again right after the current poll.

Efficiency claims should be measured, not assumed.

> In case when [`Framed`](https://github.com/tokio-rs/tokio/blob/74c7a879855cac99ca54f292305d85894b2e0b0c/tokio-util/src/codec/framed_impl.rs#L287) cannot write a frame to a socket (I think its possible if machine has a small window and full frame cannot be written, but not sure it is possible in real-world scenarios).

The next poll of the connection will continue to drive the flush.

> p.s. Also found that both [`TcpStream`](https://github.com/tokio-rs/tokio/blob/74c7a879855cac99ca54f292305d85894b2e0b0c/tokio/src/net/tcp/stream.rs#L1348) and [`UnixStream`](https://github.com/tokio-rs/tokio/blob/74c7a879855cac99ca54f292305d85894b2e0b0c/tokio/src/net/unix/stream.rs#L981) supports vectored write, but [`tokio_postgres::Socket`](https://github.com/sfackler/rust-postgres/blob/c5ff8cfd86e897b7c197f52684a37a4f17cecb75/tokio-postgres/src/socket.rs#L47) doesn't use it, is there any reason for that?

Because there aren't any vectored writes happening in the client AFAIK.

PR #1070, opened 2023-10-11
Adds support for zero-copy FromSql derive

JosephMoniz:
Similar in spirit and implementation to how `serde` supports borrowed deserialization lifetimes. Like `serde`, it uses `'de` as the lifetime of the trait (`FromSql<'de>`) and it adds all borrowed liftimes as lifetime bounds on `'de`.

Also like `serde` all top level references automatically have their lifetimes borrowed, but container types carrying references must explicity be borrowed with the `#[postgres(borrow)]` annotation.

Follows up on: https://github.com/sfackler/rust-postgres/pull/574

### DIFF
--- a/postgres-derive/src/fromsql.rs
+++ b/postgres-derive/src/fromsql.rs
@@ -1,6 +1,8 @@
+use std::collections::{BTreeSet, HashSet};
 use proc_macro2::{Span, TokenStream};
 use quote::{format_ident, quote};
 use std::iter;
+use std::iter::FromIterator;
 use syn::{
     punctuated::Punctuated, token, AngleBracketedGenericArguments, Data, DataStruct, DeriveInput,
     Error, Fields, GenericArgument, GenericParam, Generics, Ident, Lifetime, PathArguments,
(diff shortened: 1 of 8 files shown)


postgres-derive/src/fromsql.rs
```diff
@@ -233,16 +247,15 @@ fn composite_body(ident: &Ident, fields: &[Field]) -> TokenStream {
...
 }
 
-fn build_generics(source: &Generics) -> (Generics, Lifetime) {
-    // don't worry about lifetime name collisions, it doesn't make sense to derive FromSql on a struct with a lifetime
-    let lifetime = Lifetime::new("'a", Span::call_site());
-
+fn build_generics(source: &Generics, borrowed_lifetimes: HashSet<Lifetime>) -> (Generics, Lifetime) {
+    // This is the same parent lifetime name serde uses
+    let lifetime = Lifetime::new("'de", Span::call_site());
+    // Sort lifetimes for deterministic code-gen
```
sfackler, line 253: This could also preserve the declared order of lifetimes rather than going through a hashset.

JosephMoniz: You're right. I was over-optimizing on the extraction path to handle lifetime dedup. In the real-world though, a simple `borrowed_lifetimes.contains(lifetime)` shouldn't ever be a problem. I made the changes to go through a vec instead of a hashset and preserve declaration order.

JosephMoniz:
Can anyone w/ maintainer privileges please kick off the test workflow 🙏

PR #1078, opened 2023-11-01
make simple_query_raw public

uuip:
make simple_query_raw public, so we can iter result without collect it when there are 100 million rows.


uuip:
We should use portal rather than simple query when handling a large number of rows that need to be fetched incrementally or processed in batches.

PR #1084, opened 2023-12-05
Add table_oid and column_id to column structure of prepared statements

bikeshedder:
I want to access the underlying table and column of a query and found those two information not exposed in the `statement::Column` structure.

I also simplified the `Debug` implementation of `Column` by using `r#type` rather than `type_` so `#[derive(Debug)]` can be used. As the argument list grew longer I also removed the `new` method in favor of making all the fields `pub(crate)`.

Should we maybe expose the three remaining fields as well?
- `type_size: i16`
- `type_modifier: i32`
- `format: i16`

I don't have an immediate need for them. They could provide useful in the future, though.

What are your thoughts on using `NonZeroU32` and `NonZeroI16`? I'm not 100% sure if that's really a good choice for the type. Yet it felt somewhat natural to use it as `0` denotes that this information is not available.

### DIFF
--- a/tokio-postgres/CHANGELOG.md
+++ b/tokio-postgres/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Disable `rustc-serialize` compatibility of `eui48-1` dependency
 * Remove tests for `eui48-04`
-
+* Add `table_oid` and `field_id` fields to `Columns` struct of prepared statements.
 
 ## v0.7.10 - 2023-08-25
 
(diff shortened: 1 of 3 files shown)


bikeshedder:
@sfackler Sorry for bothering you. Is there a chance this PR could be merged soon? I'm in need of this feature for an upcoming code generation tool for type safe queries based on tokio-postgres.

sfackler:
Thanks! Sorry for the delay.

sfackler:
Hmm, actually I think I'm going to adjust the return values to `Option<Uid>` rather than use `NonZeroU32`. That's what we use elsewhere in the API.

bikeshedder:
@sfackler Is there a chance you could release a new `tokio-postgres` version. I'm in progress of releasing a crate that does depend on this feature. I can't push this to `crates.io`unless a new version of `tokio-postgres` is published first.

I plan on releasing my crate in 1-2 weeks so a release of `tokio-postgres` would be highly appreciated.

PR #1094, opened 2024-01-15
feat(types): add support for prost messages

TroyKomodo:
Adds a wrapper type for prost messages to help deserialize bytea values directly into the protobuf structure.


TroyKomodo:
Unsure if this change is welcome, i thought it might be useful to others since this is something i use in my projects.
I am in the process of migrating from sqlx. So i thought i would PR some of the utils I use.

sfackler:
I think this is something that would be best left to an external library.

PR #1095, opened 2024-01-16
Add smol_str 0.2 support

novacrazy:
Add `smol_str 0.2` support, effectively a copy of the 0.1 file. Though I noticed the feature flags for 0.1 are missing an underscore, but changing that would break things so I just matched it for the new one.

### DIFF
--- a/postgres-types/Cargo.toml
+++ b/postgres-types/Cargo.toml
@@ -22,6 +22,7 @@ with-geo-types-0_6 = ["geo-types-06"]
 with-geo-types-0_7 = ["geo-types-0_7"]
 with-serde_json-1 = ["serde-1", "serde_json-1"]
 with-smol_str-01 = ["smol_str-01"]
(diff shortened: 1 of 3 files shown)


novacrazy:
Any chance this could be merged sometime? Starting to get cyclic dependency errors due to how many crates I've had to patch `ToSql`/`FromSql` support into.

novacrazy:
This has been open long enough that there's now a 0.3 version of `smol_str`. Updated and rebased to latest for easier merging.

paolobarbolini:
These lists also need to be updated:

https://github.com/rust-postgres/rust-postgres/blob/e1cd6beef3a1530642a2abaf3584d6bd8ed6cd45/postgres/src/lib.rs#L56-L66

https://github.com/rust-postgres/rust-postgres/blob/e1cd6beef3a1530642a2abaf3584d6bd8ed6cd45/tokio-postgres/src/lib.rs#L108-L119

Issue #924, opened 2022-07-19
tokio_postgres client query awaits forever


marco-viewpoint:
My first try with this library and although I can establish a connection with the DB running a query always locks up my lambda until the 10 minute timer runs out.
Last output is "query sent", client.is_closed() returns false, making the user/pw wrong results in a connection error instead as expected.

I tried different queries and different functions (query_one, query_raw, etc.).
The code runs in a AWS lambda and connects to a serverless Aurora Postgresql DB inside a VPC.

Is there any additional debugging I can do to find the root cause?

src/main.rs
```
let cert = fs::read("global-bundle.pem")?;
println!("file read");
let cert = Certificate::from_pem(&cert)?;
println!("cert created");

let connector = TlsConnector::builder()
		.add_root_certificate(cert)
		.build()?;
println!("connector built");
let connector = MakeTlsConnector::new(connector);
println!("connector created");

let connect_result = tokio_postgres::connect(
		"host=#####.us-west-2.rds.amazonaws.com port=5432 user=### password=### dbname=### connect_timeout=10",
		connector,
).await;
println!("connection finished");

if connect_result.is_err() {
		let error_msg = connect_result.err().unwrap().to_string();
		println!("Error: {}", error_msg);
		return Ok(json!({ "error": error_msg }))
}

let (client, _connection) = connect_result.unwrap();
println!("connection unwrapped");
println!("Client closed: {}", client.is_closed());

let db_response = client.query_one("SELECT name FROM workspaces WHERE id = 760", &[]);
println!("query sent");
let db_response = db_response.await;
println!("query finished");
```

Cargo.toml
```
[package]
name = "test"
version = "0.1.0"
edition = "2021"
autobins = false

[dependencies]
lambda_http = "*"
lambda_runtime = "*"
graphql_client = "*"
env_logger = "*"
serde = "*"
serde_json = "*"
ndarray = "*"
tokio = { version = "1", features = ["macros"] }
tracing = { version = "0.1", features = ["log"] }
tracing-subscriber = { version = "0.3", default-features = false, features = ["fmt"] }
tokio-postgres = "*"
postgres-native-tls = "*"
native-tls = "*"
openssl = { version = "0.10", features = ["vendored"] }
futures = "*"

[[bin]]
name = "bootstrap"
path = "src/main.rs"

[profile.release]
strip = true
```

marco-viewpoint:
1. `SELECT 1` fails as well.
[...]

PR #1099, opened 2024-01-28
Improve the documentation of tokio-postgres connect

JJaviMS:
As it can be seen on #924 and from my personal experience, it can be easily missed the use of the connection object when connecting using the tokio-postgres. I have improved the docs for that method so it indicates the importance of the connection object and give an example of how to use it.

### DIFF
--- a/tokio-postgres/src/lib.rs
+++ b/tokio-postgres/src/lib.rs
@@ -180,12 +180,32 @@ mod transaction_builder;
 pub mod types;
 
 /// A convenience function which parses a connection string and connects to the database.
+/// 
+/// This method returns the client which will allow you to interact with the database and a connection
+/// object which is the one that performs the communication with the dabase. **It's important to use the
+/// connection object as shown in the below example, if it is ignored then the client object will not work
+/// and when a .wait is called it will await forever.**
 ///
 /// See the documentation for [`Config`] for details on the connection string format.
 ///
 /// Requires the `runtime` Cargo feature (enabled by default).
 ///
 /// [`Config`]: config/struct.Config.html
+/// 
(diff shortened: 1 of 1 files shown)


tokio-postgres/src/lib.rs
```diff
@@ -180,12 +180,32 @@ mod transaction_builder;
 pub mod types;
 
 /// A convenience function which parses a connection string and connects to the database.
+/// 
+/// This method returns the client which will allow you to interact with the database and a connection
+/// object which is the one that performs the communication with the dabase. **It's important to use the
+/// connection object as shown in the below example, if it is ignored then the client object will not work
+/// and when a .wait is called it will await forever.**
```
paolobarbolini, line 187: I've cleaned it up a bit.

```suggestion
/// This method returns the client object, which will allow you to interact with the database,
/// and a connection object, which performs the communication with the database.
///
/// **The connection object must be actively polled, which is usually done by spawning it into
/// it's own tokio task, as shown below. If this is not done, calls to the client object will hang forever.**
```

tokio-postgres/src/lib.rs
```diff
@@ -180,12 +180,32 @@ mod transaction_builder;
...
 ///
 /// See the documentation for [`Config`] for details on the connection string format.
 ///
 /// Requires the `runtime` Cargo feature (enabled by default).
 ///
 /// [`Config`]: config/struct.Config.html
```
paolobarbolini, line 193: This should go below the examples section.

paolobarbolini (changes requested):
Good idea. I've left some minor changes

PR #1101, opened 2024-02-11
Shrink query_opt/query_one codegen size very slightly

novacrazy:
Avoids needing an extra await point and shrinks the future size by around 72 bytes. Incredibly minor, but still. 

Was just looking for places to shrink future sizes in my application and tried this out.

### DIFF
--- a/tokio-postgres/src/client.rs
+++ b/tokio-postgres/src/client.rs
@@ -274,19 +274,9 @@ impl Client {
     where
         T: ?Sized + ToStatement,
     {
-        let stream = self.query_raw(statement, slice_iter(params)).await?;
-        pin_mut!(stream);
-
-        let row = match stream.try_next().await? {
-            Some(row) => row,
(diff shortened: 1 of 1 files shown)


tokio-postgres/src/client.rs
```diff
@@ -310,16 +300,17 @@ impl Client {
         let stream = self.query_raw(statement, slice_iter(params)).await?;
         pin_mut!(stream);
 
-        let row = match stream.try_next().await? {
-            Some(row) => row,
-            None => return Ok(None),
-        };
+        let mut first = None;
+
+        while let Some(row) = stream.try_next().await? {
```
sfackler, line 305: Probably deserves a comment about why it's structured this way.

novacrazy: Sure, added a comment.

sfackler:
Rebasing should fix the build.

PR #1110, opened 2024-03-03
Add `ReadOnly` in `TargetSessionAttrs`

chandr-andr:
Hi!
I've found that there is no option to set `read-only` in `target-session-attrs` for connection to PostgreSQL.  
Please, take a look at this pull request.

### DIFF
--- a/tokio-postgres/src/config.rs
+++ b/tokio-postgres/src/config.rs
@@ -34,6 +34,8 @@ pub enum TargetSessionAttrs {
     Any,
     /// The session must allow writes.
     ReadWrite,
+    /// The session allow only reads.
+    ReadOnly,
 }
 
 /// TLS configuration.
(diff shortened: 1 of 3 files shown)


tokio-postgres/src/config.rs
```diff
@@ -34,6 +34,8 @@ pub enum TargetSessionAttrs {
     Any,
     /// The session must allow writes.
     ReadWrite,
+    /// The session allow only reads.
+    ReadOnly,
```
sfackler, line 38: This presumably needs to be used somewhere in the connection code.

chandr-andr: Sry, haven't committed my latest changes. Please have a look.

sfackler:
Thanks!

PR #1118, opened 2024-03-13
use `split_once` instead of `split` to parse lsn strings

laxjesse:
[`str::split`](https://doc.rust-lang.org/std/primitive.str.html#method.split) allocates a vector and generates considerably more instructions when compiled than [`str::split_once`](https://doc.rust-lang.org/std/primitive.str.html#method.split_once).

[`u64::from_str_radix(split_lo, 16)`](https://doc.rust-lang.org/std/primitive.u64.html#method.from_str_radix) will error if the `lsn_str` contains more than one `/` so this change should result in the same behavior as the current implementation despite not explicitly checking this.

### DIFF
--- a/postgres-types/CHANGELOG.md
+++ b/postgres-types/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


sfackler:
Thanks! The CI failure is unrelated - I'll fix that later.

Issue #1120, opened 2024-03-20
`TransactionBuilder::start` Future dropped before completion


wquark:
`batch_execute` submits the creation of the transaction (https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/src/transaction_builder.rs#L109), but the `Transaction` object itself may never be constructed in case the future is dropped.

https://github.com/sfackler/rust-postgres/commit/f6189a95f24af45068ecfd6e3c4e6b71ac8e43fe fixed it for `Client::transaction`.

sfackler:
Good find. We should probably implement `transaction` with `TransactionBuilder` to avoid having two places that need to deal with this.

wquark:
Makes sense. Anything else you need from my end? Is this something you'd knock out quickly or would you rather have a PR?

sfackler:
Either way!

PR #1121, opened 2024-03-20
Fix transction rollback on `Future` drop

wquark:
PR for https://github.com/sfackler/rust-postgres/issues/1120. Also rollsback the savepoint (found while implementing this).


tokio-postgres/src/client.rs
```diff
@@ -480,8 +490,13 @@ impl Client {
                     return;
                 }
 
+                let query = if let Some(sp) = self.savepoint {
+                    format!("ROLLBACK TO {}", sp.name)
+                } else {
+                    "ROLLBACK".to_string()
+                };
```
sfackler, line 497: nit: this'd be cleaner as a match IMO

sfackler:
Looks good to me other than the one nit!

tokio-postgres/src/client.rs
wquark: I matched style with the other instances (e.g https://github.com/sfackler/rust-postgres/blob/64caf4c947cfde6c292f42406c046dce79f771bf/tokio-postgres/src/transaction.rs#L42-L46 and https://github.com/sfackler/rust-postgres/blob/64caf4c947cfde6c292f42406c046dce79f771bf/tokio-postgres/src/transaction.rs#L83-L87). Would you like to switch all of those or just this one?

sfackler: Oh huh, let's keep it as-is then.

sfackler:
Looks like it needs a `cargo fmt`, and this should fix the chrono deprecation warning:

```diff
diff --git a/postgres-types/src/chrono_04.rs b/postgres-types/src/chrono_04.rs
index d599bde0..64166309 100644
--- a/postgres-types/src/chrono_04.rs
+++ b/postgres-types/src/chrono_04.rs
@@ -111,9 +111,10 @@ impl ToSql for DateTime<FixedOffset> {
 impl<'a> FromSql<'a> for NaiveDate {
     fn from_sql(_: &Type, raw: &[u8]) -> Result<NaiveDate, Box<dyn Error + Sync + Send>> {
         let jd = types::date_from_sql(raw)?;
+        let jd = Duration::try_days(jd)?;
         base()
             .date()
-            .checked_add_signed(Duration::days(i64::from(jd)))
+            .checked_add_signed(jd)
             .ok_or_else(|| "value too large to decode".into())
     }
```

wquark:
Ran `cargo fmt` and fixed warnings. Left the warning fixes in a separate commit since they seemed unrelated.

wquark:
Hey @sfackler, would be great to get this pushed forward so we can remove our patched dependencies. Let me know if there are other blockers, thanks.

PR #1125, opened 2024-03-31
Updated whoami dependency

SarthakSingh31:
Updating due to [RUSTSEC-2024-0020](https://rustsec.org/advisories/RUSTSEC-2024-0020).


sfackler:
This is not necessary. Run `cargo update` locally.

PR #1126, opened 2024-04-03
Support scalar queries.

rikvdkleij:
This PR adds support for querying scalars with the following functions:
- `query_scalar` returns a vector of scalars
- `query_one_scalar` return one scalar
- `query_opt_scalar` rerun optional scalar

It removes the need to always handle a row. 

Also in the process of migrating from `sqlx` to `tokio-postgres` it's convenient to have support for scalar query functions.

Questions and/or suggestions are welcome!


paolobarbolini (commented):
`query_one` considers it an error for the query to return more than one row. Should the scalar methods consider it an error for the query to return more than one column?

Issue #834, opened 2021-10-28
`Transaction` isn't rolled back in case `Client::transaction()` is dropped before completion


ilslv:
Currently `Transaction` can be left open in case `Future` created by `Client::transaction` dropped before polling to completion.

```rust
pub async fn transaction(&mut self) -> Result<Transaction<'_>, Error> {
    self.batch_execute("BEGIN").await?; // suspension point 

    Ok(Transaction::new(self))
}
```

`batch_execute()`  under the hood synchronously sends `RequestMessages` to the `Connection` and then asynchronously polls `Responses`. While `Pending` on `Responses`,  `Future` can be dropped before `Transaction::new()` is created. Then `Transaction::drop()` won't be called and `Transaction` will be left open. 

### Reproduction

```rust
pin_project! {
    /// Polls `F` at most `polls_left` times returning `Some(F::Output)` if
    /// [`Future`] returned [`Poll::Ready`] or [`None`] otherwise.
    struct Cancellable<F> {
        #[pin]
        fut: F,
        polls_left: usize,
    }
}

impl<F: Future> Future for Cancellable<F> {
    type Output = Option<F::Output>;

    fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
        let this = self.project();
        match this.fut.poll(ctx) {
            Poll::Ready(r) => Poll::Ready(Some(r)),
            Poll::Pending => {
                *this.polls_left = this.polls_left.saturating_sub(1);
                if *this.polls_left == 0 {
                    Poll::Ready(None)
                } else {
                    Poll::Pending
                }
            }
        }
    }
}

async fn current_transaction_id(client: &Client) -> i64 {
    client
        .query("SELECT txid_current()", &[])
        .await
        .unwrap()
        .pop()
        .unwrap()
        .get::<_, i64>("txid_current")
}

async fn in_transaction(client: &Client) -> bool {
    current_transaction_id(client).await == current_transaction_id(client).await
}

#[tokio::test]
async fn transaction_future_cancellation() {
    let mut client = connect("user=postgres").await;

    for i in 0.. {
        let done = {
            let txn = client.transaction();
            let fut = Cancellable {
                fut: txn,
                polls_left: i,
            };
            fut.await
                .map(|res| res.expect("transaction failed"))
                .is_some()
        };

        assert!(!in_transaction(&client).await);

        if done {
            break;
        }
    }
}
```

sfackler:
Done!

PR #1127, opened 2024-04-03
Fix transaction not being rolled back on `TransactionBuilder::start()` `Future` dropped before completion

irkill:
Fixes the same issue as described in the https://github.com/sfackler/rust-postgres/issues/834, but for a `TransactionBuilder::start()` `Future`.

### DIFF
--- a/tokio-postgres/src/client.rs
+++ b/tokio-postgres/src/client.rs
@@ -1,4 +1,4 @@
-use crate::codec::{BackendMessages, FrontendMessage};
+use crate::codec::BackendMessages;
 use crate::config::SslMode;
 use crate::connection::{Request, RequestMessages};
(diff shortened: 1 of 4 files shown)


tokio-postgres/src/client.rs
```diff
@@ -469,43 +469,7 @@ impl Client {
...
-            let mut cleaner = RollbackIfNotDone {
-                client: self,
-                done: false,
-            };
-            self.batch_execute("BEGIN").await?;
-            cleaner.done = true;
-        }
-
-        Ok(Transaction::new(self))
+        Transaction::begin(self).await
```
sfackler, line 472: I think it'd be simpler to just have this call `self.transaction_builder().start().await` and have the cancellation logic in one place.

PR #1128, opened 2024-04-05
Update whoami to newest version

LeNei:
Due to an issue in [whoami](https://github.com/ardaku/whoami) versions < `1.5.0` can lead to buffer overflow as discussed [here](https://github.com/ardaku/whoami/issues/91). This PR is just a bump on the version to avoid possible exploits.


sfackler:
https://github.com/sfackler/rust-postgres/pull/1125

PR #1137, opened 2024-05-02
Avoid extra clone in config if possible

nyurik:
Using `impl Into<String>` instead of `&str` in a fn arg allows both `&str` and `String` as parameters - thus if the caller already has a String object that it doesn't need, it can pass it in without extra cloning.

The same might be done with the password, but may require closer look.

### DIFF
--- a/tokio-postgres/src/config.rs
+++ b/tokio-postgres/src/config.rs
@@ -248,8 +248,8 @@ impl Config {
     /// Sets the user to authenticate with.
     ///
     /// Defaults to the user executing this process.
-    pub fn user(&mut self, user: &str) -> &mut Config {
-        self.user = Some(user.to_string());
+    pub fn user(&mut self, user: impl Into<String>) -> &mut Config {
+        self.user = Some(user.into());
         self
     }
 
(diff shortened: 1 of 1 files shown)


sfackler:
I'm not super opposed to this, but at the same time I'm not sure it'll make a huge difference. What portion of runtime is spent constructing config?

nyurik:
@sfackler config specifically is not that big of a deal of course.  My concern is that any frequently used code teaches users on how to do things - and in this case, IMO, the API should have only allowed `String` as the parameter, but because breaking API is not worth it, might as well introduce a backwards compatible solution.

Reasoning:
* Function should require owned value when it needs to keep it - to let caller know how to use it and optimize for it
* [API recommendations for builders](https://rust-lang.github.io/api-guidelines/type-safety.html?highlight=String#non-consuming-builders-preferred) - suggesting the structure of APIs
* [My question to community](https://users.rust-lang.org/t/consumed-param-that-can-be-owned-or-a-ref/110744/8) on how to do this types of API -- lots of good responses there

P.S. community recommended using explicit generic rather than `impl Into<String>` as a param type.

PR #1144, opened 2024-05-29
Adds RowDescription to the SimpleQueryMessage

### LINKED ISSUE #1143: Idea: Add RowDescription to SimpleQueryMessage
Currently when processing the `SimpleQueryMessage`, if there are no `DataRows` coming from postgres the column data is inaccessible since the data is embedded in the `SimpleQueryRow`. In particular, this makes it difficult if you're displaying user submitted or dynamic query results where being able to determine the columns from the query after it's been sent to the server would be useful. Adding an enum value of `RowDescription(Arc<[SimpleColumn]>)` would alleviate this since you could separately process the columns even without a `DataRow` being returned.

rigby-dane:
Adds RowDescription to the SimpleQueryMessage to allow column information to be consumed without needing to consume any DataRows

Resolves #1143

### DIFF
--- a/tokio-postgres/src/simple_query.rs
+++ b/tokio-postgres/src/simple_query.rs
@@ -85,36 +85,34 @@ impl Stream for SimpleQueryStream {
 
     fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         let this = self.project();
-        loop {
-            match ready!(this.responses.poll_next(cx)?) {
-                Message::CommandComplete(body) => {
-                    let rows = extract_row_affected(&body)?;
-                    return Poll::Ready(Some(Ok(SimpleQueryMessage::CommandComplete(rows))));
-                }
-                Message::EmptyQueryResponse => {
-                    return Poll::Ready(Some(Ok(SimpleQueryMessage::CommandComplete(0))));
-                }
-                Message::RowDescription(body) => {
-                    let columns: Arc<[SimpleColumn]> = body
-                        .fields()
-                        .map(|f| Ok(SimpleColumn::new(f.name().to_string())))
-                        .collect::<Vec<_>>()
-                        .map_err(Error::parse)?
-                        .into();
-
-                    *this.columns = Some(columns.clone());
-                    return Poll::Ready(Some(Ok(SimpleQueryMessage::RowDescription(columns.clone()))));
-                }
-                Message::DataRow(body) => {
-                    let row = match &this.columns {
-                        Some(columns) => SimpleQueryRow::new(columns.clone(), body)?,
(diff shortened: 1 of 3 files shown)


tokio-postgres/src/simple_query.rs
```diff
@@ -85,35 +85,36 @@ impl Stream for SimpleQueryStream {
...
-                        Some(columns) => SimpleQueryRow::new(columns.clone(), body)?,
-                        None => return Poll::Ready(Some(Err(Error::unexpected_message()))),
-                    };
-                    return Poll::Ready(Some(Ok(SimpleQueryMessage::Row(row))));
-                }
-                Message::ReadyForQuery(_) => return Poll::Ready(None),
-                _ => return Poll::Ready(Some(Err(Error::unexpected_message()))),
+                *this.columns = Some(columns.clone());
+                Poll::Ready(Some(Ok(SimpleQueryMessage::RowDescription(
+                    columns.clone(),
```
sfackler, line 106: You shouldn't need to clone here.

rigby-dane: I'm confused on how to best avoid this clone. Previously, it was able to move the Arc<[SimpleColumn]> over to *this.columns because the Arc wasn't shared, but now since it has to go into the RowDescription and *this.columns, it can no longer move the Arc (from my understanding).

The only way thus far I've been able to avoid 2 clones has been to do:
```
*this.columns = Some(columns);
Poll::Ready(Some(Ok(SimpleQueryMessage::RowDescription(
    this.columns.as_mut().unwrap().clone(),
))))
```

Which is more opaque to me, but I might be missing some benefit to this or that might be the preferred style.

rigby-dane: I was originally confused about this, but thought it over and figured out a solution. I've added a change to reduce it to just 1 clone instead of 2.

tokio-postgres/src/simple_query.rs
```diff
@@ -101,9 +101,9 @@ impl Stream for SimpleQueryStream {
                     .map_err(Error::parse)?
                     .into();
 
-                *this.columns = Some(columns.clone());
+                *this.columns = Some(columns);
                 Poll::Ready(Some(Ok(SimpleQueryMessage::RowDescription(
-                    columns.clone(),
+                    this.columns.as_ref().unwrap().clone(),
                 ))))
```
sfackler, line 107: Wouldn't this work?

```rust
                *this.columns = Some(columns.clone());
                Poll::Ready(Some(Ok(SimpleQueryMessage::RowDescription(
                    columns
                ))))
```

rigby-dane: Yes, it does. I was clearly overthinking it. Thank you!

sfackler:
Thanks!

PR #1145, opened 2024-05-29
Expose type modifier in Column struct

SergeiPatiakin:
## What
Expose the 4-byte type-specific type modifier on the Column struct

## Why
I would like access type modifiers for prepared statement columns. E.g. for a numeric column I would like  to get the numeric precision and scale. For a varchar column I would like to get the varchar length limit.

## Context
This is an extension of https://github.com/sfackler/rust-postgres/pull/1084 where other table and column OIDs were exposed in Column, and exposing type modifiers was noted as a possibility.


rishitesh-snt:
+ 1 for this PR. This would enable the result set handling in a much better way rather than falling back to defaults.

PR #1146, opened 2024-06-03
Support build on wasm32

L-jasmine:
[In this line](https://github.com/sfackler/rust-postgres/blob/98f5a11bc0a8e451552d8941ffa078c7eb6cd60c/tokio-postgres/src/config.rs#L207-L208), the parameter KeepaliveConfig is ignored when compiled to wasm32, while there is no corresponding conditional compilation elsewhere. Therefore, I added a conditional compilation for wasm32 in the places where Keepalive is needed. This modification allows the crate to be used normally in wasm32 after replacing the underlying tokio with a patch.

### DIFF
--- a/tokio-postgres/src/cancel_query.rs
+++ b/tokio-postgres/src/cancel_query.rs
@@ -34,6 +34,7 @@ where
         config.port,
         config.connect_timeout,
         config.tcp_user_timeout,
+        #[cfg(not(target_arch = "wasm32"))]
         config.keepalive.as_ref(),
     )
     .await?;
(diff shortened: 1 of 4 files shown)


sfackler:
We already test that the crate compiles against wasm32: https://github.com/sfackler/rust-postgres/blob/master/.github/workflows/ci.yml#L50-L75.

If tokio gains direct wasm32 support I'd be happy to expand the supported feature set here, but I don't want to make changes that only work when building against an unspecified tokio fork.

L-jasmine:
This is not specific to a particular tokio; it’s just that this crate does not define KeepaliveConfig in wasm32. I’ve ignored all references to KeepaliveConfig in wasm32, and I believe these changes are reasonable.

paolobarbolini:
> Your CI still passes with this PR.

Could you change the CI configuration so that the wasm32 job enables the `runtime` feature? Currently it does not.

PR #1147, opened 2024-06-04
Work with pools that don't support prepared statements

ramnivas:
Introduce a new `query_with_param_types` method that allows to specify Postgres type parameters. This obviated the need to use prepared statements just to obtain parameter types for a query. It then combines parse, bind, and execute in a single packet.

Related: #1017, #1067


tokio-postgres/src/client.rs
```diff
@@ -364,6 +364,88 @@ impl Client {
...
+        &self,
+        statement: &str,
+        params: &[(&(dyn ToSql + Sync), Type)],
+    ) -> Result<RowStream, Error> {
+        fn slice_iter<'a>(
+            s: &'a [(&'a (dyn ToSql + Sync), Type)],
+        ) -> impl ExactSizeIterator<Item = (&'a dyn ToSql, Type)> + 'a {
+            s.iter()
+                .map(|(param, param_type)| (*param as _, param_type.clone()))
+        }
```
sfackler, line 444: There's no need to factor this into a separate function here since it's only being called one place.

tokio-postgres/src/query.rs
```diff
@@ -50,13 +54,125 @@ where
...
     Ok(RowStream {
-        statement,
+        statement: statement,
         responses,
         rows_affected: None,
         _p: PhantomPinned,
     })
 }
 
+enum QueryProcessingState {
```
sfackler, line 64: It feels like this is a lot of boilerplate over a few `match responses.next().await? { ... }` in a sequence.

sfackler:
Thinking about it, it seems like we could make query_typed do this when invoked with a non-prepared statement instead of adding a new method.

tokio-postgres/src/client.rs
ramnivas: Fixed. Earlier thought was to allow access to the raw `RowStream`.

tokio-postgres/src/query.rs
ramnivas: Fixed.

ramnivas:
> Thinking about it, it seems like we could make query_typed do this when invoked with a non-prepared statement instead of adding a new method.

Did you mean `prepare_typed` (there is no `query_typed`). One nice thing about the new `query_with_param_types` is that it takes param values and param types through tuples, so the client cannot specify fewer (or more) types than parameter length.

sfackler:
> Did you mean `prepare_typed` (there is no `query_typed`).

Ah yeah my idea didn't really make much sense :D

sfackler:
Let's rename the method to `query_typed`, which aligns with `prepare_typed` and is a bit shorter.

ramnivas:
> Let's rename the method to `query_typed`, which aligns with `prepare_typed` and is a bit shorter.

Done!

sfackler:
LGTM but there's one small clippy error to fix.

ramnivas:
> LGTM but there's one small clippy error to fix.

Fixed.

sfackler:
Thanks!

ramnivas:
> Thanks!

Thank you!

PR #1151, opened 2024-07-06
feat: sslnegotiation and direct ssl for postgres 17

sunng87:
This patch adds Direct SSL support which is a new feature coming in PostgreSQL 17. With this option, we issue TLS handshake directly without `SslRequest`. This will save a roundtrip on connection opening.

I just follow how libpq is designed, to offer an option `sslnegotiation` to control this behaviour (because it won't work on Postgres 16 or earlier).

### DIFF
--- a/postgres-native-tls/src/lib.rs
+++ b/postgres-native-tls/src/lib.rs
@@ -53,6 +53,7 @@
 //! ```
 #![warn(rust_2018_idioms, clippy::all, missing_docs)]
 
+use native_tls::TlsConnectorBuilder;
 use std::future::Future;
 use std::io;
 use std::pin::Pin;
@@ -179,4 +180,11 @@ where
             None => ChannelBinding::none(),
         }
     }
+}
+
+/// Set ALPN for `TlsConnectorBuilder`
+///
+/// This is required when using `sslnegotiation=direct`
+pub fn set_postgresql_alpn(builder: &mut TlsConnectorBuilder) -> &mut TlsConnectorBuilder {
+    builder.request_alpns(&["postgresql"])
 }
--- a/postgres-native-tls/src/lib.rs
+++ b/postgres-native-tls/src/lib.rs
@@ -185,6 +185,6 @@ where
 /// Set ALPN for `TlsConnectorBuilder`
 ///
 /// This is required when using `sslnegotiation=direct`
-pub fn set_postgresql_alpn(builder: &mut TlsConnectorBuilder) -> &mut TlsConnectorBuilder {
-    builder.request_alpns(&["postgresql"])
+pub fn set_postgresql_alpn(builder: &mut TlsConnectorBuilder) {
+    builder.request_alpns(&["postgresql"]);
 }
--- a/postgres/src/config.rs
+++ b/postgres/src/config.rs
@@ -10,6 +10,7 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::runtime;
+use tokio_postgres::config::SslNegotiation;
 #[doc(inline)]
 pub use tokio_postgres::config::{
     ChannelBinding, Host, LoadBalanceHosts, SslMode, TargetSessionAttrs,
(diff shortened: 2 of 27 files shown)


postgres/src/config.rs
```diff
@@ -10,6 +10,7 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::runtime;
+use tokio_postgres::config::SslNegotiation;
```
sfackler, line 13: This should go in the pub use block below.

tokio-postgres/src/connect_tls.rs
```diff
@@ -25,18 +26,20 @@ where
         SslMode::Prefer | SslMode::Require => {}
```
sfackler, line 26: Should there be any special behavior for `sslmode=prefer sslnegotiation=direct`?

sunng87: No. When sslmode is prefer or required, we go directly to ssl handshake just like other ssl based protocol, for example, https. And those TLS libraries will initialize the handshake process, we don't need to do anything here.

postgres-native-tls/src/lib.rs
```diff
@@ -180,3 +181,10 @@ where
         }
     }
 }
+
+/// Set ALPN for `TlsConnectorBuilder`
+///
+/// This is required when using `sslnegotiation=direct`
+pub fn set_postgresql_alpn(builder: &mut TlsConnectorBuilder) -> &mut TlsConnectorBuilder {
```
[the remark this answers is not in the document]
sunng87: It's only required for direct ssl mode. 

Actually the method is just to provide our user a shortcut for setting ALPN. Because TlsConnectionBuilder is out of our library scope, but ALPN is only writable at this builder level.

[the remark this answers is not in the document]
sfackler: This should return `()` imo.

postgres/src/config.rs
```diff
@@ -40,6 +40,8 @@ use tokio_postgres::{Error, Socket};
 ///     path to the directory containing Unix domain sockets. Otherwise, it is treated as a hostname. Multiple hosts
 ///     can be specified, separated by commas. Each host will be tried in turn when connecting. Required if connecting
 ///     with the `connect` method.
+/// * `sslnegotiation` - TLS negotiation method. If set to `direct`, the client will perform direct TLS handshake, this only works for PostgreSQL 17 and newer.
+///     If set to `postgres`, the default value, it follows original postgres wire protocol to perform the negotiation.
```
sfackler, line 44: I think we should note the ALPN requirement here.

sfackler:
Seems reasonable to me, but I think it might make sense to wait on merging until 17 releases.

sunng87:
No problem. I'm using this to test my own [postgresql wire protocol implementation](https://github.com/sunng87/pgwire/pull/189)

And the clippy issue seems not related to this patch, I can create another one to fix that.

sunng87:
@sfackler Postgres 17 just released

tokio-postgres/src/config.rs
```diff
@@ -50,6 +50,16 @@ pub enum SslMode {
     Require,
 }
 
+/// TLS negotiation configuration
```
marcbowes, line 53: Consider linking to https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLNEGOTIATION

tokio-postgres/src/connect_tls.rs
marcbowes: libpq does:

> psql: error: weak sslmode "prefer" may not be used with sslnegotiation=direct (use "require", "verify-ca", or "verify-full")

tokio-postgres/src/config.rs
```diff
@@ -50,6 +50,16 @@ pub enum SslMode {
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum SslNegotiation {
+    /// Use PostgreSQL SslRequest for Ssl negotiation
+    Postgres,
+    /// Start Ssl handshake without negotiation, only works for PostgreSQL 17+
+    Direct,
+}
```
marcbowes, line 61:
```suggestion
#[derive(Default, Debug, Copy, Clone, PartialEq, Eq)]
#[non_exhaustive]
pub enum SslNegotiation {
    /// Use PostgreSQL SslRequest for Ssl negotiation
    #[default]
    Postgres,
    /// Start Ssl handshake without negotiation, only works for PostgreSQL 17+
    Direct,
}
```

tokio-postgres/src/config.rs
```diff
@@ -106,6 +116,9 @@ pub enum Host {
 ///     path to the directory containing Unix domain sockets. Otherwise, it is treated as a hostname. Multiple hosts
 ///     can be specified, separated by commas. Each host will be tried in turn when connecting. Required if connecting
 ///     with the `connect` method.
+/// * `sslnegotiation` - TLS negotiation method. If set to `direct`, the client will perform direct TLS handshake, this only works for PostgreSQL 17 and newer.
+///     Note that you will need to setup ALPN of TLS client configuration to `postgresql` when using direct TLS.
```
marcbowes, line 120: Consider linking to the methods (or naming them).

marcbowes (approved):
I'd love to see this merged. I made some small suggestions, but I don't think they're blocking.

sunng87:
@marcbowes Thank you for the comments. I have updated the code to address those issues/suggestions.

marcbowes (approved):

marcbowes:
@sfackler I went through this carefully. There are few changes since you last reviewed, and Postgres 17 was released last year. It'd be great to get this shipped.

PR #1153, opened 2024-07-06
impl Debug for Statement

cammeresi:
The lack of this common trait bound caused some unpleasantness. For example, the following didn't compile:

let x = OnceLock::new();
let stmt = db.prepare(...)?;
x.set(stmt).expect(...); // returns Result<(), T=Statement> where T: Debug

### DIFF
--- a/tokio-postgres/src/statement.rs
+++ b/tokio-postgres/src/statement.rs
@@ -61,6 +61,16 @@ impl Statement {
     }
 }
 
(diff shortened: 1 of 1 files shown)


sfackler:
LGTM,  but the specific example in the PR description would probably be more cleanly implemented as `x.set(stmt).ok().expect(...)` since the debug representation of the second statement is probably not going to provide any extra clarity.

PR #1154, opened 2024-07-06
Fix a few nits pointed out by clippy

cammeresi:
- ...::max_value() -> ..::MAX
- delete explicit import of signed integer types


sfackler:
Thanks!

PR #1164, opened 2024-07-22
Add Jiff support

### LINKED ISSUE #1163: Support for jiff datetime types
This is the issue tracker for supporting https://github.com/BurntSushi/jiff datetime types.

PR incoming

allan2:
This PR adds support for version 0.1 of [Jiff](https://github.com/BurntSushi/jiff).

Jiff is inspired by [Temporal](https://tc39.es/proposal-temporal/docs/index.html), a [TC39](https://tc39.es/) proposal to improve datetime handling in JavaScript.

The implementation in *jiff_01.rs* is based on *chrono_04.rs*. When using `FromSql` and `ToSql`, `Zoned` sets the timezone to UTC.
Jiff does not support leap seconds, which is nice because Postgres doesn't either.

Closes #1163


postgres-types/src/jiff_01.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    accepts!(TIMESTAMP);
+    to_sql_checked!();
+}
+
+impl<'a> FromSql<'a> for JiffTimestamp {
+    fn from_sql(type_: &Type, raw: &[u8]) -> Result<JiffTimestamp, Box<dyn Error + Sync + Send>> {
+        Ok(DateTime::from_sql(type_, raw)?
+            .to_zoned(TimeZone::UTC)?
+            .timestamp())
```
BurntSushi, line 39: I'm not intricately familiar with how PostgreSQL represents time, but it seems like you can avoid going through a `DateTime` here and just compute a `JiffTimestamp` directly.

postgres-types/src/jiff_01.rs
```diff
@@ -0,0 +1,118 @@
...
+        Ok(IsNull::No)
+    }
+
+    accepts!(TIMESTAMPTZ);
+    to_sql_checked!();
+}
+
+impl<'a> FromSql<'a> for Zoned {
+    fn from_sql(type_: &Type, raw: &[u8]) -> Result<Zoned, Box<dyn Error + Sync + Send>> {
+        Ok(JiffTimestamp::from_sql(type_, raw)?.to_zoned(TimeZone::UTC))
```
BurntSushi, line 61: Is it correct to assume UTC here? There is no time zone information in PostgreSQL?

I wonder if it makes sense to leave out the impl for `Zoned` and instead require that callers just use a `jiff::Timestamp` and do their own conversions to `Zoned` as needed.

allan2: UTC is correct. From the  [Postgres docs](https://www.postgresql.org/docs/current/datatype-datetime.html):

> For `timestamp` with time zone, the internally stored value is always in UTC... An input value that has an explicit time zone specified is converted to UTC using the appropriate offset for that time zone. If no time zone is stated in the input string, then it is assumed to be in the time zone indicated by the system's [TimeZone](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-TIMEZONE) parameter, and is converted to UTC using the offset for the timezone zone.

Consumers will usually reach for `Timestamp`, but `Zoned` being `ToSql` and `FromSql` can occasionally be convenient. I am in favour of including them. It would also match `chrono::DateTime<FixedOffset>` and `time::OffsetDateTime` having them.

BurntSushi: I'm skeptical that a `Zoned` implementation makes sense here personally. Think of a `Zoned` as "an assertion that a particular datetime is in a specific geographic region." Just because you _can_ pick some kind of default here doesn't mean you should. And the civil datetime types are also somewhat questionable given that you're freely going to and from instants in time and back to inexact time. That also seems wrong.

I understand you're starting from a position of "this is how the Chrono integration works," but Chrono's `DateTime<FixedOffset>` and time's `OffsetDateTime` are much closer to Jiff's `Timestamp` than Jiff's `Zoned`. The closest analog Chrono has to a Jiff `Zoned` is `DateTime<chrono_tz::Tz>`.

The only trait implementation I feel 100% confident about here is for `jiff::Timestamp`.

allan2: I have always felt weird about how Postgres handles time zones and how the `timestamp with time zone` type does not in fact contain the time zone information.  Keeping the actual geographic region requires an additional column. What you wrote about `Zoned` resonates with me when it comes to correctness. I agree that `jiff::Timestamp` matches up with `timestamptz` and `Zoned` does not match with any current Postgres datetime type. I was concerned about conforming to Postgres and this crate's implementation details but I agree with you now on omitting impl for `Zoned`.

I am not sure I agree with your civil type concerns though. I feel like the Postgres types match up quite nicely for these inexact expressions of time

(Postgres also has a `timetz`. Can we express that in Jiff?)

postgres-types/src/jiff_01.rs
allan2: Addressed in [f00d208](https://github.com/sfackler/rust-postgres/pull/1164/commits/f00d208959c8c76c9bbe943f53a9a261ef1d2315)

postgres-types/src/jiff_01.rs
allan2: `timestamp` in Postgres is *inexact*, being a combination of `date` and `time` without time zone. It can be used to record the general idea of the 2025 New Year's Countdown, taking place on Dec 31, 2024 at 11:59pm, without regard for the geographic location.

`timestamptz` is exact., asserting the time zone (albeit in UTC only).

BurntSushi: > timestamp in Postgres is inexact, being a combination of date and time without time zone.

Whoa. OK. Let me backup and read their docs more carefully... Because if a `timestamp` is just inexact time and `timestamptz` is just inexact time with a time zone, then that does _not_ make a `timestamptz` exact. It could be ambiguous. But [according to this](https://wiki.postgresql.org/wiki/Don%27t_Do_This#Date.2FTime_storage), yes, PostgreSQL is saying that it is exact. So maybe they are doing disambiguation on insert or something. And also, this seems to indeed agree that just a `timestamp` is actually inexact time:

> timestamp (also known as timestamp without time zone) doesn't do any of that, it just stores a date and time you give it. You can think of it being a picture of a calendar and a clock rather than a point in time. Without additional information - the timezone - you don't know what time it records.

Wild.

I had assumed that `timestamp` was, well, just a Unix timestamp. Which is always in UTC. And `timestamptz` stored some extra bit of information like an offset. OK, so I think this is the mapping then:

* `date` -> `jiff::civil::Date`
* `time` -> `jiff::civil::Time`
* `timestamp` -> `jiff::civil::DateTime`
* `timestamptz` -> `jiff::Timestamp`

I think where I got confused is that in this patch, the `timestamp -> jiff::civil::DateTime` is implemented by using what appears to be a representation for exact time (number of seconds since an epoch). But I guess that's just a representational thing.

allan2: > I had assumed that `timestamp` was, well, just a Unix timestamp. Which is always in UTC. And `timestamptz` stored some extra bit of information like an offset. OK, so I think this is the mapping then:
> 
> * `date` -> `jiff::civil::Date`
> * `time` -> `jiff::civil::Time`
> * `timestamp` -> `jiff::civil::DateTime`
> * `timestamptz` -> `jiff::Timestamp`
> 
> I think where I got confused is that in this patch, the `timestamp -> jiff::civil::DateTime` is implemented by using what appears to be a representation for exact time (number of seconds since an epoch). But I guess that's just a representational thing.

It's certainly confusing.

`timestamp` and `timestamptz` both internally store microseconds since Y2K midnight UTC. The difference is that `timestamptz` converts the input to UTC before storing the number of microseconds. It bakes in the offset so we get the correct exact time, but the actual source time zone/offset is discarded.

For the record, `DateTime` and `Timestamp` are much better names than `timestamp` and `timestamptz` to describe what they really are.

> And yes, it looks like the above mapping lines up with what you have in this PR. OK, sorry for the noise!

Comments were much appreciated :) At least we're only dealing with Terran time.

sfackler:
Can you also add some tests? Here are chrono's for reference: https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/tests/test/types/chrono_04.rs

allan2:
This PR is complete and ready for review. All tests are passing.

One minor note:

In the tests, I use jiff's `FromStr` implementations to parse strings into date types. <sup>[[1]](https://docs.rs/jiff/latest/jiff/fmt/index.html)</sup>

```rs
let s = "'1970-01-01 00:00:00.010000000Z'";
let ts: Timestamp = s.trim_matches('\'').parse().unwrap();  // I did this
let ts: Timestamp = Timestamp::strptime("'%Y-%m-%d %H:%M:%S.%f %#z'", s); // I could do this if you prefer
```

Special thanks to @cmarkh who opened [allan2/rust-postgres#1](https://github.com/allan2/rust-postgres/pull/1), which I unfortunately missed.

BurntSushi:
The FromStr impl should be used if possible IMO. It's more flexible (it permits a `T` separator between the date and time) and probably 2x faster since it's a dedicated parser.

sfackler:
Thanks!

PR #1165, opened 2024-07-22
For `query_typed`, deal with the no-data case.

ramnivas:
If a query returns no data, we receive `Message::NoData`, which signals the completion of the query. However, we treated it as a no-op, leading to processing other messages and eventual failure.

This PR fixes the issue and updates the `query_typed` tests to cover this scenario.


sfackler:
Thanks!

ramnivas:
Thanks!

PR #1170, opened 2024-08-15
Fix `chrono::DateTime<Utc>` tests

allan2:
`test_date_time_params` and `test_with_special_date_time_params` could fail when run with a non-UTC system time zone.
I encountered these failing tests when writing the tests for jiff.

In Postgres, if no time zone is stated in the input string, then it is assumed to be in system time.<sup>[1]</sup> This means that in these tests, a correctly parsed `DateTime<Utc>` could be compared to an incorrectly offset time.

The offset component is now included in the test strings of these tests. This component is already present in the test strings for `time::OffsetDateTime`.

<sup>[1]</sup> [Postgres docs](https://www.postgresql.org/docs/current/datatype-datetime.html):

>For timestamp with time zone, the internally stored value is always in UTC (Universal Coordinated Time, traditionally known as Greenwich Mean Time, GMT). An input value that has an explicit time zone specified is converted to UTC using the appropriate offset for that time zone. If no time zone is stated in the input string, then it is assumed to be in the time zone indicated by the system's [TimeZone](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-TIMEZONE) parameter, and is converted to UTC using the offset for the timezone zone.

### DIFF
--- a/tokio-postgres/tests/test/types/chrono_04.rs
+++ b/tokio-postgres/tests/test/types/chrono_04.rs
@@ -1,4 +1,4 @@
-use chrono_04::{DateTime, NaiveDate, NaiveDateTime, NaiveTime, TimeZone, Utc};
+use chrono_04::{DateTime, NaiveDate, NaiveDateTime, NaiveTime, Utc};
 use std::fmt;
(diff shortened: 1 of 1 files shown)


sfackler:
Thanks!

PR #1185, opened 2024-11-01
Make error::Kind public

levkk:
Make `error::Kind` public and provide a getter. This allows lib/app code to handle different kind of errors, e.g.

```rust
match err.kind() {
  Kind::Column(name) => panic!("You forgot to include column {} into the query", name),
  Kind::Parameter(_, _) => panic!("You forgot to pass a parameter to the prepared statement"),
  _kind=> panic!("{:?}", kind),
}
```

I documented errors I know about, but the ones I don't know what they do I left undocumented, hiding the doc warning.

### DIFF
--- a/tokio-postgres/src/error/mod.rs
+++ b/tokio-postgres/src/error/mod.rs
@@ -337,14 +337,20 @@ pub enum ErrorPosition {
     },
 }
 
-#[derive(Debug, PartialEq)]
-enum Kind {
+/// Error kind returned by the database.
+#[allow(missing_docs)]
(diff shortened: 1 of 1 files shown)


BaptisteRoseau:
Also made it here https://github.com/sfackler/rust-postgres/pull/1138 with a few test cases but it seems like it won't be merge any time soon..

Aceeri:
This is something I also would love, I've just ended up doing `err.to_string()` and re-parsing it which is pretty counterproductive.

paolobarbolini (changes requested):

PR #1205, opened 2025-02-01
fix rand 0.9 deprecations

kristof-mattei:
- **Update rand requirement from 0.8 to 0.9**
- **chore: addressed rand 0.9's deprecations**


sfackler:
It looks like the wasm test needs to have a `RUSTFLAGS=--cfg=wasm_js` added.

kristof-mattei:
> It looks like the wasm test needs to have a `RUSTFLAGS=--cfg=wasm_js` added.

Oops. I fixed that in another commit, which ended up not being pushed. Sorry.

PR #1206, opened 2025-02-01
Getrandom 0.3.1

kristof-mattei:
#1205 should go first, because `rand@0.9` needs `getrandom@0.3.1`, and doing `getrandom` only makes `rand` being included twice.

Notice I've had to update the `RUSTFLAGS` for the wasm build. This means that customers of this crate who use wasm must do the same. It might warrant an update in the docs. 

If you don't, fails like this:

```
error: The wasm32-unknown-unknown targets are not supported by default; you may need to enable the "wasm_js" configuration flag. Note that enabling the `wasm_js` feature flag alone is insufficient. For more information see: https://docs.rs/getrandom/#webassembly-support
   --> /home/kristof/.cargo/registry/src/index.crates.io-6f17d22bba15001f/getrandom-0.3.1/src/backends.rs:159:9
    |
159 | /         compile_error!(
160 | |             "The wasm32-unknown-unknown targets are not supported by default; \
161 | |             you may need to enable the \"wasm_js\" configuration flag. Note \
162 | |             that enabling the `wasm_js` feature flag alone is insufficient. \
163 | |             For more information see: \
164 | |             https://docs.rs/getrandom/#webassembly-support"
165 | |         );
    | |_________^
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,8 @@ jobs:
   check-wasm32:
     name: check-wasm32
     runs-on: ubuntu-latest
+    env:
+        RUSTFLAGS: ${env.RUSTFLAGS} --cfg getrandom_backend="wasm_js"
     steps:
       - uses: actions/checkout@v3
       - uses: sfackler/actions/rustup@master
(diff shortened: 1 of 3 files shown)


sfackler:
Folded into #1201

PR #1215, opened 2025-02-12
Only call set_tcp_user_timeout when enabled

### LINKED ISSUE #1214: Fails to Connect in WSL 1 Due to TCP User Timeout (os error 92)
### Description:
Hello, 
After upgrading Prisma from 5.19.1 to 6.3.0, I encountered an issue where Prisma could no longer connect to the database in WSL 1.
```bash
Error: P1001: Can't reach database server at `localhost:5432`
```

To isolate the issue, I tested it directly with the tokio-postgres library and observed the following error message:
```bash
Failed to connect to database: error connecting to server: Protocol not available (os error 92)
```
This issue is not limited to localhost (localhost:5432)—it also occurs when connecting to any valid PostgreSQL URL, making database connections completely unusable in WSL 1.
Upon investigation, the issue is very likely caused by the use of TCP user timeout, which may not be supported in WSL 1. However, the existing approach of setting TCP  keepalive works fine.
The relevant code can be found here: [connect_socket.rs#L33](https://github.com/sfackler/rust-postgres/blob/1e1f6bff755ce8d05e300ab2d060dbced41e390a/tokio-postgres/src/connect_socket.rs#L33)

### Regression Details:
- This change was introduced in Prisma 5.22.0, when **tokio-postgres** was upgraded from 0.7.7 to 0.7.12.
- Before Prisma 5.22.0, database connections in WSL 1 worked as expected.

### Steps to Reproduce:
1. Use **tokio-postgres** in a WSL 1 environment
2. Attempt to connect to a PostgreSQL database running on localhost:5432
3. Observe the connection failure with os error 92

### Expected Behavior:
**tokio-postgres** should be able to connect to the database in WSL 1, as it did in previous versions.

### Environment:
- Library: tokio-postgres
- WSL Version: 1
- Distro: Ubuntu 22.04.5 LTS
- Database: PostgreSQL v16
- Operating System: Windows 10

Thanks! Please let me know if I can provide any further details.

sfackler:
This avoids unnecessary errors in environments that don't support the feature.

Closes #1214

### DIFF
--- a/tokio-postgres/src/connect_socket.rs
+++ b/tokio-postgres/src/connect_socket.rs
@@ -27,10 +27,11 @@ pub(crate) async fn connect_socket(
             stream.set_nodelay(true).map_err(Error::connect)?;
 
             let sock_ref = SockRef::from(&stream);
+
             #[cfg(target_os = "linux")]
-            {
+            if let Some(tcp_user_timeout) = tcp_user_timeout {
                 sock_ref
-                    .set_tcp_user_timeout(tcp_user_timeout)
+                    .set_tcp_user_timeout(Some(tcp_user_timeout))
                     .map_err(Error::connect)?;
             }
 


sfackler:
@swnoh can you confirm that this fixes things for you?

swnoh:
Superb! Thanks for the quick fix! I just tested it, and it works perfectly in WSL 1. Actually, I had already built a custom engine as a workaround, but this will be really helpful since I won’t need it once the fix is included in a future version. Really appreciate your help!

PR #1218, opened 2025-02-18
impl ToStatement for String → impl<T: Sealed + AsRef<str>> ToStatemen…

i18nsite:
as title, for easy use


paolobarbolini:
This seems very broad, and it's also a breaking change. What string type are you trying to use this with?

PR #1220, opened 2025-02-20
Add target_session_attrs Modes

### LINKED ISSUE #1211: Support New Target Session Attributes
Currently libpq offers options for `primary`, `standby`, and `prefer-standby`. Looking at the codebase it looks like rust-postgres currently does not support these modes.

https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-TARGET-SESSION-ATTRS

AndrewJackson2020:
This PR attempts to add Primary, Standby, and PreferStandby modes to rust-postgres. This still needs some work but raising it as a draft for now.

Attempts to close #1211.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,3 +8,27 @@ services:
       - ./docker/sql_setup.sh:/docker-entrypoint-initdb.d/sql_setup.sh
     environment:
       POSTGRES_PASSWORD: postgres
+  postgres_replica:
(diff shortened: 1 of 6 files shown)


AndrewJackson2020:
This should be good to review. Added some test coverage and implemented added a read replica postgres container to the docker compose setup to facilitate further testing.

PR #1229, opened 2025-03-26
improve is_closed effectiveness

conradludgate:
Hi. We noticed that `is_closed()` is not always super effective, especially in the `postgres` crate. I've added a couple utilities to try and improve this state.

Problem 1: `is_closed()` checks if the channel is closed, not if the connection is closed. These can be the same, but currently the channel is only closed when `Connection` is dropped. In `postgres::Client`, `Connection` never gets dropped. Fix is to force close the receiver on terminal state.

Problem 2: The connection will only turn if something happens. Sometimes closure can be silent, so even still it seems open. You only can check it's valid if you try to run a query or if the TCP keepalive triggers. Running a `SELECT 1;` is a common workaround, but in my understanding, a single `SYNC` message is cheaper and doesn't cause any logging - so it's a better liveness check than running a full query.

Thanks for your consideration

### DIFF
--- a/postgres/src/client.rs
+++ b/postgres/src/client.rs
@@ -487,6 +487,11 @@ impl Client {
         self.connection.block_on(self.client.batch_execute(query))
     }
 
+    /// Check the connection is alive and wait for the confirmation.
+    pub fn check_connection(&mut self) -> Result<(), Error> {
+        self.connection.block_on(self.client.check_connection())
+    }
+
     /// Begins a new database transaction.
     ///
     /// The transaction will roll back by default - use the `commit` method to commit it.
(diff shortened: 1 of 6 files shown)


postgres/src/client.rs
```diff
@@ -487,6 +487,11 @@ impl Client {
         self.connection.block_on(self.client.batch_execute(query))
     }
 
+    /// Check the connection is alive and wait for the confirmation.
```
tristan957, line 490:
```suggestion
    /// Check that the connection is alive and wait for the confirmation.
```

tokio-postgres/src/client.rs
```diff
@@ -531,6 +531,12 @@ impl Client {
         simple_query::batch_execute(self.inner(), query).await
     }
 
+    /// Check the connection is alive and wait for the confirmation.
```
tristan957, line 534:
```suggestion
    /// Check that the connection is alive and wait for the confirmation.
```

paolobarbolini (approved):
Thanks for the PR!

PR #1230, opened 2025-04-02
expose sync rollback as a private api method

fredr:
Don't know if this is the correct approach, I just saw that sync close was exposed the same way.

It would be useful to have this exposed so that one can implement their own transaction type with a drop implementation that rolls back.

The reason I would like to implement my own transaction type is that I want to expose it to nodejs via napi-rs, and it is not possible to share things that have a lifetime over that boundary.

### DIFF
--- a/tokio-postgres/src/client.rs
+++ b/tokio-postgres/src/client.rs
@@ -1,4 +1,4 @@
-use crate::codec::BackendMessages;
+use crate::codec::{BackendMessages, FrontendMessage};
 use crate::config::{SslMode, SslNegotiation};
 use crate::connection::{Request, RequestMessages};
 use crate::copy_out::CopyOutStream;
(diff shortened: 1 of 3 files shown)


tokio-postgres/src/client.rs
```diff
@@ -605,6 +606,21 @@ impl Client {
...
+                frontend::query(&format!("ROLLBACK TO {}", name), buf).unwrap();
+            } else {
+                frontend::query("ROLLBACK", buf).unwrap();
+            }
+            buf.split().freeze()
+        });
+        let _ = self
+            .inner()
+            .send(RequestMessages::Single(FrontendMessage::Raw(buf)));
+    }
```
paolobarbolini, line 623: NIT: I'd move this above `__private_api_close`

paolobarbolini (changes requested):
Sounds like a good idea.

fredr:
> Sounds like a good idea.

Thanks for reviewing! I've addressed your comment

paolobarbolini (approved):

PR #1231, opened 2025-04-08
postgres-protocol: add pg numeric type

fredr:
A first stab at adding support for the Numeric type as well as a to/from implementation for string as that is kind of the only std type that can hold a numeric.

Would appreciate if you had the time to review this. Happy to fix any feedback. I'm thinking any ToSql and FromSql implementations can be added as followups to this one.

The protocol is quite strait forward, the string conversion is more finicky, but added a bunch of test to hopefully catch all corner cases. Let me know if you think there is any tests that should be added for those.

### DIFF
--- a/postgres-protocol/src/types/numeric.rs
+++ b/postgres-protocol/src/types/numeric.rs
@@ -1,0 +1,804 @@
+//! Conversions to and from Postgres's binary format for the numeric type.
+use byteorder::{BigEndian, ReadBytesExt};
+use bytes::{BufMut, BytesMut};
+use std::boxed::Box as StdBox;
+use std::collections::VecDeque;
+use std::error::Error;
+use std::str::{self, FromStr};
+
+/// Serializes a `NUMERIC` value.
+#[inline]
+pub fn numeric_to_sql(v: Numeric, buf: &mut BytesMut) {
+    let num_digits = v.digits.len() as u16;
+    buf.put_u16(num_digits);
+    buf.put_i16(v.weight);
+    buf.put_u16(v.sign.into_u16());
+    buf.put_u16(v.scale);
+
+    for digit in v.digits {
+        buf.put_i16(digit);
+    }
+}
+
+/// Deserializes a `NUMERIC` value.
+#[inline]
+pub fn numeric_from_sql(mut buf: &[u8]) -> Result<Numeric, StdBox<dyn Error + Sync + Send>> {
+    let num_digits = buf.read_u16::<BigEndian>()?;
+    let mut digits = Vec::with_capacity(num_digits.into());
+
+    let weight = buf.read_i16::<BigEndian>()?;
+    let sign = NumericSign::try_from_u16(buf.read_u16::<BigEndian>()?)?;
+
+    let scale = buf.read_u16::<BigEndian>()?;
+
+    for _ in 0..num_digits {
+        digits.push(buf.read_i16::<BigEndian>()?);
+    }
+
+    Ok(Numeric {
+        sign,
+        scale,
+        weight,
+        digits,
+    })
+}
+
+/// A Posgres numeric
+#[derive(Debug, PartialEq, Eq)]
+pub struct Numeric {
+    sign: NumericSign,
+    scale: u16,
+    weight: i16,
+    digits: Vec<i16>,
+}
+
+impl Numeric {
+    /// Returns the number of digits.
+    #[inline]
+    pub fn num_digits(&self) -> usize {
+        self.digits.len()
+    }
+
+    /// Returns the weight of the numeric value.
+    #[inline]
+    pub fn weight(&self) -> i16 {
+        self.weight
+    }
+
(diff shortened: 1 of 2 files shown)


postgres-protocol/src/types/numeric.rs
```diff
@@ -0,0 +1,1583 @@
...
+    if leading_zero_count > 0 {
+        if leading_zero_count == digits.len() {
+            // All digits are zero
+            *weight = 0;
+            digits.clear();
+        } else {
+            // Remove leading zeros
+            digits.drain(0..leading_zero_count);
+            *weight -= leading_zero_count as i16;
+        }
```
paolobarbolini: Is this micro-optimization worth it over always going through `drain`?

postgres-protocol/src/types/numeric.rs
```diff
@@ -0,0 +1,1583 @@
...
+fn split_e(s: &[u8]) -> (&[u8], Option<&[u8]>) {
+    let mut s = s.splitn(2, |&b| b == b'e' || b == b'E');
+    let first = s.next().unwrap();
+    let second = s.next();
+    (first, second)
+}
+
+fn split_decimal(s: &[u8]) -> (&[u8], Option<&[u8]>) {
+    let mut s = s.splitn(2, |&b| b == b'.');
+    let first = s.next().unwrap();
```
paolobarbolini: Quickly scrolling through the code it's non-obvious that this panic is unreachable.

postgres-protocol/src/types/numeric.rs
```diff
@@ -0,0 +1,1583 @@
...
+        if e.is_empty() {
+            return Err("empty scientific notation exponent".into());
+        }
+
+        for &b in e {
+            if !b.is_ascii_digit() {
+                return Err("scientific notation string contain non-digit character".into());
+            }
+            // Prevent integer overflow in exponent
+            exp = exp.saturating_mul(10).saturating_add((b - b'0') as i32);
```
paolobarbolini: Can this really overflow if we're checking for `exp` not to get too big in the next line?

postgres-protocol/src/types/numeric.rs
```diff
@@ -0,0 +1,1583 @@
...
+            let numeric = Numeric::from_str(&prop_val).expect("parse numeric");
+            let str_val = numeric.to_string();
+            assert_eq!(prop_val, str_val, "proprty test value {value}");
+        }
+        #[test]
+        fn test_arbitrary_i64_from_string_and_back(value in any::<i64>()) {
+            let prop_val = value.to_string();
+            let numeric = Numeric::from_str(&prop_val).expect("parse numeric");
+            let str_val = numeric.to_string();
+            assert_eq!(prop_val, str_val, "proprty test value {value}");
```
paolobarbolini:
```suggestion
            assert_eq!(prop_val, str_val, "property test value {value}");
```

postgres-protocol/src/types/numeric.rs
```diff
@@ -0,0 +1,1583 @@
...
+            NumericSign::Positive => 0x0000,
+            NumericSign::Negative => 0x4000,
+            NumericSign::NaN => 0xC000,
+            NumericSign::PositiveInfinity => 0xD000,
+            NumericSign::NegativeInfinity => 0xF000,
+        }
+    }
+}
+
+/// A Posgres numeric
```
paolobarbolini: Copies the description from `PgLsn`.

```suggestion
/// Postgres `NUMERIC` type.
```

paolobarbolini (commented):
I've left a couple of comments, including one on how you see this get used.

The final version should also have tests in `tokio-postgres/tests/test/types/` and commits squashed.

postgres-protocol/src/types/numeric.rs
```diff
@@ -0,0 +1,1583 @@
...
+        return Ok(Some(Numeric::nan()));
+    }
+    if s.eq_ignore_ascii_case("Infinity") || s.eq_ignore_ascii_case("Inf") {
+        return Ok(Some(Numeric::infinity()));
+    }
+    if s.eq_ignore_ascii_case("-Infinity") || s.eq_ignore_ascii_case("-Inf") {
+        return Ok(Some(Numeric::negative_infinity()));
+    }
+    Ok(None)
+}
```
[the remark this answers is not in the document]
fredr: good catch!

postgres-protocol/src/types/numeric.rs
fredr: very true, good catch!

fredr:
Thanks for reviewing this! I'll wait with squashing until we are happy with the PR. Regarding tests in `tokio-postgres`, from what I can tell, those are more tests of types that implement `ToSql` and `FromSql`, so not sure exactly what I should add for tests for this pr? I assume when we start adding `bigdecimal` and other types, those will be tested there, but not the protocol type? or should I add something there that uses `FromStr` and `Display`?

PR #1234, opened 2025-04-11
Fix latest clippy warnings

paolobarbolini:
`div_ceil` is from Rust v1.73, `io::Error::other` from Rust v1.74.

I've allowed `clippy::doc_overindented_list_items` for now because it made the markdown more readable.


sfackler:
Thanks!

PR #1238, opened 2025-05-07
Chrono TimeDelta support

mshauneu:
Add Chrono TimeDelta support

### DIFF
--- a/postgres-protocol/src/types/mod.rs
+++ b/postgres-protocol/src/types/mod.rs
@@ -421,6 +421,27 @@ pub fn time_from_sql(mut buf: &[u8]) -> Result<i64, StdBox<dyn Error + Sync + Se
     Ok(v)
 }
 
+/// Serializes an `INTERVAL` value.
+///
+/// The value should represent the number of microseconds.
+#[inline]
+pub fn interval_to_sql(v: i128, buf: &mut BytesMut) {
+    buf.put_i128(v);
+}
+
+/// Deserializes an `INTERVAL` value.
+///
+/// The value represents the number of microseconds.
+#[inline]
+pub fn interval_from_sql(mut buf: &[u8]) -> Result<i128, StdBox<dyn Error + Sync + Send>> {
+    let v = buf.read_i128::<BigEndian>()?;
+    if !buf.is_empty() {
+        return Err("invalid message length: interval not drained".into());
+    }
+    Ok(v)
+}
+
+
 /// Serializes a `MACADDR` value.
 #[inline]
 pub fn macaddr_to_sql(v: [u8; 6], buf: &mut BytesMut) {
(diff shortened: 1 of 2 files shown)


paolobarbolini:
I don't think you've tested this. The build is broken:

```
❯ cargo check --features with-chrono-0_4
    Checking postgres-types v0.2.9 (/tmp/fuck/rust-postgres/postgres-types)
error[E0425]: cannot find value `jd` in this scope
   --> postgres-types/src/chrono_04.rs:166:51
    |
166 |         if usec > i128::from(i64::max_value()) || jd < i128::from(i64::min_value()) {
    |                                                   ^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
error: could not compile `postgres-types` (lib) due to 1 previous error
```

If I fix it with:

```diff
diff --git a/postgres-types/src/chrono_04.rs b/postgres-types/src/chrono_04.rs
index 74df5373..72ed30f3 100644
--- a/postgres-types/src/chrono_04.rs
+++ b/postgres-types/src/chrono_04.rs
@@ -163,7 +163,7 @@ impl ToSql for NaiveTime {
 impl<'a> FromSql<'a> for TimeDelta {
     fn from_sql(_: &Type, raw: &[u8]) -> Result<TimeDelta, Box<dyn Error + Sync + Send>> {
         let usec = types::interval_from_sql(raw)?;
-        if usec > i128::from(i64::max_value()) || jd < i128::from(i64::min_value()) {
+        if usec > i128::from(i64::max_value()) || usec < i128::from(i64::min_value()) {
             return Err("value too large to transmit".into());
         }
         Ok(TimeDelta::microseconds(usec as i64))
```

...and then test it with this:

```rust
#!/usr/bin/env cargo

---cargo
[dependencies]
tokio = { version = "1", features = ["rt", "macros"] }
tokio-postgres = { git = "https://github.com/mshauneu/rust-postgres.git", features = ["with-chrono-0_4"] }
chrono = { version = "0.4", default-features = false }
---

#[tokio::main(flavor = "current_thread")]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let (client, connection) =
        tokio_postgres::connect("host=localhost port=5433 user=postgres", tokio_postgres::NoTls).await?;

    tokio::spawn(async move {
        if let Err(e) = connection.await {
            eprintln!("connection error: {}", e);
        }
    });

    let rows = client
    .query("SELECT INTERVAL '1 day'", &[])
    .await?;

    // And then check that we got back the same string we sent over.
    let value: chrono::TimeDelta = rows[0].get(0);
    assert_eq!(value, chrono::TimeDelta::days(1));

    Ok(())
}
```

It fails with:

```
thread 'main' (1658335) panicked at pr1238.rs:28:5:
assertion `left == right` failed
  left: TimeDelta { secs: 4294, nanos: 967296000 }
 right: TimeDelta { secs: 86400, nanos: 0 }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

---

Also `chrono::TimeDelta` doesn't seem to be right type to represent a PostgreSQL `INTERVAL`. See https://github.com/rust-postgres/rust-postgres/issues/60#issuecomment-57099402.

reneleonhardt:
CI checks are not allowed to run automatically.

Why did you immediately close the pull request without giving the contributor a chance to react and apply the review comment after waiting for 4 months until the first review?

PR #1252, opened 2025-06-25
Remove mentions of require-full and require-ca

maxammann:
Both of the options are not valid in rust-postgres and they are not supported.

Actually rust-postgres deviates from postgres docs by checking the certificate when specifying "require".

related: https://github.com/sfackler/rust-postgres/pull/988

Would it make sense to parse "require-full" into SslMode:Require? That would make it compatible with normal postgres connection URLs

cc @nyurik hi! nice to see you here!


paolobarbolini (approved):
Good catch!

PR #1262, opened 2025-08-16
fix an Error/Display impl

CobaltCause:
Types implementing the Error trait should print some inner Error **xor** return that inner Error from its Error::source impl. Otherwise, the same error will be printed multiple times when printing an Error's entire source chain, which is unhelpful.


paolobarbolini (approved):

paolobarbolini:
Some tests need to be fixed.

CobaltCause:
Done.

Also FWIW, and I'm not 100% sure, but it seems like there might be another case of this kind of thing here: https://github.com/rust-postgres/rust-postgres/blob/bbd13014eda04797d9c794515829f4da4495e425/postgres-openssl/src/lib.rs#L169-L192

PR #1291, opened 2025-11-03
Add a function to read the inner value of sql types.

rydb:
Currently, there is no way to know the actual type of a sql value because inner is private. This pr adds a .inner() function to expose the read value of inner to fix this.

### DIFF
--- a/postgres-types/src/lib.rs
+++ b/postgres-types/src/lib.rs
@@ -368,6 +368,11 @@ impl Type {
     pub fn name(&self) -> &str {
         self.0.name()
     }
+
+    /// Returns the inner value of this type.
+    pub fn inner(&self) -> &Inner {
+        &self.0
+    }
 }
 
 /// Represents the kind of a Postgres type.


sfackler:
If it were possible to exhaustively match against all Inner variants, it would be a breaking change to add new types that show up in new Postgres releases.

If I wanted to enable that API, I would just make `Type` an enum directly.

rydb:
> If it were possible to exhaustively match against all Inner variants, it would be a breaking change to add new types that show up in new Postgres releases.
> 
> If I wanted to enable that API, I would just make `Type` an enum directly.

The option to put `_, =>` is not voided with this change though? The option as an end user for new types to be a breaking change is desirable as well. If there are new variants, I want to be notified as a developer when I need to add new cases to the match statement rather then have it show up as a bug report later.

As things are currently structured, when breaking changes do occur, updating match statements would require manually adding the specific variants, rather then just clicking `auto-fix`, and either moving matches into the newly named/reformatted match blocks, or replacing the todo!()s with impls.

sfackler:
What is a use case for matching against every built-in postgres type?

rydb:
> What is a use case for matching against every built-in postgres type?

I want to convert untyped sql structs into string variants since there is no .to_string() function for column values.

```rust

pub fn try_convert_to_string_from_sql(row: &Row, i: usize) -> Result<String, ConversionError> {
    let sql_type = row.columns()[i].type_();
    
    let value: Option<Result<String, Error>> = match *sql_type {
        Type::INT4 => {
            Some(row.try_get::<_, i32>(i).map(|n| n.to_string()))
        },
        Type::VARCHAR => {
            Some(row.try_get::<_, &str>(i).map(|n| n.to_string()))
        }
        _ => None
    };

    if let Some(value) = value {
        let value = match value {
            Ok(value) => Ok(value),
            Err(err) => Err(err.into()),
        };
        return value
    } else {
        return Err(ConversionError::Unimplemented(format!("CONVERSION UNIMPLMENTED FOR: {:#}", sql_type)))

    }
}
```

I would like the option to ensure the above, at some point, doesn't use `_ =>`, and instead, implements all of the available sql types for string conversions. However, removing it causes.

```
non-exhaustive patterns: `Type(postgres_types::type_gen::Inner::Bool)`, `Type(postgres_types::type_gen::Inner::Bytea)`, `Type(postgres_types::type_gen::Inner::Char)`

and 181 more not covered
```
but because `Type` is not an enum, if I did decide I wanted to add the rest of the string impls, thats 181 `Type::<type> => { }`s that I have to manually type by hand because quick fix doesn't work for them.

sfackler:
What would your match look like for these types: `VOID`, `TRIGGER`, `INTERNAL`, `ANYCOMPATIBLEMULTIRANGE`, and `INDEX_AM_HANDLER`?

rydb:
> What would your match look like for these types: `VOID`, `TRIGGER`, `INTERNAL`, `ANYCOMPATIBLEMULTIRANGE`, and `INDEX_AM_HANDLER`?

depends on the context. For string conversion, it would just be the names of the types them selves.

PR #1298, opened 2025-11-23
Add smol_str 0.2 and 0.3 support (continued #1095)

JakubKoralewski:
Continued @novacrazy's #1095

I can commit half an hour to make any changes, but pushing (use the Force maintainers) to this branch would be appreciated.


tokio-postgres/src/lib.rs
```diff
@@ -97,30 +97,17 @@
 //! `postgres-openssl` and `postgres-native-tls` crates provide implementations backed by the `openssl` and `native-tls`
 //! crates, respectively.
 //!
-//! # Features
+//! 
+#![doc=include_str!("../../postgres-types/feature_docs.md")]
```
paolobarbolini: ...same here

paolobarbolini (changes requested):
Thanks for taking over #1095.

If you can could you split individual changes, like reordering items in `Cargo.toml` and moving features to a shared `.md` file, into separate commits?

PR #1304, opened 2025-12-04
Implement debug for public objects

latot:
Hi all! The motivation for this PR is be able to use tracing on projects with this crates, ```instrument``` macro requires to the objects be ```Debug```.

The start point is use ```#![deny(missing_debug_implementations)]``` on all crates, we only need ```Debug``` for the exposed items of each crate.

The only one I do not implemented this for talk is ```postgres-ssl``` because it uses a lot of the ```openssl``` crate, which are mainly bindings and most of object in ```postgres-ssl``` have very few fields and uses them... so I do not know how to properly implement ```Debug```.

While for the other crates, in its majority there was no issue implementing ```Debug```, but there is some edge cases:

- Fn traits: This ones can not be ```Debug``` so I choose to implement all other fields and ignore the ones that has functions/closures.
- Buf/BufMut: This traits from the ```bytes``` crate do come with the ```Debug``` trait, so I make a new ones in ```postgres-protocol::types::debuf_bytes``` there is the supertraits which are the same ones but with ```Debug```, are also implemented automatically, I changed a little some modules to use this new traits, still I need to call the ones from bytes as ```_Buf``` to keep available its functionalities.

So my two remaining points rn are:

- Is acceptable handle closure/functions like this?
- how to implement ```Debug``` on ```postgres-ssl```?


paolobarbolini:
Adding `+ Debug` around the code makes this into a breaking change (also `ItemConnection`), which I'd like to avoid for now, so I'd revert those for now and where necessary replace the derived `Debug` implementations with manual ones that skip fields which may not implement `Debug`.

Other than that this looks like a very nice improvement.

PR #1340, opened 2026-04-18
add schema support in derive

### LINKED ISSUE #627: Postgres enums within schema
The derived `FromSql` instance appears to fail when translating a Rust type to a Postgres type qualified by a schema name.  

For example, this works fine:

```rust
  #[derive(Clone, Copy, Debug, PartialEq, Eq, Deserialize, Serialize, FromSql, ToSql)]
  #[postgres(name = "service")]
  pub enum Service {
    Facebook,
    Google,
    Twitter,
  }
```

But this does not, the only difference being the "users." schema prefix:

```rust
  #[derive(Clone, Copy, Debug, PartialEq, Eq, Deserialize, Serialize, FromSql, ToSql)]
  #[postgres(name = "users.service")]
  pub enum Service {
    Facebook,
    Google,
    Twitter,
  }
```

The Postgres types in each case would correspond, i.e.:

```pgsql
CREATE TYPE [users.]service AS ENUM (
  'Facebook',
  'Google',
  'Twitter'
);
```

Perhaps though there is a way to use the qualified Postgres type that I've overlooked?

purple-ic:
`#[postgres(schema = ...)]` can be used to make it so the type only accepts types with the matching schema name. if the attribute is omitted, any schema is allowed (same as before).

unfortunately this adds tests that create non-temporary types (because we use our own schemas, and afaik temporary things are only possible under `pg_temp_*` schemas). the tests fail if you run `cargo test` twice with the same database (though it seems there were already a few tests with this issue). lmk if there's a fix.

also for whatever reason the `postgres::test::is_closed` test fails on my machine? i can only assume this is a bug with the test suite or some issue with my setup, because i touched absolutely nothing that would affect that...

fixes #627


purple-ic:
actually hold on: i guess i could make the schemas use a `pg_temp_` prefix

purple-ic:
ah, nevermind, you can't create your own `pg_` schemas. the issue remains

PR #1353, opened 2026-05-28
feat(gaussdb-mcp): defer keychain read until first tool invocation

c2j:
## Summary

Defer OS keychain access from gaussdb-mcp process startup to the first MCP tool call. The binary now starts instantly without triggering macOS Keychain prompts, and only reads credentials when a database operation is actually requested.

## Changes

- **`ConnectionState::Pending`**: New state variant holding a lazy URL resolver closure that reads the keychain only on first tool invocation
- **`LazyResolvedConfig`**: New enum distinguishing between ready (env var / plaintext) and pending (keychain) credential resolution at startup
- **`resolve_connection_url_lazy()`**: New startup path that parses config but skips keychain access; the eager path (`resolve_connection_url_or_check`) is preserved for `--check-connection`
- **`new_lazy()` + `connect_with_url()`**: Refactored server constructors and connection logic for clean lazy/ready dual-path
- **Error reporting**: Enrich all MCP tool error responses with SQLSTATE/SQLCODE diagnostics

## Behavior

| Scenario | Before | After |
|---|---|---|
| Startup + env var password | No keychain | No keychain (unchanged) |
| Startup + plaintext password | No keychain | No keychain (unchanged) |
| Startup + `"keyring"` sentinel | **Reads keychain** | **Skips keychain** ✅ |
| Startup + no password | **Reads keychain** | **Skips keychain** ✅ |
| First gaussdb-mcp tool call | Uses cached connection | Reads keychain + connects |
| `--check-connection` flag | Reads keychain | Reads keychain (unchanged) |

## Test Plan

- [x] `cargo build -p gaussdb-mcp` passes (dev + release)
- [x] LSP diagnostics clean (0 errors)
- [ ] Manual: start gaussdb-mcp without calling tools → verify no keychain prompt
- [ ] Manual: call any gaussdb-mcp tool → verify keychain read + successful connection
- [ ] Manual: `gaussdb-mcp --check-connection` still works as before


paolobarbolini:
Slop.

PR #1354, opened 2026-05-28
Replace `async-trait` with `impl Future` in GenericClient

CathalMullan:
Related to https://github.com/rust-postgres/rust-postgres/issues/1332

Drop `async-trait` in favor of `impl Future<...> + Send` usage.


CathalMullan:
From a little more research, this shouldn't be merged until the MSRV is bumped to 1.92 if breaking inference is a concern.

`deadpool` hit this same issue when trying to replace `async-trait`: https://github.com/deadpool-rs/deadpool/issues/323

The inference problem was fixed by: https://github.com/rust-lang/rust/pull/144064

PR #1358, opened 2026-05-29
feat: expose type OID and modifier on SimpleColumn

tristan957:
The simple query protocol's RowDescription message carries the type OID and type modifier for each column, but SimpleColumn was dropping both. Clients rendering column types (e.g. distinguishing varchar(50) from text) need access to this metadata, matching what's already exposed on Column for the extended query protocol.

Closes: https://github.com/rust-postgres/rust-postgres/issues/1329

### DIFF
--- a/tokio-postgres/tests/test/main.rs
+++ b/tokio-postgres/tests/test/main.rs
@@ -340,9 +340,10 @@ async fn simple_query() {
         .simple_query(
             "CREATE TEMPORARY TABLE foo (
                 id SERIAL,
-                name TEXT
+                name TEXT,
+                description VARCHAR(50)
             );
-            INSERT INTO foo (name) VALUES ('steven'), ('joe');
+            INSERT INTO foo (name, description) VALUES ('steven', 'a'), ('joe', 'b');
             SELECT * FROM foo ORDER BY id;",
         )
         .await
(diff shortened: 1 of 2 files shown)


tristan957:
I see a PR was created at https://github.com/rust-postgres/rust-postgres/pull/1338. The PR has a lot of changes that seem _strange_, since this is a pretty simple change. That PR also only exposes the OID and not the type modifier. Happy to close the PR if the #1338 is a better candidate for merging.

Thanks for rust-postgres. I used it at a previous job, and it is quite nice to use.

tokio-postgres/tests/test/main.rs
```diff
@@ -340,9 +340,10 @@ async fn simple_query() {
         .simple_query(
             "CREATE TEMPORARY TABLE foo (
                 id SERIAL,
-                name TEXT
+                name TEXT,
+                description VARCHAR(50)
```
tristan957, line 344: I added this column to be able to test that the type modifier is not always -1.

tristan957:
Should we expose other information in https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-ROWDESCRIPTION? The data type size, the column attribute number, format code, and the table OID sound interesting.

tristan957:
@paolobarbolini anything I can do to help push this PR along?

PR #1359, opened 2026-06-01
Draft: Named prepared statements behind feature flag for 

ReinierMaas:
Would you even be interested in guarding the named prepared statements behind a feature flag?

This allows us to disable these named prepared statements and ensure at compile time that we are only using unnamed prepared statements. Please let me know if you want some commit (range) to review/land as separate PR(s) instead.

I also have to go over the issue tracker to link issues to commits that resolve them, I found a few flakes and moved internal constructs to unnamed prepared statements in order to expose at least `query_typed` in this mode. I am sure there is the possibility to expose more of the interface without named named prepared statements.

Naming of the feature flag, now that I am re-reading/reviewing my PR I think `named-prepared-statements` is a better fit. It is also easier to find/understand for the downstream users.

## Summary

This adds a default-on `named-prepared-statements` feature to `tokio-postgres` and `postgres`.

When the feature is enabled, the API behaves as it does today. When disabled, APIs that create protocol-level named prepared statements are not exposed. This gives downstream users a compile-time escape hatch for environments where prepared statement caching is not safe, such as PgBouncer pool modes that do not support session-local prepared statement state.

The branch is split into small commits so the mechanical/test/CI groundwork can be reviewed separately from the feature gating.

## Commit Breakdown

- `377dcb68 Stabilize is_closed test`

  Makes the existing sync `is_closed` test drive the connection a few more times after terminating the backend before asserting closure. This avoids a race where the failed health check has been observed but the internal client channel has not yet reported closed.

- `6400771b Add Docker Compose healthcheck`

  Adds a PostgreSQL healthcheck to `docker-compose.yml` so the standard `docker compose up --wait` interface waits for the test database to be ready.

- `00a50d23 Make derive domain test idempotent`

  Moves the nested composite/domain setup in one derive test into `pg_temp`. The previous test created globally named types, so running workspace tests multiple times against the same database could fail with “type already exists.”

- `c8fc5279 Fix postgres-openssl no-default build`

  Fixes postgres-openssl so it compiles with default features disabled. This is needed before adding broader no-default CI coverage.

- `de30bbba Use unnamed queries for type metadata`

  Changes internal custom type metadata lookup to use unnamed typed queries instead of an internally cached named prepared statement. This separates type metadata resolution from implicit prepared statement caching.

- `eb9a701d Make postgres runtime dependency explicit`

  Makes the sync postgres crate’s dependency on `tokio-postgres` runtime support explicit. This keeps feature forwarding clearer before introducing the new feature gate.

- `094ace37 Use internal Statement imports`

  Updates internal code to refer to `Statement` through internal modules instead of relying on the public re-export. This lets the public `Statement` API be feature-gated without breaking internal implementation code.

- `69874ea7 Add implicit prepared statements feature`

  Introduces the default-on `implicit-prepared-statements` feature in `tokio-postgres` and forwards it to `postgres`.

- `7a94cf5d Gate implicit prepared statement APIs`

  Applies the feature gate to APIs that implicitly prepare or use named statements. Typed unnamed query/execute APIs, simple query APIs, batch execution, transactions, cancellation, and type cache APIs remain available without the feature. This commit also adds CI coverage for no-default and all-features combinations.

- `e43c7adf Test no-default typed query surface`

  Adds no-default integration tests for the API surface that remains available when implicit prepared statements are disabled, including custom type metadata resolution.

- `3ff0be94 Document implicit prepared statements feature`

  Adds changelog entries documenting the new default-on feature and the intended no-default behavior.

## Validation

Each commit in the stack was validated independently with its own CI command set, plus formatting, `clippy`, and `wasm` checking.

Commands covered include:

```sh
cargo fmt --all -- --check
RUSTFLAGS=-Dwarnings cargo clippy --all --all-targets
RUSTFLAGS='--cfg getrandom_backend="wasm_js"' cargo check --target wasm32-unknown-unknown --manifest-path tokio-postgres/Cargo.toml --no-default-features --features js
cargo check --all-targets --all-features
cargo check --all-targets --no-default-features
cargo check --manifest-path tokio-postgres/Cargo.toml --all-targets --no-default-features
cargo check --manifest-path postgres/Cargo.toml --all-targets --no-default-features
cargo test --all
cargo test --manifest-path tokio-postgres/Cargo.toml --no-default-features
cargo test --manifest-path postgres/Cargo.toml --no-default-features
cargo test --all --all-features
```

The final stack was also checked with `cargo test --all` followed by `cargo test --all --all-features` against the same fresh Docker database to ensure repeated workspace test runs are idempotent.


paolobarbolini:
> When the feature is enabled, the API behaves as it does today. When disabled, APIs that implicitly create protocol-level named prepared statements are not exposed. This gives downstream users a compile-time escape hatch for environments where prepared statement caching is not safe, such as PgBouncer pool modes that do not support session-local prepared statement state.

Not a fan of this. This is not the first time this issue is reported, and I wish we would come up with a proper solution either on our side or on the proxy side instead of piling on workarounds.

ReinierMaas:
> Not a fan of this. This is not the first time this issue is reported, and I wish we would come up with a proper solution either on our side or on the proxy side instead of piling on workarounds.

Thanks for the feedback! I do want to note that the `query_typed` still issues prepared statements, see:

* https://github.com/rust-postgres/rust-postgres/issues/1204.

This is done for resolving type metadata which prevents enabling transaction pooling in production environments.

I can split of [`Use unnamed queries for type metadata`](https://github.com/rust-postgres/rust-postgres/pull/1359/commits/de30bbba846335c94c0dfa8a30d98256d5cb8a09) to resolve that issue with a more targetted approach.

Please let me know if you want me to split the PR at that boundary!

---

Preventing the use of named prepared statements would then need to be ensured in the test suite instead of guarding it with a feature flag. Preventing it from being compile-time checked.

ReinierMaas:
I separated the compilation fixes when toggling the feature flags into: https://github.com/rust-postgres/rust-postgres/pull/1366

PR #1363, opened 2026-06-11
fix wrong scram-sha-256 server-final-message error parser

### LINKED ISSUE #1362: wrong scram-sha-256 server-final-message error parser
https://github.com/rust-postgres/rust-postgres/blob/a6507ccb8418f837c29097156bd76c97677e963e/postgres-protocol/src/authentication/sasl.rs#L393-L395

It looks like a logical negation is missing in the error value parser.

exedealer:
fix #1362


paolobarbolini (approved):
Good catch. Thanks!

Issue #1081, opened 2023-11-13
Memory leak in long-lived connection pool


seanlinsley:
In the image below, memory usage is stable with `prune_db_connections` as described in https://github.com/bikeshedder/deadpool/pull/273, plus `malloc_trim(0)`. Then disabling `prune_db_connections` shows a clear memory leak.

<img width="1227" alt="Screenshot 2023-11-11 at 7 46 29 PM" src="https://github.com/sfackler/rust-postgres/assets/688886/87d9b4cc-2a4f-46eb-b6a9-41d122bbec7c">

Heaptrack shows two leaks, both around `BytesMut` usage:

<img width="1254" alt="Screenshot 2023-11-12 at 4 48 13 PM" src="https://github.com/sfackler/rust-postgres/assets/688886/1dfca275-6da4-462f-b1c7-a020026f8ce5">
<img width="853" alt="Screenshot 2023-11-12 at 5 12 03 PM" src="https://github.com/sfackler/rust-postgres/assets/688886/a1946473-9e5f-44b9-bb97-883cea0ef80b">

Related discussions online:
- https://stackoverflow.com/questions/73725299
- https://github.com/actix/actix-web/issues/1780
- https://github.com/hyperium/hyper/issues/1790
- https://github.com/linkerd/linkerd2/issues/11012
- https://github.com/tokio-rs/bytes/issues/634

Possible debugging steps:
- add code to track every message sent/received, where the buffer size has grown since the last message (might require a fork of the bytes crate)
- write a reproduction script that's similar to the production workload (RDS, binary COPY IN, queries that cause unique constraint error)

Possible fixes:
- revert https://github.com/sfackler/rust-postgres/commit/ffd7245e54e8524881e9a9ebb5d6661bcd4c010c
  - this would likely fix the leak if it's actually a bug in `BytesMut` usage and not something like closed connections never being freed. and it'd be a good opportunity to confirm/deny the performance difference that motivated the change originally

sfackler:
Do you have a self contained reproduction?

seanlinsley:
I've tried to write a reproduction script, but it doesn't leak memory. If there is a bug in `BytesMut` usage it might be around the exact handling of messages sent/received to a remote database, using both regular queries and binary COPY IN, and with some queries resulting in unique constraint errors.

inzanez:
@seanlinsley is there any update on this? I am seeing similar things (queries with many parameters) increasing memory usage quite heavily.
[...]

PR #1367, opened 2026-06-27
Fix memory retention issue in BytesMut and add unit tests

tsharp:
This PR reduces memory retained by long-lived `tokio-postgres` connections after unusually large reads or writes. This fixes issue #1081 

Previously, several internal `BytesMut` buffers could grow to handle a large response or request and then keep that allocation for the lifetime of the connection, even after the data had been fully consumed. This could make connection memory usage remain high after one large query, COPY operation, or encoded request.

The fix introduces a configurable `buffer_size` connection setting, defaulting to `8192`, and uses it as the target retained capacity for connection buffers.

**Changes**
- Adds `Config::buffer_size(...)` and `Config::get_buffer_size()`.
- Adds `buffer_size` support to connection strings and URL parameters:
  - `buffer_size=16384`
  - `postgresql://localhost?buffer_size=16384`
- Uses `Framed::with_capacity(..., config.buffer_size)` when constructing the connection stream.
- Initializes `BackendMessages` and client write buffers with the configured size.
- Drops oversized drained response batches before waiting for the next batch.
- Drops oversized client write scratch buffers after `InnerClient::with_buf` completes.
- Adds deterministic unit tests for the memory-retention paths.

**Why**
`BytesMut::clear()` resets length but does not release capacity. Similarly, a drained `BackendMessages` could continue holding a large allocation while a `Responses` object waited for more data. On long-lived connections, these retained allocations could look like memory growth that never recovers after large operations.

**Testing**
Added tests covering:
- Drained oversized backend response batches are reduced back to the configured buffer size.
- Oversized client write scratch buffers are reduced after use.
- `buffer_size` parsing for key-value and URL connection strings.
- Invalid `buffer_size=0` is rejected.

### DIFF
--- a/tokio-postgres/src/client.rs
+++ b/tokio-postgres/src/client.rs
@@ -41,7 +41,6 @@ use tokio::io::{AsyncRead, AsyncWrite};
 pub struct Responses {
     receiver: mpsc::Receiver<BackendMessages>,
     cur: BackendMessages,
-    buffer_size: usize,
 }
 
 impl Responses {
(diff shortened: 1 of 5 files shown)


tsharp:
I am still verifying the fix and will add a follow up comment and additional findings here when done.

tsharp:
@paolobarbolini - the builds are ready

mymtw:
@sfackler plz can you review this fix?

PR #1370, opened 2026-07-11
Add host fallback of `/run/postgresql` on Unix

### LINKED ISSUE #82: Default values for peer auth
Currently there's a discrepancy between what psql (`libpq`) does and what postgres-rust does:
To get the equivalent of this:

```
$ psql -q postgresql:///agg
agg=>
```

I had to issue:

```
    let pgconn = Connection::connect("postgresql://user@%2Ftmp/agg", &SslMode::None).unwrap();
```

which shows that the user name and the `/tmp` location are mandatory with `rust-postgres`. The user name is not a big issue, because the error message is very explicit, but I find the path to the sockets dir to be more problematic. 

Actually, `libpq` has several levels of default values for the connection string (or the URL).
In the case of peer authentications, which I believe to be really important for newcomers starting to tinker with the library, absence of host and port specifications fall back to
- PGHOST and PGPORT environment variables
- statically compiled default values

The tricky part is that the statically compiled default value depends on the OS / distribution.
For instance, on Linux distributions from the RedHat family, it's `/tmp` (I believe this to be the default from upstream), whereas on the Debian family, it's `/var/run/postgresql`

Of course it's really too early to have rust-postgres packaged by all major distributions with their own traditions, but these default values actually give back some uniformity.

I don't know yet Rust build tools enough, but I suppose that providing such a configurable default should be doable.

jtgeibel:
This implements a fallback unix socket path when the host is empty, allowing for shorthand such as `postgres:///` to connect with all default parameters.

There is one subtle change in behavior. Previously `host=,/does/not/exist port=5433 user=postgres` and `postgres://postgres@:5433/?host=/does/not/exist` would first attempt a TCP connection to an empty host name and then try the 2nd host. Now, it could successfully connect to the fallback path. In the former case it already looks weird with the comma, but in the latter case the first host is much more implicit. Thus, its not clear to me if this would be an acceptable change in a point release.

Probably closes #82, though the basing the fallback on environment variables was not implemented.

### DIFF
--- a/tokio-postgres/src/config.rs
+++ b/tokio-postgres/src/config.rs
@@ -369,7 +369,13 @@ impl Config {
     ///
     /// Multiple hosts can be specified by calling this method multiple times, and each will be tried in order. On Unix
     /// systems, a host starting with a `/` is interpreted as a path to a directory containing Unix domain sockets.
-    /// There must be either no hosts, or the same number of hosts as hostaddrs.
+    ///
+    /// On Unix systems, when [`connect`] is called an empty host string will be interpreted as `/run/postgresql` if
+    /// and only if the `hostaddr` vec is empty. This is done at connect time because at configure time we cannot be
+    /// sure that a `hostaddr` will not be added later, and doing so would force every connection to TCP where we want
+    /// to preserve the empty host name.
+    ///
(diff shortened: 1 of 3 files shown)


jtgeibel:
The unrelated clippy error is addressed in #1371.

PR #1374, opened 2026-08-03
fix: don't leak server-side state when a future is cancelled

bikeshedder:
While working on [deadpool-postgres](https://crates.io/crates/deadpool-postgres/) I discovered memory leaks caused by cancelled futures. Most of the code - was written by Claude Code but meticulously reviewed by me.

The only code where I let Claude do it's thing and didn't bother improving upon were the tests. This part of the code is AI slop at its best but shows the problem to the point. 

You can just check out the [test(tokio): add failing tests for cancellation resource leaks](https://github.com/rust-postgres/rust-postgres/pull/1374/commits/1d135cb8a2d8d6ff727e956637678ae64bc54e74) commit and see the leak by running the test suite.

This PR also includes a bit of refactoring which made patching the actual bug a bit more straight forward.

----

Fun fact: This kind of error plagued the first few versions of `deadpool`. I learned the hard way that every `.await` could mean that the code following this expression will never be executed if the future gets dropped. The only way to clean up reliably is by utilizing the `Drop` trait.


bikeshedder:
btw. please don't squash and merge this PR. The commits are stacked in a meaningful way:
1. Add tests that show the resource leak
2. Refactor code a bit to make the fix simpler
3. The Actual fix and the new DropGuard primitive
4. Refactor TransactionBuilder to use DropGuard as well

PR #1378, opened 2026-08-29
feat: support OAUTHBEARER SASL authentication

### LINKED ISSUE #1377: Support SASL OAUTHBEARER authentication (PostgreSQL 18 `oauth` HBA method)
PostgreSQL 18 added the `oauth` HBA method, which authenticates a connection with an OAuth 2.0 bearer token over the SASL `OAUTHBEARER` mechanism ([RFC 7628](https://datatracker.ietf.org/doc/html/rfc7628)) instead of a password. `tokio-postgres` cannot connect to a server configured that way: `authenticate_sasl` matches only `SCRAM-SHA-256` and `SCRAM-SHA-256-PLUS`, so the server's mechanism list is refused with `authentication error: unsupported SASL mechanism`.

### Proposed scope: token-first only

The caller supplies a token; the crate performs the SASL exchange. **No discovery, no device authorization flow, no conversation with an identity provider.** 

### Proposed config shape

A closure returning a future, so an expiring token can be refreshed per connection attempt rather than baked into the `Config`:

```rust
config.token_provider(|| async { Ok(fetch_token().await?) });
```

This is additive: a new field, a new setter, a new getter. Nothing existing changes signature, so as far as I can tell it does not need #1332.

### Prior art in this repo

#1325 opened this ground in February with the same token-first scope and a token-provider closure. It has been dormant since and does not merge cleanly. I have a branch that picks it up rather than competing with it and will open it as a PR against this issue — happy to close either in favour of the other if @JoHaHu would rather carry it themselves.

geoHeil:
Closes #1377.

PostgreSQL 18 added the `oauth` HBA method, which authenticates a connection with an OAuth 2.0 bearer token over SASL `OAUTHBEARER` ([RFC 7628](https://datatracker.ietf.org/doc/html/rfc7628)) instead of a password. `authenticate_sasl` matched only the SCRAM mechanisms, so such a server was refused with `unsupported SASL mechanism`.

### Scope

Token-first only. The caller supplies a token, the crate performs the SASL exchange. **No discovery, no device authorization flow, no conversation with an identity provider** — the same line `node-postgres` draws, and the same thing libpq does when an application installs a custom flow. No token is requested unless the server asks for `OAUTHBEARER`.

```rust
let mut config = Config::new();
config.token_provider(|| async { Ok(fetch_token().await?) });
```

Verified in Pg18 docker postgres.

FYI: Clippy was failing on some files in CI - I have fixed this (inlined here) to have a green CI


geoHeil:
The `clippy` job here is red on something this branch does not touch, so before anyone spends time on it:

```
error: this lifetime isn't used in the type
  --> tokio-postgres/tests/test/types/mod.rs:51:24
   |
51 |     T: PartialEq + for<'a> FromSqlOwned + ToSql + Sync,
   |                        ^^
   = note: `-D clippy::extra-unused-lifetimes` implied by `-D warnings`
```

That line was last changed in 10a7724 (March) and is not in this diff — the four files here are `postgres-protocol/src/authentication/sasl.rs`, `tokio-postgres/src/{config,connect_raw}.rs` and `postgres/src/config.rs`. It reproduces on the base commit:

```
$ git checkout 1084ca8f
$ cargo +1.98.0 clippy -p tokio-postgres --test test --all-features
warning: this lifetime isn't used in the type
  --> tokio-postgres/tests/test/types/mod.rs:51:24
```

`extra_unused_lifetimes` learned to see through an HRTB in a clippy released after master's last CI run on 27 July, and the `clippy` job pins `stable` rather than a version, so this PR is just the first push to meet the new one. `test`, `rustfmt` and `check-wasm32` are green.

Happy to fold the one-line fix (drop the vacuous `for<'a>`) into this branch, or send it separately so it isn't tangled up with a feature — whichever you prefer. I left it out for now rather than mixing an unrelated fix into a review.

geoHeil:
## Verified end-to-end against a live PostgreSQL 18.6

This PR declared an authenticated OAUTHBEARER connection unproven. It is proven now.

### The server was trusted before the driver was tested

That ordering was deliberate: testing an unverified client against an unverified server makes a pass ambiguous — two bugs can cancel. So the server was stood up first (`postgres:18-trixie`, PostgreSQL 18.6, an `oauth` HBA line, a validator module, a static RS256 issuer) and verified with **two independent clients** — Go `pgx` v5.10.0 and libpq via `PQsetAuthDataHook` — which agreed on every case. Only then was this branch pointed at it, from the host over the published port, so it is wire-level independent of the container.

### Results

| Case | Result |
|---|---|
| `aud` matches | **ACCEPTED** — `select current_user, version()` returned `testuser \| PostgreSQL 18.6` |
| `aud` names another service | **REFUSED** — `DbError`, SQLSTATE **28000** |
| `aud` claim absent | **REFUSED** — same |

```
CONNECT FAILED: db error
  caused by: FATAL: OAuth bearer authentication failed for user "testuser"
  db_error severity=FATAL code=28000
          message="OAuth bearer authentication failed for user \"testuser\"" detail=None
```

### The #1325 bug is fixed, not merely absent — and I have a negative control for it

The failure round this PR adds — the lone RFC 7628 kvsep reply to the server's error challenge — is **load-bearing**, and I proved it rather than asserting it. Suppressing that reply, rebuilding, and retrying the rejected-token case:

```
exit=124  elapsed=25s   (no output — killed by timeout)
```

It **hangs indefinitely**, waiting for a message the server will not send until it receives the kvsep. Restored → immediately back to a clean `28000`. The happy path was unaffected by the control, which localises the effect to the failure path.

So without that reply the failure mode is *worse* than the `unexpected message from server` that #1325 would have produced: it is an indefinite hang. Worth knowing for review, since it is the one part of this diff that has no visible effect on the success path.

### One observation, not a request

Neither this driver nor sqlx surfaces the RFC 7628 JSON status document — both discard the challenge and report the server's subsequent `ErrorResponse`, which is the better diagnostic. The cost is that `invalid_token` and `insufficient_scope` are not distinguishable to the caller, where `pgx` does expose them. Fine as a first cut; noting it so it is a known limit rather than an oversight.

### Limits

No TLS in this run — `oauth` does not require it server-side and PostgreSQL's own suite authenticates over a `local` socket, so this was correct for isolating the mechanism. It does mean the `sslmode` interaction is unexercised, and the point in the PR body about a bearer token wanting an encrypted channel stands untested either way.

*Verified with Claude Code against PostgreSQL 18.6 in Docker on 2026-08-29. Environment torn down; nothing committed or pushed to this branch — the fork is clean at `973624b7`.*

Issue #1016, opened 2023-04-05
Allow passing `search_path` as client-side params


Weakky:
## Overview

It's currently impossible to define the `search_path` without issuing a query to the database, although it is possible to set it up as client options when connecting to PG. See https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-SEARCH-PATH

Are you open to allowing this additional config?

I have a PR opened on our fork https://github.com/prisma/rust-postgres/pull/1. Let me know if you'd be ok with roughly having the content of that PR merged in your repo.

Thanks 🙏

sfackler:
Like with libpq, you'd pass that in the `options` field: https://stackoverflow.com/a/66287119

Issue #608, opened 2020-06-04
rust-postgres setting timezone to UTC on connecting is not appropriate


jmafc:
If you issue the query `SELECT CURRENT_TIMESTAMP` in `psql` or from other language adapters such as psycopg2, the Postgres server will return a value adjusted to what it considers its timezone (specifically, based on the value in the `timezone` parameter in `postgresql.conf`). In `psql` the value returned shows the timezone adjustment to UTC, e.g., if the `timezone` in `postgresql.conf` is set to `America/Denver` (MST), then the value is suffixed with `-07` (or `-06` during DST).
rust-postgres overrides this normal situation by sending `timezone=UTC` when connecting to the server. https://github.com/sfackler/rust-postgres/blob/2ce4f08f46b48bb42522cba08a0986da00d1de4c/tokio-postgres/src/connect_raw.rs#L114 This not only affects the values of the query above but any query that depends on the server's notion of "local time". For example, if a website shows a different result, based on whether it's past midnight in the local sense (i.e., based on the Postgres server's actual timezone location), if retrieved via rust-postgres the results will change starting at 0:00 UTC, e.g., possibly eight/seven hours earlier than desired if the server is located in the Pacific timezone. The only workaround is for the client to send a `SET TIMEZONE timezone-value` query *before* sending such a time-sensitive query.
Therefore I think the timezone parameter should not be sent at all, or if sent, it should be configurable (with the default being "don't send it").

sfackler:
This is necessary to enable conversion of types to and from time datatypes. If you aren't using those, you can set your own timezone via the `options` component of the connection config: https://docs.rs/postgres/0.17.3/postgres/config/struct.Config.html#keys.

jmafc:
I *am* using the conversions, but I'm not sure why overriding the server's notion of its timezone is needed to convert timestamps. In `psql`, as mentioned above the server returns an indication of its "local time" in the form of the `(+|-)nn` suffix. In C libpq, IIRC a timestamptz value is also returned as a string with the same suffix. In psycopg2, as I showed in an email earlier, the timezone is also available, e.g., as `tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=-240, name=None)`, if I'm not mistaken, as the result of converting from the retrieved string with suffix.
I will try using `options` as suggested, but I still think overriding the server's timezone is not needed.

sfackler:
The timezone offset is not included in the binary encoding of the values.

jmafc:
[...]

Issue #147, opened 2015-12-03
More flexible timezone configuration


barosl:
rust-postgres currently sets `timezone` to `GMT` on startup. Recently, I was bitten by this behavior because I assumed it to be my local timezone. It seems that the GMT forcing is done to have an interoperability with the underlying `time` crate. So I wondered:
1. Would that be possible to remove that limitation? The official driver (libpq) doesn't have such a behavior, so it can be a bit confusing (to me, at least). However, I get that defaulting to GMT is [almost always a good idea](http://justatheory.com/computers/databases/postgresql/use-timestamptz.html), though.
2. Is it safe to `SET timezone to [local timezone]` manually? If the concern with the `time` crate still holds, it may break some assertions that the `Timespec` values are stored in GMT.

sfackler:
`SET timezone TO whatever` should work fine as long as you aren't trying to send `time` or `chrono` types back and forth.

The right solution is for the `ToSql` and `FromSql` implementations to be able to translate properly based on the session's timezone. I haven't seen any Rust crates that deal with timezones, though, unfortunately.

sfackler:
Oh yeah, one open question is: once tzdata support does make it in, what should the timezone be set to? The JDBC driver sets it to the local timezone, I believe, but we could also leave it as the default, or stick with GMT.

ohmyarch:
@sfackler Is there any progress? I don't think such a default behavior is reasonable. Whenever a connection is established, I have to execute a statement to modify the time zone setting. This is really annoying.

Jakubzik:
The local timezone seems to be the default setting of psql and JDBC, and would have been less surprising to me, too.

Also, I do not understand the effect of adding "options='-c timezone=Europe/Berlin'" to the connection configuration. There *is* an effect, because adding a non-existant timezone causes an error ("options='-c timezone=Curope/Berlin' results in a SqlState 22023 from postgres specifically complaining about the wrong timezone), but the timezone on the connection *remains* GMT even if I add the correct option for Berlin.
 
Would it be possible to make sure the option that is set in the configuration ("options=...") is transported to the connection?

sfackler:
I was looking through the PostgreSQL source and realized that ever since switching from the text format to the binary format, we haven't actually needed to set the time zone to UTC - the server just always uses UTC for the values of timestamp-aware types. Be aware that the `TimeZone` session parameter has no effect on values transmitted via ToSql and FromSql, but we should be able to stop explicitly forcing it to UTC.

PR #1382, opened 2026-09-17
feat(tokio): add Config::param for arbitrary startup parameters

### LINKED ISSUE #631: Allow setting arbitrary key-value connection parameters
In `StartupMessage`, Postgres accepts connection parameters (key-values) such as "user", "database". There are many more that can be set, e.g. "search_path", but `rust-postgres` does not support them directly.

They can be set using the "options" string, but it's pointless to bother encoding it into a string, dealing with escapes etc., just so that the server can decode them again. [Protocol docs](https://www.postgresql.org/docs/12/protocol-message-formats.html) also say this is deprecated:

> This is deprecated in favor of setting individual run-time parameters.

It would be nice if `rust-postgres` allowed setting an arbitrary vector of k-v tuples.

adriangb:
## What this adds

`Config` can set only a fixed set of run-time parameters in the StartupMessage: `user`, `database`, `application_name`, a hardcoded `client_encoding=UTF8`, and the `options` string. This PR adds a generic pair of methods, in the existing builder style:

```rust
let mut config = Config::new();
config
    .host("localhost")
    .user("postgres")
    .param("TimeZone", "UTC")
    .param("DateStyle", "ISO, MDY")
    .param("search_path", "myschema, public");

let (client, connection) = config.connect(NoTls).await?;
```

`param(name, value)` adds a run-time parameter. A second call with the same name replaces the value and keeps the position, so `Config` stays idempotent under repeated builder calls. `get_params()` reads the pairs back, in insertion order. The parameters are held in an ordered `Vec<(String, String)>`, not a map, so the startup packet is deterministic and the dependency set does not change.

Requested in #631 (open since 2020) and in #1016 (closed unimplemented). Closes #631.

## Why not `options`?

For a direct connection with simple values, `options` is functionally equivalent, and this PR is mostly about ergonomics. Two points are more than ergonomics:

**1. `options` makes each user write a shell-style escaper.** The backend receives `options` as command-line arguments, so a value which contains a space needs a backslash:

```rust
config.options(r"-c DateStyle=ISO,\ MDY -c search_path=myschema,\ public");
```

Get it wrong and the argument is mis-split, which the server accepts as a different setting or rejects with a message about the option parser, not about your value. `param` passes the value as it is.

**2. The protocol prefers individual parameters.** `options` is ["deprecated in favor of setting individual run-time parameters"](https://www.postgresql.org/docs/current/protocol-message-formats.html). The reporter of #631 asked pgsql-hackers in 2020 whether a driver may send an arbitrary GUC in the startup packet, and Tom Lane [answered](https://www.postgresql.org/message-id/2463369.1594514611%40sss.pgh.pa.us) "Sure": libpq rejects unknown *connection-string keywords* because that name set is its own and evolves separately from the GUC names, not because the protocol forbids the parameters.

This answers "the connection parameter API here mirrors libpq's currently" (#631) and "like with libpq, you'd pass that in the `options` field" (#1016). libpq's `options` keyword is the workaround for libpq's own closed keyword namespace; it is not the protocol's preferred path.

## Compared to `SET`

`SET` after connecting costs a round trip, and its value is lost on `RESET ALL` and `DISCARD ALL`, which connection pools run when they recycle a connection. A startup parameter acts as a session default, so `RESET ALL` restores it instead of discarding it.

Behind a transaction-mode pooler neither mechanism is a general answer, and the ability to set a parameter depends on the parameter, not on the mechanism. The `param` rustdoc says this, so nobody reads the new method as a cure for pooled connections.

## What this PR does not do

- **No behaviour change for existing users.** With no `param` call the startup packet is identical to today's, in today's order. A unit test asserts the full list.
- **No new defaults.** `TimeZone`, `DateStyle` and the rest stay unset. fbc42ec ("Stop setting timezone to UTC", #147, #608) removed the last implicit default because the driver overriding the server surprised users. A test asserts that a connection with no `param` call keeps the server's own `TimeZone`.
- **No new connection-string keyword.** An unknown keyword is still an error, so `sslmodee=require` keeps failing as a configuration error instead of reaching the server as a run-time parameter. Connection strings keep `options=-c%20TimeZone%3DUTC`, and a test asserts that `TimeZone=UTC` in a key-value string and in a URL is still rejected. The private `Config::param` which backs both parsers is renamed to `set_keyword`; its behaviour is unchanged.
- **`options` is untouched.** Same field, same semantics, same position in the packet.
- **No per-setting helpers.** No `timezone()` or `search_path()`. In #513 reviewers preferred `options` over a dedicated timezone knob, so a generic pair avoids a method per setting.
- **Additive and non-breaking.** One field plus two methods per crate. No trait or signature changes. It can ship in a patch release, so it does not need to wait for #1332.

## Tests

- `tokio-postgres/src/connect_raw.rs`: the parameter assembly moves into `startup_params(config, user)`, which `startup` calls. Three unit tests: today's list with no params, params appended in insertion order, and the documented precedence when `client_encoding` or `application_name` is passed as a param.
- `tokio-postgres/src/config.rs`: `param` replaces a value of the same name, `get_params` returns what was set, `Config` equality and `Debug` hold, and run-time parameters are still not connection-string keywords.
- `tokio-postgres/tests/test/main.rs`: against the compose database, a connection built with `.param("TimeZone", "UTC").param("DateStyle", "ISO, MDY").param("search_path", "pg_catalog")` reports those values from `SHOW`, with `pg_settings.source` set to `client`. A companion test covers the no-param case.

`cargo fmt --all` and the suite against the compose database are clean, and this change adds no clippy warning. The `clippy` job fails on `clippy::extra_unused_lifetimes` in `tokio-postgres/tests/test/types/mod.rs:51`, a file which this PR does not touch and which fails the same way on a clean `master`. See the comment below.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VbAiXxMoSQMQ4x7sz58xDp


adriangb:
The `clippy` job fails on a lint in a file which this PR does not touch:

```
error: this lifetime isn't used in the type
  --> tokio-postgres/tests/test/types/mod.rs:51:24
   |
51 |     T: PartialEq + for<'a> FromSqlOwned + ToSql + Sync,
   |                        ^^
   = note: `-D clippy::extra-unused-lifetimes` implied by `-D warnings`
```

This also happens on a clean `master` with the current stable toolchain, so it is a new lint in a newer clippy, not a regression from this change. `cargo clippy --all --all-targets` on `master` at 1084ca8 reports the same warning, which CI promotes to an error.

The fix is to delete the unused `for<'a>`. I left it out of this PR to keep the diff single-purpose, but I am happy to add it here or to send it as a separate PR — whichever you prefer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MqYAL8UNki8BBsZJasNL1K

postgres/src/config.rs
```diff
@@ -222,6 +236,67 @@ impl Config {
...
+    /// [server configuration](https://www.postgresql.org/docs/current/runtime-config.html).
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use postgres::Config;
+    /// let mut config = Config::new();
+    /// config.param("TimeZone", "UTC");
+    /// config.param("search_path", "myschema, public");
+    /// ```
```
adriangb: This is too wordy, let's cut it down.

tokio-postgres/src/config.rs
```diff
@@ -339,6 +355,72 @@ impl Config {
...
+    /// [server configuration](https://www.postgresql.org/docs/current/runtime-config.html).
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use tokio_postgres::Config;
+    /// let mut config = Config::new();
+    /// config.param("TimeZone", "UTC");
+    /// config.param("search_path", "myschema, public");
+    /// ```
```
adriangb: Also to wordy.

postgres/src/config.rs
adriangb: Cut down in 5210cc5f: 35 lines of prose down to 12. It keeps same-name replacement, order and precedence, no client-side validation, the pooler caveat and the `options` alternative for connection strings.

tokio-postgres/src/config.rs
adriangb: Cut down in 5210cc5f: 35 lines of prose down to 12. It keeps same-name replacement, order and precedence, no client-side validation, the pooler caveat and the `options` alternative for connection strings.

